PR #15, opened 2026-08-07
Fix the required-layer table's drift from the shipped descriptors (#14)

### LINKED ISSUE #14: scene-layer-protocol.md's required-layer table has drifted from the shipped descriptors
The required-profiles table in `docs/instruments/scene-layer-protocol.md`
("Required panel profiles") says the PFD requires `Attitude, Tapes,
Annunciation` with `Guidance` optional. The shipped descriptor requires
Guidance too: `crates/pilotage-instrument-panels/src/descriptors.rs:59-62`
sets `Attitude | Tapes | Guidance | Annunciation`. And the monitor panel
(`descriptors.rs:277`, `Tapes | Annunciation`) is absent from the table
entirely.

One of the two is wrong, and the table is exactly the prose a backend author
will trust — the document is the layer-protocol contract. Decide which side
is authoritative:

- if the descriptor is right (every PFD frame opens the Guidance band; the
  CDI/flight-director band is load-bearing), fix the table and add the
  monitor row — doc-only;
- if the document is right (Guidance genuinely optional for a PFD), the
  descriptor change moves the scene digest and re-runs admission — a
  different, bigger PR with its own pin advance.

Either way the guardrail lands with the fix: the table is derived from, or
checked against, the `required_layers` masks of `BUILTIN_PANELS`, so it
cannot drift again — the same family as the standards-registry drift guard.
Belongs alongside #7, which will cite this table.

Found while mapping the crates for the composition/resize architecture work
(#11 / #12). Doc-only branch has no display behavior change; the
descriptor branch would be its own reviewed digest move under AIR-BAS-005.

luofang34:
Fixes #14.

## Which side was authoritative

The descriptor. `draw_pfd` opens the `Guidance` band unconditionally —
its own comment says so ("The band is always emitted so the layer
contract holds under every degradation; its content disappears with the
director"). So this is the issue's **doc-only branch**: the table was
wrong, not the mask.

## What changed

- The PFD row gains `Guidance` in required and loses it from optional.
- The monitor panel gains the row it never had
  (`Tapes`, `Annunciation` required).
- Prose records *why* the PFD requires a band whose content is often
  empty, so a backend author does not read an empty `Guidance` band as
  a missing one.

## The guardrail

`layer_profile_doc_tests` derives the required column from each
descriptor's `required_layers` mask and the optional column from its
complement over the six defined layers, parses this table out of the
document, and fails on any disagreement — the drift-guard family the
issue asked for.

Both halves of the reported drift were verified to fail the new tests
when reintroduced:

- old PFD row → `documented required layers disagree with required_layers`
- monitor row deleted → `the profiles table must list every shipped panel`

## No display behavior change

Scene digest still `bd85b8537f0b3e4abf8cf3ad3d36c6abfdceac15355639af2804d58dd9c61931`
(bench reports "matches pin"), corpus and raster baselines untouched.

## Gates run locally

fmt, clippy `-D warnings`, test `--all-targets`, doc `-D missing_docs`,
release build, `thumbv7em-none-eabihf` no_std closure, check-structure,
instrument-requirements, certification-claims, standards-registry,
instrument-bench, and both HARD evidence-gate steps — all green.

### DIFF
--- a/crates/pilotage-instrument-panels/src/descriptors.rs
+++ b/crates/pilotage-instrument-panels/src/descriptors.rs
@@ -462,4 +462,6 @@ pub const BUILTIN_SCENE_DIGEST: &str =
 #[cfg(test)]
 mod digest_tests;
 #[cfg(test)]
+mod layer_profile_doc_tests;
+#[cfg(test)]
 mod tests;
--- a/crates/pilotage-instrument-panels/src/descriptors/layer_profile_doc_tests.rs
+++ b/crates/pilotage-instrument-panels/src/descriptors/layer_profile_doc_tests.rs
@@ -15,7 +15,9 @@ use super::{BUILTIN_PANELS, layer_bit};
 
 const PROTOCOL_DOC: &str = include_str!("../../../../docs/instruments/scene-layer-protocol.md");
 
-const SECTION: &str = "## Required panel profiles";
+const HEADING: &str = "## Required panel profiles";
+
+const HEADER: [&str; 3] = ["Panel", "Required layers", "Optional layers"];
 
 /// Every defined layer, ascending — the order both table columns list.
 fn defined_layers() -> Vec<LayerId> {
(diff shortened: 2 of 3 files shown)


luofang34:
## Independent review — findings addressed

An independent reviewer verified the change against the repository rather than against this description. Summary of what it confirmed and what it caught.

**Confirmed independently:**
- The "descriptor is authoritative" decision is correct. `draw_pfd` has no `return` at all; `declutter` gates only the band's *content*, and `BackgroundMode` touches only the `Background` band. The sole Attitude/Tapes-without-Guidance exit is `?` propagation, which returns `Err` and is converted to `AdmissionError::Draw` before the required-layer check ever runs.
- The reviewer also noted the PFD Guidance requirement is *already* empirically enforced on `main`: `builtin_panels_pass_admission` runs 121 state × withholding cases through the required-layer check.
- All three table rows match the masks exactly, in ascending layer order.
- `encode.rs` emits `BeginLayer` before `Save`, so an empty Guidance band still sets its present bit — the doc's new "must not read an empty band as a missing one" claim is accurate.

**Caught — three ways the guard could pass vacuously**, each reproduced by the reviewer and each now fixed in 9bdb147:

| Hole | Wrong-but-passing edit | Now |
|---|---|---|
| Header sentinel was content-blind | a false row whose first cell is literally `Panel` | fails |
| `is_rule` vacuously true on empty cells | a blank row `\|  \|  \|  \|` | fails |
| Parser stopped at the first blank line | a second table added inside the section | fails |

Parsing is now positional (header, delimiter, then data), bounded to the section by the next heading, plus a new test pinning the section to exactly one table. I re-ran the reviewer's three attack edits and the original #14 drift: all four now fail the suite.

Also addressed: `a_drifting_mask_is_caught` was near-tautological — it passed even when the PFD mask genuinely drifted. It now checks both directions against the cell the guard actually reads. And `documented[0]` no longer indexes directly. The prose no longer hard-codes "six defined layers" while the test derives from `LAYER_COUNT`.

Digest still `bd85b85…` (matches pin). CI green.

Issue #13, opened 2026-08-07
Repository layout: no tier rules and no home for future instrument sets


luofang34:
The workspace is a flat list of thirteen `pilotage-*` crates plus tools
(`Cargo.toml`). The tier structure — the `no_std` kernel a panel may use, the
verification machinery, the shipped panels — exists only as prose in
`crates/README.md`. Nothing in the tree says where a second panel set goes,
what it may depend on, or what it must export. #6 (no set mechanism), #7 (no
authoring contract), and #9 (consumer-named crates) are three symptoms of one
missing decision: what this repository looks like when instruments of any
kind keep arriving.

## Tiers, enforced

- **Kernel** — the `no_std` closure a panel may depend on: frames, alerts,
  sha256, state, scene, glyphs, symbology. Stays under `crates/`.
- **Verification and registry** — raster, conformance, registry, evidence.
  Stays under `crates/`. This tier consumes sets; it is never a dependency of
  one.
- **Sets** — panel providers, one crate per set, each exporting one
  `PanelSet` (#6), living under `sets/`. The builtin panels crate moves there
  as the first inhabitant. A set crate depends on the kernel only — never on
  raster, conformance, or the registry.
- **Tools** — unchanged under `tools/`.

The dependency law *is* the structure, so it gets the same treatment as the
consumer boundary: `scripts/check-structure.sh` (already in the gate battery)
gains the tier rules, and a set crate whose manifest reaches outside the
kernel fails CI exactly as the downstream-agnostic gate holds the consumer
boundary closed today.

## One authoring path

A set author should find, in order: `sets/` (where it lives),
`docs/instruments/panel-contract.md` (#7 — what it must honour), and a worked
example that compiles. The strongest form of #7's worked example is a real
minimal set crate in `sets/`, admitted in CI, doubling as the template — one
source of truth instead of a doc snippet and a skeleton drifting apart. (The
deliberately dishonest panel that must *fail* admission belongs in the
conformance tests, not in the template.)

`crates/README.md` reorganizes by tier, each tier stating its dependency law,
and the crate list is checked against the workspace members so the map cannot
drift — the same guardrail family as the standards-registry drift guard.

## Naming

This issue executes whatever #9 decides; the recommendation recorded there is
`indicate-*`. One boundary worth restating here because it shapes the tree:
renames touch crate names, directory names, and doc prose — never hashed or
pinned byte strings (`SCENE_DIGEST_DOMAIN`, corpus metadata). Identifiers are
not names.

[...]

PR #16, opened 2026-08-07
Rename the crate family after this repository, not after a consumer (#9)

### LINKED ISSUE #9: Crates and docs are named after a downstream consumer
Every crate in this repository is named `pilotage-*`, and this repository is
upstream of Pilotage. The crate map lists thirteen of them; the docs still open
with "what Pilotage instrument displays may do"
(`docs/instruments/README.md:3`), and `README.md` describes the boundary in
terms of consumers of which Pilotage is only one.

The names claim ownership by a downstream consumer. A third-party shell, or the
avionics-link EFB path, depends on `pilotage-instrument-panels` while having
nothing to do with Pilotage. `pilotage-frames` and `pilotage-alerts` came along
in the extraction and are now general-purpose leaves under a consumer's name.

## Why now rather than later

Renaming is mechanical but touches every consumer's manifest, and the cost
scales with how many manifests exist and how many crate names each one pins.
Right now that is Pilotage plus avionics-link's seven pinned crates. A second
panel set adds provider crates whose names would be chosen to match the
existing scheme, and every one of those is another name to change later. This
is the cheapest it will ever be.

The counter-argument is real: churn for no functional gain, and every pinned
revision on either side of the rename resolves to a different crate name. If
the answer is "keep the names", that is a decision worth writing down in the
README so it stops looking like an oversight.

## Ask

Decide, and act on it before the second set:

- rename to the repository's own identity (`indicate-instrument-panels`, or
  drop the redundant prefix entirely — `instrument-panels`, `instrument-scene`);
- or record why `pilotage-` stays, so consumers stop reading it as a
  dependency on Pilotage.

Either way, the doc text that still says "Pilotage instrument displays" should
name the family rather than one of its consumers.

## Plan

Decision recommended: rename to the repository's own identity — `indicate-*`,
keeping the role part of each name (`indicate-instrument-scene`,
`indicate-frames`, …). Dropping the prefix entirely reads well in-tree but
squats generic names; the repo-named prefix keeps a vendored crate
attributable to its source.

Scope — **names, never identifiers**:

- crate directories, package names, workspace `members` and
  `workspace.dependencies`, the thumbv7em check list in CONTRIBUTING.md and
  CI, `-p` flags in scripts, doc prose ("Pilotage instrument displays" →
  the family's own name, `docs/instruments/README.md:3`), `crates/README.md`.
- **Not** any hashed or pinned byte string:
  `SCENE_DIGEST_DOMAIN = b"pilotage-scene-digest-v1"` (`digest.rs:25`),
  corpus metadata, anything content-addressed by the evidence graph.
  Those are identifiers, not names — renaming one moves every digest and
  reddens every pin for zero paint change. The README records this so the
  surviving string stops looking like an oversight, which is the same trap
  this issue's title names, one level down.

Mechanics: one PR — `git mv` the directories, mechanical rename of package
names and paths, full gate battery green. Acceptance: `BUILTIN_SCENE_DIGEST`,
the corpus hash, and all raster baselines unchanged (nothing behavioral
moved). Guardrail in the same PR: `scripts/check-structure.sh` asserts no
`pilotage-`-named crate remains, so a stray revert cannot reintroduce one.

Timing, from the shared roadmap (#13): **first** — before #6 mints a
set-crate naming scheme, before #13 creates `sets/`, and before #8's
first tag freezes a name a consumer would pin. Consumers advance pins on
their usual discipline; historical revs keep historical names, so nothing
breaks mid-history.

luofang34:
Fixes #9.

Decision taken as recommended in the issue: rename to the repository's own identity, `indicate-*`, keeping the role part of each name. Dropping the prefix entirely would squat generic names; the repo-named prefix keeps a vendored crate attributable to its source.

## Acceptance criterion met

`BUILTIN_SCENE_DIGEST` is unchanged — `bd85b8537f0b3e4abf8cf3ad3d36c6abfdceac15355639af2804d58dd9c61931`, bench reports "matches pin". Corpus hash and all raster baselines unchanged. Nothing behavioral moved.

## Names vs identifiers

Renamed: crate directories, package names, workspace members and `workspace.dependencies`, `-p` flags in CI and scripts, module paths, `structure-function-baseline.tsv`, and doc prose describing the family as Pilotage's.

**Not** renamed, because they are identifiers — minted once, hashed or pinned downstream, and not tracking what the crates are called:

- `SCENE_DIGEST_DOMAIN` (`b"pilotage-scene-digest-v1"`) — hashed into every composition digest. Rewriting it moves the digest for zero paint change.
- The corpus provenance record and the sha256 test vector.
- Prose naming the Pilotage cockpit, repository, or browser shell — that is a consumer, which is what the word is for.

`crates/README.md` now records this distinction in a table so the survivors stop reading as an oversight.

## Guardrail

`check-structure.sh` fails on a `pilotage-`-named crate directory **or** package name. Verified it fires on both (a planted `crates/pilotage-probe/` was rejected on both counts).

Also fixed: the downstream-agnostic gate's grep alternation was `(pilotage|avionics|…)`, which after the rename matches none of our own crates — it would have passed **vacuously**. `indicate` is now in the filter alongside `pilotage`, so neither a family crate escaping the allowlist nor a consumer crate reappearing goes unseen.

## Deviation from the issue's plan, and why

The issue's plan says "one PR — `git mv`, mechanical rename, full gate battery green." That turned out to be wrong, and the second commit explains why: the evidence graph pins **content hashes of test sources** and resolves `git show <baseline-commit>:<locator>`. Renaming the directories therefore broke the recorded ATT-01 evidence three ways — changed source blobs, a baseline commit whose tree lacks the new paths, and run records naming packages that no longer exist.

So the three suites were **re-run** and their records re-captured, not re-stamped. The recorded pass counts genuinely move (panel 62 → 77, presentation 165 → 167) because the old records predate tests added since. Both HARD evidence gates report `verdict: VALID`.

## Follow-up this PR requires

Squash-merging orphans the baseline commit this evidence names, exactly as happened before (`baf56e2`, "Evidence: re-anchor the ATT-01 baseline to the merged euler-retirement squash"). **A follow-up must re-anchor to the merged commit**, or main goes red on a HARD gate. I will land that immediately after merge.

## Gates run locally

fmt, clippy `-D warnings`, test `--all-targets`, doc `-D missing_docs`, release build, `thumbv7em-none-eabihf` no_std closure, check-structure, instrument-requirements, certification-claims, standards-registry, trace-report, detect-target, timing (REN-04), glyph pack (REN-02), raster frame hashes (REN-03), instrument-bench, both HARD evidence gates, the exception-expiry pair, and **all 26 negative fixtures still correctly fail**.


luofang34:
## Independent review — findings addressed

An independent reviewer verified the change by running it, not by reading the description.

**Confirmed independently:** `SCENE_DIGEST_DOMAIN` unchanged; scene digest still `bd85b853…`; `corpusSha256` byte-identical to main; every raster baseline unchanged; all three `output-digest` values equal the sha256 of their run record; all three `source-digest` values equal `git hash-object` of their locators; `CFG-WORKTREE`'s declared tree correct. The reviewer also confirmed the moving pass counts are genuine stale-record catch-up — the panels/state source diff versus main is import-only, so 62→77 and 165→167 are tests that already existed and were never re-recorded.

**Fixed in 7557937:**

1. **`check_crate_naming` missed five cases**, each reproduced by the reviewer: TOML literal quotes, a `[[bin]]` name above `[package]`, a `pilotage-` directory under `tools/`, the same in a future `sets/` tier, and `name="x"` without spaces (which aborted the script under `set -e` with no diagnostic). It now parses the `[package]` table with awk over *every* Cargo.toml in the tree. All five verified caught. The `sets/` case mattered most — #9 is sequenced before #13 creates that tier, so the guardrail would have been blind exactly where it was ordered to be ready.

2. **The downstream-agnostic filter had no floor.** A pattern matching nothing reported a closed boundary it never inspected. It now fails when it sees fewer names than the tree has crates — the pre-fix pattern matched **1** against a floor of **13**.

3. **Two false claims removed.** The crate map said recorded run records keep the old name "because a record is a statement about the past" — true only of the records I did *not* re-execute. And both the map and the guardrail comment listed the corpus provenance record as unrenamed while the first commit had renamed it.

   On that rename the reviewer's verdict is that it was **safe and right**: `corpus_sha256` hashes the concatenated case bytes, so the header sits outside it, and the published contract pins `corpusVersion` + `corpusSha256` rather than the file. Naming a generator crate that no longer exists would have been the actual oversight. Caveat now recorded in the crate map: an interpreter that hashes the whole JSON document rather than the published field is the one consumer this disturbs.

4. **The safety baseline's index contradicted everything it indexed** — `docs/instruments/README.md` said Indicate while `fha.md`, `pssa.md`, `intended-functions.md`, `evidence-plan.md`, `standards-applicability.md`, and `standards-registry.toml` still named the subject Pilotage. The phrases naming *this* baseline's subject now name the family; every reference to the Pilotage repository, its decision records, its issues, and the complete Pilotage application stays.

5. **This family's own env knobs** (`PILOTAGE_REPO_ROOT`, `PILOTAGE_INSTRUMENT_*`, `PILOTAGE_STANDARDS_*`) are read and written entirely in-tree and named nothing downstream. They are `INDICATE_*`.

Left alone as cosmetic, per the reviewer's INFO rating: the evidence crate's fixture run records still open with "Pilotage evidence" while their `command:` lines say `indicate-*`. Digests are consistent and all 26 negative fixtures behave.

**Merge note:** this graph anchors to a branch commit, so the squash will orphan it and turn the HARD `--require-resolvable` gate red on main until the follow-up re-anchor. That is the same condition the `baseline-unreachable` fixture exists to enforce, and it has precedent in `baf56e2`. I am landing the re-anchor immediately after this merges.

PR #17, opened 2026-08-07
Evidence: re-anchor the ATT-01 baseline to the merged rename squash

luofang34:
Follow-up to #16, which its own description and merge comment flagged as required.

Squash-merging #16 rewrote the lane commit that `CFG-WORKTREE` and all three `RESULT-*` nodes named, orphaning the baseline. **Main is currently red** on both HARD evidence steps:

```
[placeholder-result] RESULT-RASTER: baseline 458e3d29… is not reachable from HEAD:
a non-ancestor commit is a force-push artifact that a fresh clone cannot fetch;
an unreachable baseline fails closed
```

That is the gate working — the `baseline-unreachable` negative fixture exists to prove it does — and this is the same lifecycle as `baf56e2` ("Evidence: re-anchor the ATT-01 baseline to the merged euler-retirement squash").

## What moves

- `CFG-WORKTREE` commit and tree, and every result's `config-digest` / `run-id`, re-anchor to the merged commit `66494ff`.
- The three run records re-stamp their config identity, so their `output-digest` values follow.
- The evidence crate's two positive fixtures carry the same anchor and move with it.

## What does not

The **`source-digest` values are unchanged** — `39567de1…`, `d1159558…`, `015edf11…`. Nothing was re-run, and nothing claims to have been: only the baseline identity moved, which is exactly what the squash changed.

Precisely scoped, since the trees are not globally identical: the squash preserved **the bound test sources** byte-for-byte, and every *build input* of the three recorded suites — all ten closure crate directories, `Cargo.toml`, `Cargo.lock`, `rust-toolchain.toml`, `.cargo`, `clippy.toml`, `rustfmt.toml` — is identical across it. The paths that do differ between the two trees are docs, CI scripts, evidence data, and the `indicate-evidence` crate itself, none of which any recorded suite builds against. So the re-anchored config identity is sound for the runs it names.

Both HARD gates report `verdict: VALID`. Full test suite, check-structure, and fmt green locally.

### DIFF
--- a/crates/indicate-evidence/tests/fixtures/att01-run-output-baselined.txt
+++ b/crates/indicate-evidence/tests/fixtures/att01-run-output-baselined.txt
@@ -2,9 +2,9 @@ Pilotage evidence — captured verification run (SIM / NOT FOR FLIGHT)
 scope: ATT-01
 suite: fixture band behavior
 command: cargo test -p indicate-instrument-raster
-config-digest: 458e3d29dd03ebb37777585459452288e50698e9
+config-digest: 66494ff9a1b9ccd0eaedd2a48211c027c71f841c
 tool-version: rustc 1.95.0
-run-id: local:458e3d29dd03ebb37777585459452288e50698e9
+run-id: local:66494ff9a1b9ccd0eaedd2a48211c027c71f841c
 outcome: pass
 summary:
 test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
--- a/crates/indicate-evidence/tests/fixtures/attitude-slice.evg
+++ b/crates/indicate-evidence/tests/fixtures/attitude-slice.evg
@@ -17,12 +17,12 @@ attr test band_matches_reference
 
 node RESULT-BAND verification-result
 attr command cargo test -p indicate-instrument-raster
-attr config-digest 458e3d29dd03ebb37777585459452288e50698e9
+attr config-digest 66494ff9a1b9ccd0eaedd2a48211c027c71f841c
 attr tool-version rustc 1.95.0
 attr source-digest git-blob:f399b5de714ad59fc3c75ba830354c8e95e958fd
 attr artifact tests/fixtures/att01-run-output-baselined.txt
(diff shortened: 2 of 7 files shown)


luofang34:
## Independent review — sound, one wording fix applied

A reviewer verified every claim by execution rather than reading, on an isolated worktree.

**Confirmed:** main is genuinely red (three findings, one per result, all baseline-unreachable, and on *both* hard steps — I had said only `--require-resolvable`; the PR body now reflects that). Branch VALID on both. All three `source-digest` values equal `git hash-object` of their locators **and** the blob at `66494ff` **and** the blob at the orphaned `458e3d2` — a `diff` of the source-digest lines between main and this branch is empty, so the "nothing was re-run" claim holds. All four `output-digest`s match their records. `CFG-WORKTREE.commit` is an ancestor of HEAD and its declared tree matches. Every run-record field matches its node attr **and occurs exactly once**, so the cross-check passes for the right reason rather than by shadowing.

**26/26 negative fixtures still fail**, and notably the one edited negative fixture, `unresolved-review-entry.evg`, still fails for its *intended* reason — three `review-incomplete` findings and zero digest/baseline findings. The re-anchor did not mask its failure mode.

**Pass-count honesty re-confirmed** by re-running all three suites on the recorded toolchain: raster 86, state 167, panels 77 — exactly the recorded counts.

**Fixed:** the PR body claimed "the squash preserved file content byte-for-byte", which is false globally — 23 paths differ between the two trees, because #15 and the #16 review follow-ups landed in between. The reviewer proved the claim true for the scope that matters by diffing every build input of the three suites (all ten closure crates plus the toolchain and lint config) and finding it empty. The body now says that precisely instead. No re-run needed; the commit message was already correctly scoped to "the bound test sources".

**Noted as systemic, not this PR's defect:** `indicate-evidence`'s documented ASSURE-04 remedy says to re-record against the rewritten commit *first*, which is impossible under squash-merge — the squash SHA does not exist before the merge. Every evidence-touching squash therefore guarantees a red-main window, and five prior re-anchor commits confirm the pattern. Worth a separate issue; this PR is the established remediation.

PR #18, opened 2026-08-07
Release markers: a changelog keyed to tags, guarded against drift (#8)

### LINKED ISSUE #8: No release markers: every consumer pins a bare revision and reads the log to learn what it contains
This repository has no tags and no changelog. Every consumer pins a bare
40-character revision:

- Pilotage pins the family;
- avionics-link pins seven crates, all at `045ccc9`;
- InstrumentSceneKit vendors the conformance corpus with a provenance record
  naming a revision;
- any further shell will do the same.

Nothing in a bare rev says what it contains. "Which revision has ABI v6?",
"which has corpus v4?", "does the rev my backend was verified against include
the flight director?" are all answerable only by reading the log. That is
already awkward across four repositories and gets worse with a second panel
set, because the interesting question becomes "which revision has the set I
need *and* the corpus my backend passes".

The contract surfaces already version themselves individually — `abi::v6`,
`corpusVersion: 4`, `BUILTIN_SCENE_DIGEST` — which is the hard part and is
done. What is missing is one marker that says which combination a given commit
carries.

## Ask

Annotated tags on the revisions consumers are meant to pin, and a short
`CHANGELOG.md` keyed to them, each entry naming the contract versions in that
release: state ABI, corpus, scene format, composition digest, and whether the
panel set changed.

The cheap version is worth more than the complete one: a tag whose message is
those five values, applied when any of them moves, removes the archaeology
entirely.

## Related

The tracking issue for this work, INST-01, is `luofang34/Pilotage#268`, filed
before the extraction. The panels now live here and Pilotage is a consumer, so
the tracking issue is in the wrong repository — worth moving or mirroring so a
second-set author finds the acceptance criteria where the code is.

luofang34:
Fixes #8.

The issue's own advice was "the cheap version is worth more than the complete one" — a marker naming five values, applied when any of them moves. That is what this is.

## What lands

`CHANGELOG.md`, newest entry first, each naming the five values that decide whether a revision is the one a consumer wants:

| Value | 0.1.0 |
|---|---|
| State ABI | 6 |
| Scene format | 1 |
| Corpus | 4 |
| Composition digest | `bd85b853…` |
| Panel set | `pfd`, `hsi`, `monitor` |

The 0.1.0 entry also tells anyone re-pinning the two things they need to know about this particular release: crate names changed to `indicate-*`, and the composition digest did **not** move across that rename, so a consumer pinning the digest need not re-verify.

## Guardrail

`scripts/check-release-markers.sh`, wired into CI. It extracts the newest entry's five values and compares each against the definition it names (`abi::v6::VERSION`, `SCENE_FORMAT_VERSION`, `corpusVersion`, `BUILTIN_SCENE_DIGEST`, `BUILTIN_PANELS`). A changelog that drifts from its tree answers the pinning question *wrongly*, which is worse than not answering it.

Verified to catch all six failure modes — a wrong value in each of the five rows, and a row deleted outright.

**The guard is deliberately honest about its reach.** It proves entry/tree agreement only. Tags live in the remote and CI clones without them, so it does not and cannot check that a tag exists or points where the entry says; the script prints that caveat rather than implying it checked. Tag discipline stays a human step.

## The tag

`v0.1.0` is annotated on this commit once merged, with the five values in its message. See the note below about pushing it.

## Not done here: moving the tracking issue

The issue also notes that INST-01, the tracking issue, is `luofang34/Pilotage#268` — the wrong repository now that the panels live here. I have not moved or mirrored it: it is a cross-repository change to an issue I did not author, in a repo outside this one, and it is not a code change this PR can carry. Worth doing, but as a deliberate act by someone who owns that tracker.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,14 @@ jobs:
       - name: check-structure
         run: bash scripts/check-structure.sh
 
+      - name: release marker drift guard
+        # Consumers pin a revision and read the changelog to learn what it
+        # carries. An entry that disagrees with the tree answers the
+        # question wrongly, which is worse than not answering it, so the
+        # five contract values are checked against their definitions.
+        # Tag existence is not checked: CI clones without tags.
+        run: bash scripts/check-release-markers.sh
+
       - name: instrument requirement references
         run: bash scripts/check-instrument-requirements.sh
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,11 @@ jobs:
         # carries. An entry that disagrees with the tree answers the
         # question wrongly, which is worse than not answering it, so the
         # five contract values are checked against their definitions.
-        # Tag existence is not checked: CI clones without tags.
+        # Tag existence is not checked, and not because the tags are
+        # absent here — the checkout fetches them. A release tags its own
+        # merge commit, which does not exist while the pull request that
+        # writes the entry is open, so the check would fail every release
+        # on the run that matters. Tagging is a step in CONTRIBUTING.md.
         run: bash scripts/check-release-markers.sh
 
       - name: instrument requirement references
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 Consumers pin this repository by revision. A bare revision says nothing
-about what it contains, so every revision meant to be pinned gets an
+about what it contains, so a revision meant to be pinned is given an
 annotated tag and an entry here naming the contract versions it carries.
 
 Five values decide whether a given revision is the one a consumer wants.
@@ -18,9 +18,11 @@ it was written to remove.
 | Composition digest | `BUILTIN_SCENE_DIGEST` in `indicate-instrument-panels` |
 | Panel set | `BUILTIN_PANELS` in `indicate-instrument-panels` |
 
-A release is cut whenever any of the five moves. Entries are newest
-first, and the tag's message carries the same five values so
-`git show <tag>` answers the question without a checkout.
+A release is cut whenever any of the five moves — review's job, since a
+guard comparing the newest entry to the tree cannot tell an added entry
+from a rewritten one. Entries are newest first, and the tag's message
+repeats the same five values so `git show <tag>` answers the question
+without a checkout. `CONTRIBUTING.md` has the release steps.
 
 ## [0.1.0] — 2026-08-07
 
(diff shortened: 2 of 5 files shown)


luofang34:
## Independent review — two false claims fixed, four guard holes closed

The reviewer confirmed the five stated values are correct (independently read from source, digest reproduced via the bench), that the guard catches real drift in all five rows, that fail-closed behaves, and that the CI step genuinely blocks merge under branch protection.

**Two HIGH findings, both false statements living in the tree:**

1. The prose asserted in the present tense that a pinned rev "carries" an annotated tag and that `git show <tag>` answers the pinning question — while `git tag -l` was empty and `git show v0.1.0` would fail. Reworded, and the release procedure moved out of this PR description into `CONTRIBUTING.md` so the human step survives.

2. The guard justified skipping tag checks with *"CI clones without tags"*. The reviewer pulled the actual checkout log: it fetches `+refs/tags/*:refs/tags/*`. **CI has the tags.** The stated reason was simply wrong. The real reason is timing and it is stronger: a release tags its own merge commit, which does not exist while the PR writing the entry is open, so the check would fail every release on the one run that matters. That reason now appears in the script, in CI, and in the README.

**Four ways the guard could pass while wrong** — each reproduced by the reviewer, each now closed and re-verified:

| Hole | Now |
|---|---|
| `## [Unreleased]` shadowed the real entry, which was never examined | only a versioned heading starts an entry; the entry below is read and drifts |
| a table living only inside a fenced block | 5 × `MISSING` |
| a table living only inside an HTML comment | 5 × `MISSING` |
| a descriptor const containing a digit dropped from the panel set | an added `EGT2_DESCRIPTOR` now reports `DRIFT … the tree says 'pfd, hsi, monitor, egt2'` |
| the ABI reader hardcoded to `v6.rs` | resolves the highest `pub mod vN`; a planted `v7` reports `DRIFT … the tree says '7'` |

The panel set now resolves each constant to the `id` its descriptor declares, which is what a shell selects and a consumer reads — so a constant renamed with its id intact is correctly *not* a panel-set change, and the code matches the comment that always claimed "ids".

Also fixed: an unreadable source is a finding rather than a bare `grep` error killing the script under `set -e`, and the changelog no longer implies the guard enforces "a release is cut whenever any of the five moves". It cannot — the comparison is newest-entry-against-tree, so it can't tell an added entry from a rewritten one. That is review's job and the text now says so.

Left as follow-up, per the reviewer's own ranking: version monotonicity and entry dates are unchecked, and only the newest entry is read.

Issue #7, opened 2026-08-07
No panel-authoring contract: the mirror of backend-contract.md does not exist


luofang34:
`docs/instruments/backend-contract.md` tells a backend author everything they
must honour to interpret the scene stream. There is no mirror for the other
side. Grepping the repository for an authoring path — `docs/`, `crates/README.md`,
`CONTRIBUTING.md` — returns nothing; `CONTRIBUTING.md` covers the toolchain and
the gate battery, and `docs/instruments/README.md` is the safety baseline.

Everything a panel author must get right is currently inferred by reading
`descriptors.rs` and imitating the PFD. That worked while the only panels were
written by the person who wrote the registry. The second set is the first one
that will not be.

## What a panel author has to know, none of it written down

- **The descriptor's obligations** — identity charset, required-layer mask,
  required groups, honest-status regions, config schema, design frame, extreme
  states, and the pinned raster baseline. Which are load-bearing for admission
  and which are advisory is not stated anywhere.
- **How honest-status regions are chosen.** They are the mechanism by which
  admission proves a panel does not paint a value for a withheld group. Getting
  a region wrong makes the check vacuous in exactly the place it matters, and
  nothing explains how to draw the boundary.
- **How to produce the raster baseline** for a new panel, and what a mismatch
  means later.
- **How the composition digest moves** when a panel is added, and who re-pins
  it.
- **What the admission harness actually asserts**, so an author can run it
  before opening a PR rather than discovering the contract from a CI failure.
- **Which symbology is mandatory.** `pilotage-instrument-symbology` carries the
  never-skinnable safety constants; a panel that paints its own flag colour is
  a defect, and the doc that says so does not exist.

## Ask

`docs/instruments/panel-contract.md`, the counterpart to the backend contract,
written before the second set rather than reconstructed from it. The strongest
form is one that ends with a worked example: the smallest panel that passes
admission, with its descriptor annotated field by field.

Blocks the second instrument set under #268 in practice, even though nothing
enforces it.

luofang34:
The architecture plans in #11, #12, and #13 each add a section
this contract must carry, so writing it after their shapes settle avoids
writing it twice:

- **frame constraints and the per-panel content policy** (#12): what
  min/max, aspect bounds, step, and canonical frames mean, how an author
  chooses them, and what admission asserts at and between the pinned sizes;
- **composition obligations** (#11): what `group_regions` accuracy means
  once admission asserts attributed runs land inside their declared regions,
[...]

PR #19, opened 2026-08-07
A registry composition can span two panel provider crates (#6)

### LINKED ISSUE #6: A registry composition cannot span two panel provider crates
`Registry::new` takes `&'static [PanelDescriptor]`
(`crates/pilotage-instrument-registry/src/registry.rs:119`), and
`BUILTIN_PANELS` is a `const` slice inside `pilotage-instrument-panels`
(`descriptors.rs:448`). That is exactly right for one provider crate and has no
answer for two.

The family is `no_std` with no allocator, so a shell cannot concatenate two
`&'static` slices at runtime. A second panel set — an engine page, a six-pack,
a customer suite, a third-party panel — therefore leaves a consumer two
choices:

- **Grow `pilotage-instrument-panels`.** Every set ships to every consumer, the
  crate becomes the enumeration again, and a third-party set cannot exist at
  all.
- **Hand-write the composition.** `const ALL: &[PanelDescriptor] = &[panels::PFD_DESCRIPTOR, panels::HSI_DESCRIPTOR, panels::MONITOR_DESCRIPTOR, engine::EGT_DESCRIPTOR, …]` in each shell — which is the by-hand panel list ADR-0029 removed, reintroduced one layer up. Miss an entry in one shell and that shell silently lacks a panel; the registry cannot detect an omission it was never shown.

Neither is acceptable as the answer for the second set, and the second set is
the first time it is forced.

## What is already right

`scene_digest` takes a `Registry` rather than `BUILTIN_PANELS`
(`digest.rs:58`), so a composed registry already gets its own digest —
the cross-shell identity property in #268 survives composition. `Registry::new`
already rejects duplicate ids, malformed ids, and empty compositions. The gap
is only in *how* a shell names the parts.

## What to decide

How a provider crate contributes descriptors, such that:

- adding a set to a shell is one declaration naming the *set*, not each panel;
- a set can live outside this repository;
- the composition stays `const`-evaluable and alloc-free;
- `Registry::new`'s validation still runs over the whole composition, so
  cross-set duplicate ids are caught at init.

A const builder with a fixed capacity, a declarative macro that flattens
provider slices, or a `PanelSet` descriptor that carries its own slice, are all
plausible. What matters is that the shell names sets and the registry sees
panels.

Blocks the second instrument set under #268.

## Plan

The recommended shape is the **two-level registry** — of the three candidates
the issue names, it is the only one that is alloc-free, macro-free, and keeps
the shell's declaration to one line naming sets.

- `PanelSet { id: &'static str, panels: &'static [PanelDescriptor] }` — set
  ids pass the same charset rule as panel ids; an empty set is an error.
- `Registry::from_sets(sets: &'static [&'static PanelSet]) -> Result<Registry, RegistryError>`.
  `Registry` holds the sets slice, stays `Copy`, and its accessors traverse
  set order then panel order (`panels()` becomes an iterator; `by_id` walks
  sets). `Registry::new(panels)` remains as the single-anonymous-set
  convenience.
- Validation is today's rules run over the flattened traversal — cross-set
  duplicate panel ids fail as `DuplicateId` at init, which is exactly the
  omission-detection property this issue asks for — plus two new variants:
  duplicate set ids and an empty set.
- **Set ids stay out of the scene digest.** The digest already iterates
  panels in shell order (`digest.rs:58-128`); with sets it iterates the
  flattened sequence and nothing else changes. Acceptance criterion:
  `Registry::from_sets(&[&BUILTIN_SET])` reproduces `BUILTIN_SCENE_DIGEST`
  (`bd85b85…`) byte-identically — the proof the mechanism is a pure
  refactor. Keeping set identity out of the digest also means regrouping
  panels into different sets without reordering them is digest-neutral,
  which is right: sets are packaging, not paint.
- The flattened traversal order is contractual for the digest, so the
  shell's set list order is the composition order — documented where the
  slice-order contract lives today.

Alternatives considered and dropped: a const-capacity concat builder leaks a
const-generic `N` into every shell's types and needs the shell to count
panels — the by-hand list again, one layer down; a declarative macro cannot
know a foreign crate's slice length, so it cannot flatten provider slices.
Both fail the "one declaration naming the set" test.

Where set crates live, what they may depend on, and the authoring path are
#13's; the panel-side contract is #7's. One new dependency to record:
screen compositions (#11) validate against the composed registry, so this
mechanism lands before composition admission. Ordering with the rest of the
family: the roadmap in #13.

luofang34:
Fixes #6.

Implements the two-level registry the issue's plan recommends — of its three candidates, the only one that is alloc-free, macro-free, and keeps the shell's declaration to one line naming sets.

## Shape

- `PanelSet { id, panels }` — a named group a provider crate exports. Set ids pass the same charset rule as panel ids.
- `Registry::from_sets(&'static [&'static PanelSet])` composes sets in shell order.
- `Registry::new(panels)` stays as the single-anonymous-set convenience, so a one-provider shell never learns sets exist.
- `Registry` stays `Copy` and allocation-free: the composition is an enum over the two shapes, and `panels()` returns an iterator over the flattened traversal (set order, then panel order). Only 5 call sites existed; 3 were `for` loops that needed no change.

## Validation

Today's per-panel rules run over the flattened traversal, so **a panel id claimed by two sets fails as `DuplicateId` at init** — precisely the omission the registry could not previously detect, because it was never shown the whole composition. Plus four new variants: `DuplicateSetId`, `EmptySet`, `BadSetId`, `NoSets`.

A test also pins that a malformed panel id cannot ride in unchecked merely because it arrived inside a set.

## Acceptance criterion — met

`Registry::from_sets(&[&BUILTIN_SET])` reproduces `BUILTIN_SCENE_DIGEST` (`bd85b853…`) **byte for byte**, which is the proof this is a pure refactor of composition rather than of paint.

A second test pins the corollary the issue argues for: splitting the same panels across two sets *without reordering them* digests identically, because set identity stays out of the digest — sets are packaging, not paint. Regrouping is therefore digest-neutral, while reordering still moves the digest (the pre-existing test covers that).

## Out of scope, deliberately

Where set crates live, what they may depend on, and the authoring path are #13's; the panel-side contract is #7's. Landing them separately keeps each independently revertible.

## Gates

fmt, clippy `-D warnings`, test `--all-targets`, doc `-D missing_docs`, `thumbv7em-none-eabihf` (the mechanism is in the no_std closure), check-structure, check-release-markers, instrument-bench (digest matches pin), and the HARD evidence gate — all green.

### DIFF
--- a/crates/indicate-instrument-panels/src/descriptors.rs
+++ b/crates/indicate-instrument-panels/src/descriptors.rs
@@ -9,7 +9,7 @@
 use indicate_alerts::AlertOutput;
 use indicate_instrument_registry::{
     BackgroundCapability, ConfigBlob, DesignFrame, GroupSet, PanelDescriptor, PanelDrawError,
-    Region,
+    PanelSet, Region,
 };
 use indicate_instrument_scene::{LayerId, SceneWriter};
 use indicate_instrument_state::{
@@ -447,6 +447,17 @@ fn monitor_full_channel() -> AircraftState {
 /// The panels this crate ships, in shell display order.
 pub const BUILTIN_PANELS: &[PanelDescriptor] =
     &[PFD_DESCRIPTOR, HSI_DESCRIPTOR, MONITOR_DESCRIPTOR];
+
+/// The panels this crate ships, as the set a shell names.
+///
+/// A shell composing this crate alongside another provider names sets
+/// rather than panels, so gaining a panel here does not edit any
+/// shell. Set identity stays out of the scene digest, so this composes
+/// to the same digest as the bare slice.
+pub const BUILTIN_SET: PanelSet = PanelSet {
+    id: "builtin",
+    panels: BUILTIN_PANELS,
+};
 
 /// The pinned scene digest over [`BUILTIN_PANELS`] and the canonical
 /// corpus (ADR-0033): the composition contract every build target must
--- a/crates/indicate-instrument-panels/src/descriptors/digest_tests.rs
+++ b/crates/indicate-instrument-panels/src/descriptors/digest_tests.rs
@@ -3,10 +3,10 @@
 use std::string::String;
 use std::vec;
 
-use indicate_instrument_registry::{PanelDescriptor, Registry, scene_digest};
+use indicate_instrument_registry::{PanelDescriptor, PanelSet, Registry, scene_digest};
 use indicate_instrument_scene::MAX_SCENE_BYTES;
 
-use super::{BUILTIN_PANELS, HSI_DESCRIPTOR, PFD_DESCRIPTOR};
+use super::{BUILTIN_PANELS, HSI_DESCRIPTOR, MONITOR_DESCRIPTOR, PFD_DESCRIPTOR};
 
 fn hex(digest: [u8; 32]) -> String {
     const HEX: &[u8; 16] = b"0123456789abcdef";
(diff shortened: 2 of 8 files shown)


luofang34:
## Independent review — sound; three notes addressed

The reviewer verified the acceptance criterion both directions and, notably, ran the bench in a throwaway worktree on `main` and got **byte-identical output** — same digest, same 121 admission cases, same 83 counted warnings. Behavior parity is empirical here, not inferred.

**The iterator was the risky part and it holds up.** The reviewer drove it over hand-built *invalid* compositions (empty sets at front, middle, end, and all-empty) and confirmed termination, no skips, no double-yields, and correct fusion. They also mutation-tested the `self.panel = 0` reset: deleting it fails three of this PR's own tests, so the reset is already guarded. And they confirmed a caller cannot obtain a `Panels` over an invalid registry — `Composition` and the cursor fields are private, and the only path to a `Registry` runs through validation.

**Duplicate detection verified on the hard case**: with sets of size 2, 1, 3, a clash between the *first* panel of set 0 and the *last* panel of set 2 reports `DuplicateId { index: 5 }`, and the `take(index)` prefix property was asserted exhaustively.

**Fixed in 44f505f:**

1. Ten error variants still documented `index` as "Position in the composed slice". With sets there is no composed slice — it is a flattened-composition index, and the docs say so now.
2. `Panels` derived `Copy`, which makes the slice-iterator footgun easy: consume a copy and the original silently stays put, so a half-read composition can look whole. It is `Clone` now, for the reason std's slice iterators are.
3. The unreachable `Empty` guard in `validated` keeps its place but now says why — it is the invariant the per-panel loop rests on, and a third `Composition` shape would otherwise inherit a loop that passes by iterating nothing.

**Noted, not changed:** "Panel set" is now overloaded — the release-marker label in `CHANGELOG.md` means the list of shipped panel ids, while the new type means a provider's named group. I left it: v0.1.0 is already tagged with that wording, and making the tag and changelog disagree is worse than the overload. Worth settling when #13 names the `sets/` tier.

PR #20, opened 2026-08-07
Repository layout: a home for instrument sets, and a tier law with teeth (#13)

### LINKED ISSUE #13: Repository layout: no tier rules and no home for future instrument sets

luofang34:
Fixes #13.

## Tiers, enforced

Three library tiers plus tools, each stating its dependency law, with `check-structure.sh` failing any manifest that reaches outside its own. The issue's claim is that "the dependency law *is* the structure", so it gets the same treatment as the consumer boundary rather than a paragraph asking nicely.

| Tier | Where | May depend on |
|---|---|---|
| Kernel | `crates/` | the kernel only |
| Verification and registry | `crates/` | kernel + this tier; **consumes sets, never a normal dependency of one** |
| Sets | `sets/` | kernel only (normal deps); registry allowed as a **dev**-dependency |
| Tools | `tools/` | unconstrained — shells, not a library tier |

Verified the guards fire, each reproduced: a set taking the registry as a normal dependency, a kernel crate reaching the conformance harness, and a crate map missing a row.

## The law was unsatisfiable as written — and that is the interesting part

Issue #13 says a set crate "depends on the kernel only — never on raster, conformance, or **the registry**". But a set exports a `PanelSet` and is authored against `PanelDescriptor`, and both lived *in* the registry. As written, the law forbade exactly the dependency every set must have.

Rather than quietly weaken the law, the split makes it true: the vocabulary a set is written against moves to a new kernel crate `indicate-instrument-descriptor` (descriptor, config blobs and their key schema, group sets, canonical states, `PanelSet`), and `indicate-instrument-registry` keeps what it is actually for — composing descriptors into a validated registry and deriving the cross-shell digest. The registry re-exports the vocabulary, so **conformance, raster, and the bench are untouched**.

`cargo tree -p indicate-instrument-panels -e normal` now shows alerts, descriptor, scene, state, symbology, libm — no verification-tier crate.

The one concession, stated in the crate map: the registry is allowed as a **dev**-dependency of a set, so a set can pin its own scene digest without standing up a shell. A test-graph edge is not a shipping one.

## One more gap in the issue's tier lists

`indicate-instrument-feeder` appears in neither list the issue gives. It is `no_std`, depends only on the kernel, and nothing in-tree consumes it, so it joins the kernel. Flagging it because it was an omission, not a judgment call I was asked to make.

## Crate map cannot drift

Reorganized by tier, each tier stating its law, and checked both ways: every workspace library crate needs a row, and every row must name a crate that exists.

## Evidence

Moving the panels crate moved three case locators, so `git show <baseline>:<locator>` stopped resolving. The bound sources are byte-identical across the move — only paths changed — so **source digests are unchanged**. The suites were re-run and their records re-captured; the PFD count moves 77 → 79 because the set-composition tests landed after the last record.

Both HARD gates `VALID`; all 26 negative fixtures still fail correctly.

**Follow-up required:** squash-merging orphans this baseline, so a re-anchor to the merged commit must land immediately after, as with #16/#17.

## Acceptance

Scene digest unchanged at `bd85b8537f0b3e4abf8cf3ad3d36c6abfdceac15355639af2804d58dd9c61931` — the acceptance criterion the issue sets for every structural PR here. Corpus and raster baselines untouched.


luofang34:
## Independent review — one HIGH finding fixed; the law now has the teeth the title claims

The reviewer verified the behavioral and evidence axes exhaustively: digest `bd85b853…` reproduced; the corpus **tree hash identical** to main; all three raster baselines identical; the conformance, raster, and tools tree hashes **byte-identical**, confirming they really were not edited; and a compile probe of all 24 pre-split re-exported names plus the `states` module path against the branch rlib, with a negative control, confirming the registry's public surface lost nothing.

They also verified the evidence down to the byte — including that blob `015edf11…` is identical at the *old* path under baseline `66494ff9` and the *new* path — so "byte-identical across the move" was checked, not taken on faith. The 77→79 story checks out too: `#[test]` count is 77 at `66494ff9` and 79 at `08a31a9`, with the +2 arriving from #19 before this PR.

### HIGH — the tier law was a text scanner, and six manifests walked past it

The reviewer constructed six ways to spell a set's dependency on the verification tier that `check-structure.sh` accepted while cargo built them happily:

| Spelling | Was |
|---|---|
| `dep.workspace = true` (dotted key) | missed |
| `"dep" = { … }` (quoted key) | missed |
| `[dependencies.dep]` (section style) | missed |
| `[target.'cfg(unix)'.dependencies]` | missed |
| `judge = { package = "dep", … }` (rename) | missed |
| `[dependencies] # comment` | missed |

The dotted key is the damning one — it is **this repo's own idiom**, written one line above `[dependencies]` in every manifest here. Three of the six silenced scanning for an entire table while still printing `check-structure: OK`.

**Fixed in fddbb04** by asking cargo instead of regexing TOML. `cargo metadata` reports the resolved package name, the dependency kind, and the target a dependency applies to — which closes all six, the `package =` rename in particular, since no key-name regex can see through it.

### MEDIUM — the gate shipped without the selftest this repo requires of its gates

`test-standards-registry.sh` states the house rule, and `check-standards-registry.sh` / `check-instrument-requirements.sh` both follow it. The tier law shipped with none — which is precisely how six bypasses went unnoticed.

`scripts/test-structure.sh` now plants a set crate reaching the verification tier once per spelling and requires a refusal, then plants the two shapes the law *permits* and requires acceptance, so the suite cannot be satisfied by refusing everything. `check-structure.sh` runs it, with a child guard against recursion. 9/9 cases pass.

### Also fixed

- The tier lists now fail when they name a crate that no longer exists — the same drift the crate map guards, left open one file over.
- The downstream-agnostic floor counted only `crates/`, so it undercounted the family from the moment `sets/` was created. It counts both now (`seen=15`, `floor=14`).

### Noted, not changed

The crate map is not checked for *tier placement* — a set crate listed under the Kernel table would pass. Real, and worth a follow-up; the map is checked for existence both ways today.

**Merge sequence:** both HARD gates anchor to `08a31a9`, which the squash orphans, so `main` goes red until the re-anchor lands. Disclosed in the commit message, and I am landing it immediately after, as with #16/#17.

Issue #10, opened 2026-08-07
Settle two vocabulary limits before a second panel set is authored around them


luofang34:
The opcode vocabulary is seventeen entries
(`crates/pilotage-instrument-scene/src/opcode.rs`). Two properties of it decide
what a second panel set can be, and both are worth settling deliberately rather
than discovering while authoring.

**There is no `scale`.** The transform ops are `TRANSLATE` and `ROTATE` only.
A panel is authored in the frame its descriptor declares and the backend maps
that whole frame to the viewport, which is exactly right for a full-screen
panel. It also means an instrument cannot be drawn at a different size *within*
a scene. If a second set wants the airspeed tape at two-thirds size beside an
engine column, or the HSI rose as an inset, the geometry has to be re-authored
rather than reused — the sub-instrument is not a reusable unit, only the whole
panel is.

That may be the right answer. "Panels compose; instruments inside them do not"
is a defensible contract and it keeps the coordinate story simple, which is
worth something on a display path. But if the second set is meant to *rearrange*
existing instruments rather than add unrelated ones, the vocabulary decides it
and nobody has said so out loud.

**`ARC` has no paint mode.** `LINE`, `POLYLINE`, `RECT`, `CIRCLE`, and
`POLYGON` carry a mode byte; `ARC` does not, so it is stroke-only. Engine
instruments are mostly coloured arc bands — green, yellow, red sweeps around a
dial. A thick stroked arc reproduces a band, so this is expressible, but the
band's inner and outer radii are then coupled to the stroke width and the
geometry is expressed sideways.

## Ask

Before the second set, record which of these are contract and which are
oversights:

- Is "no scale" a deliberate limit? If yes, say so in the backend contract and
  in the panel contract (#B), so an author reaching for an inset knows to stop.
  If no, adding `SCALE` is a version bump, three backends, and corpus cases —
  cheapest before more panels exist to re-verify.
- Is stroke-only `ARC` sufficient for the instrument types the second set
  covers? Answering needs the set named, which is why this belongs with its
  scoping rather than after.

Neither is blocking today. Both become expensive the moment a panel set is
authored around the wrong assumption.

## Resolution plan

**The scale question is settled by architecture, not by an opcode.** Two new
issues carry the answer:

- Screen composition (#11) makes the *panel* the unit of placement: an
  HSI inset is a small HSI slot on a composed screen, replayed whole at its
  own frame — no re-authoring, no scene rewriting.
[...]

PR #22, opened 2026-08-07
A panel-authoring contract, and a template set that proves it (#7)

### LINKED ISSUE #7: No panel-authoring contract: the mirror of backend-contract.md does not exist

luofang34:
Fixes #7.

## What lands

`docs/instruments/panel-contract.md` — the mirror of `backend-contract.md` — plus `sets/indicate-instrument-template`, a real crate that is the smallest panel passing admission and **asserts so in its own test**, so the worked example cannot drift from the contract the way a doc snippet would. That is the form #13 argued for: one source of truth instead of a doc listing and a skeleton diverging.

## The part imitation cannot teach

The document's job is the distinction the code cannot show you: **which descriptor fields admission actually enforces**, versus which are validated at registry init, versus which are declared and read by nothing.

`group_regions` is the honest case. It is geometry-validated at init and **admission does not read it** — the conformance crate says so in its own module doc. The contract says so too, rather than implying a check that does not run, and explains what *does* keep a panel honest today (the provenance machinery) and why regions should still be drawn tightly.

The rule easiest to get exactly backwards is stated twice: **a numeral must carry a claim; a dash must not.** Tagging the `---` that stands in for a withheld value claims the group that was withheld, and fails as `FabricatedNumeral` in precisely the case the dashes exist to serve.

Each documented failure mode was verified by mutating the template and reading the error the harness actually produces — `FabricatedNumeral`, `UntaggedNumeral`, `BackgroundContract`, `MissingRequiredLayers`, `ForeignClaim` — not by reading the source and guessing.

## Blind spots, recorded

A contract that lets an author read a green run as more than it is has failed them. Two limits are stated:

- The harness tests **withholding**, not resolved status, so the "source unusable, draw dashes" path is contract-required and harness-invisible. Test it yourself.
- It draws the **empty config**, so no configured behaviour is covered at all — and a numeral derived from configuration is unadmittable by construction.

## The template

One panel, 240×120, its own design frame. Kernel-only normal dependencies (the tier law forbids a set from reaching the machinery that judges it); registry and conformance are dev-dependencies. It declares `required_groups = {Air, Trust}` rather than just `Air`, because its airspeed folds `trust.quality` — so withholding `Trust` genuinely dashes the number, which is the honest mirror of what the shipped panels do.

Admission over it: 4 states × (1 fed + 2 withheld) = 12 cases, zero warnings.

## Not in this PR

The `SCALE` decision is stated on the panel side here because an author reaching for an inset needs it; the backend-contract half and the `ARC` question are #10.

## Acceptance

Scene digest unchanged at `bd85b8537f0b3e4abf8cf3ad3d36c6abfdceac15355639af2804d58dd9c61931` — the template is a set of its own and does not join the shipped composition. Full battery green including the HARD evidence gates.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,7 +96,8 @@ jobs:
           known="indicate-frames indicate-alerts indicate-sha256 \
                  indicate-instrument-scene indicate-instrument-state \
                  indicate-instrument-glyphs indicate-instrument-symbology \
-                 indicate-instrument-panels indicate-instrument-raster \
+                 indicate-instrument-panels indicate-instrument-template \
+                 indicate-instrument-raster \
                  indicate-instrument-conformance indicate-instrument-feeder \
                  indicate-instrument-descriptor indicate-instrument-registry \
                  indicate-evidence instrument-bench"
@@ -150,6 +151,7 @@ jobs:
           -p indicate-instrument-symbology
           -p indicate-instrument-descriptor
           -p indicate-instrument-panels
+          -p indicate-instrument-template
           -p indicate-instrument-raster
           -p indicate-instrument-feeder
           -p indicate-instrument-registry
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -259,6 +259,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "indicate-instrument-template"
+version = "0.1.0"
+dependencies = [
+ "indicate-alerts",
+ "indicate-instrument-conformance",
+ "indicate-instrument-descriptor",
+ "indicate-instrument-registry",
+ "indicate-instrument-scene",
+ "indicate-instrument-state",
+ "indicate-instrument-symbology",
+]
+
+[[package]]
 name = "indicate-sha256"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ members = [
     "crates/indicate-instrument-symbology",
     "crates/indicate-instrument-descriptor",
     "sets/indicate-instrument-panels",
+    "sets/indicate-instrument-template",
     "crates/indicate-instrument-raster",
     "crates/indicate-instrument-conformance",
     "crates/indicate-instrument-feeder",
@@ -38,6 +39,7 @@ indicate-instrument-glyphs = { path = "crates/indicate-instrument-glyphs" }
 indicate-instrument-symbology = { path = "crates/indicate-instrument-symbology" }
 indicate-instrument-descriptor = { path = "crates/indicate-instrument-descriptor" }
 indicate-instrument-panels = { path = "sets/indicate-instrument-panels" }
+indicate-instrument-template = { path = "sets/indicate-instrument-template" }
 indicate-instrument-raster = { path = "crates/indicate-instrument-raster" }
 indicate-instrument-conformance = { path = "crates/indicate-instrument-conformance" }
 indicate-instrument-feeder = { path = "crates/indicate-instrument-feeder" }
--- a/README.md
+++ b/README.md
@@ -35,8 +35,10 @@ repository by git rev and integrate through three published surfaces:
 
 CI enforces the direction: a step fails if the closure ever reaches a
 consumer crate. See [`crates/README.md`](crates/README.md) for the crate
-map and [`docs/instruments/backend-contract.md`](docs/instruments/backend-contract.md)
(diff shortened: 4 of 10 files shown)


luofang34:
## Independent review — a false enforcement claim, caught and fixed

The reviewer fact-checked every enforcement claim in the document against the code and proved eight of them by mutating the template and reading the error the harness actually produces. Most held. **One did not, and it was wrong in the direction a contract must never be.**

### The false claim

The document said the dash-out path was *"harness-invisible"*. It is not. `nothing-fed` and `source-unusable` are two of the four canonical states every panel meets, so the degraded branch is drawn in **9 of the template's 12 cases**, and its output is policed for untagged numerals, glyph coverage, layers, and background. Proof: changing `"---"` to `"-0-"` fails as `UntaggedNumeral { state: "nothing-fed" }`.

Telling authors a well-covered path is untested is exactly the "teaches you to trust the wrong things" failure the document opens by warning against.

There *is* a real limit underneath, and it now says that instead: **nothing asserts a failure cue is painted.** Emptying the dash branch entirely still passes admission — drawing nothing where a dash belongs is admissible, and is a defect. Plus the provenance check fires against withholding rather than resolved status, and alerts join the empty config as a third blind spot (admission and the digest both pass `None`).

### Two overstatements

- The doc said a set may not depend on the conformance harness — while prescribing a test that needs it. The whole verification tier is available as dev-dependencies, and a set needs two of them.
- "Unclipped" was too strong: `covers_frame` accepts a clip that *contains* the frame. The reviewer proved it by wrapping the ground in a containing `clip_rect` and watching it pass. The template's own comment said the stricter, wronger thing — and that is the copy an author starts from.

### Gaps closed

Concrete things a first set hits as red builds: sets are `no_std`; the four workspace registrations an in-repo set needs (members, crate-map row, thumbv7em list, downstream-agnostic allowlist); the glyph vocabulary's name, home, and how little lowercase it holds (`kts` fails — only `k` and `t` are in it); and that the out-of-frame ratchet is an assertion the *author* writes, not something the harness does.

The template is now described as the smallest panel that **exercises the honesty rules**, not the smallest that passes — the reviewer correctly noted a panel claiming nothing passes more easily and teaches less.

### Independently confirmed

`group_regions` really is unread by admission (moving it to a 4×4 corner box still passes); the numeral/dash asymmetry and its ordering; `ConfigClaim`, `ForeignClaim`, `GlyphCoverage`, `BackgroundContract` all reproduce; out-of-frame ink warns rather than refuses; the `{Air, Trust}` claim is exactly right (withholding either dashes the airspeed, withholding `Attitude`/`Nav`/`Heading` does not — so the mask is neither over- nor under-declared); the template is not in `BUILTIN_PANELS`; the digest is unchanged; the tier law holds.

PR #23, opened 2026-08-07
Record which vocabulary limits are contract and which are open (#10)

### LINKED ISSUE #10: Settle two vocabulary limits before a second panel set is authored around them

luofang34:
Fixes #10. Doc-only; no display behavior changes.

The issue asks for a decision on two properties of the opcode vocabulary, recorded before a second set is authored around the wrong assumption. Both follow the resolution the issue itself argues for.

## No `SCALE` — deliberate, and now stated

"Panels compose; instruments inside them do not" is the contract. The issue's own worry — that a set meant to *rearrange* existing instruments is blocked — is answered a layer up rather than by a transform: an instrument wanted at two-thirds size beside another is a finer-grained **panel** placed at its own frame. Reuse by decomposition, not by rescaling a replay. Which finer-grained panels exist is a decision about what a set exports, and costs no vocabulary change.

So an author reaching for an inset finds the composition contract, not a missing opcode. The panel-side half of this landed with #7; this is the backend half.

## `ARC` paint mode — deliberately left open

The issue argues this cannot be decided without naming the set that needs it, and nothing here overrides that. A thick stroked arc does reproduce a coloured band, so engine dials are expressible today; the band's inner and outer radii just end up coupled to the stroke width.

What is recorded is the **price**, so it is known when the day comes:

- A *new* filled annular-band opcode is **not** a scene-format version bump. Verified: an unknown ordinary opcode decodes to `Cmd::Unknown` with its payload skipped (`decode.rs:208`), so older backends degrade gracefully. It *is* a corpus bump, an implementation in every backend, and new admission geometry.
- Changing the existing `ARC` payload is a hard break for every pinned interpreter and is off the table.

## Claims fact-checked before writing

- `Arc` carries no `mode` field while `Circle` does — confirmed in `cmd.rs:216` vs `:227`.
- Unknown ordinary opcodes are counted and skipped — confirmed in `decode.rs:208` and the module doc at `:31`.

Gates green including both HARD evidence gates.

### DIFF
--- a/docs/instruments/backend-contract.md
+++ b/docs/instruments/backend-contract.md
@@ -50,9 +50,10 @@ A panel is authored in the logical frame its descriptor declares
 
 ## Where the vocabulary deliberately stops
 
-Two limits in the opcode set are contract, not oversight. A backend
-author reading the vocabulary and finding them missing should stop
-looking for the version that adds them.
+Two properties of the opcode set decide what an instrument can be. One
+is settled contract, and a backend author who finds it missing should
+stop looking for the version that adds it. The other is genuinely open,
+and is recorded here with its price rather than left to be discovered.
 
 **There is no `SCALE`, and there will not be one.** The transform ops
 are translate and rotate. A panel is authored in the frame its
@@ -69,21 +70,36 @@ vocabulary change. A compositor therefore never resizes a scene by
 rewriting its commands, and `panel-contract.md` tells authors the same
 thing from the other side.
 
-**`ARC` is stroke-only, and that one is genuinely open.** `LINE`,
-`POLYLINE`, `RECT`, `CIRCLE`, and `POLYGON` carry a paint mode; `ARC`
-does not. A thick stroked arc reproduces a coloured band, so an engine
-dial's green/yellow/red sweeps are expressible today, but the band's
-inner and outer radii end up coupled to the stroke width — the geometry
-expressed sideways. Whether that is sufficient cannot be decided without
-the instrument set that needs it, so it stays open.
+**A filled annular band has no opcode, and that one is genuinely open.**
+Paint mode splits the vocabulary by shape, not by oversight: the
+closed-area ops — `RECT`, `CIRCLE`, `POLYGON` — carry a `PaintMode`,
+while the open-path ops — `LINE`, `POLYLINE`, `ARC` — are stroke-only,
+having no interior to fill. `ARC` is not the odd one out.
+
+That consistency is what makes the question a real one. An engine dial's
+green/yellow/red sweeps are expressible today, because a thick stroked
+arc reproduces a band — but the band's inner and outer radii then come
+out of the stroke width and the radius together, which is the geometry
+expressed sideways. A band is an area between two radii, so what it
+wants is not a mode byte on `ARC` — a stroke-only op has nothing to fill
+— but its own geometry. Whether the sideways expression is sufficient
+cannot be decided without the instrument set that needs it, so it stays
+open.
 
 What it would cost, recorded now so the price is known when that day
 comes: a *new* filled annular-band opcode is **not** a scene-format
 version bump, because an unknown ordinary opcode inside a layer is
 counted and skipped, so older backends degrade gracefully. It is a
 corpus bump, an implementation in every backend, and new admission
-geometry. Changing the existing `ARC` payload, by contrast, is a hard
-break for every pinned interpreter and is off the table. The cheap path
+geometry — the background-coverage check counts only the painting
+commands it knows, so a new one is invisible to it until it is added.
+
+Changing the existing `ARC` payload, by contrast, is off the table, and
+for a worse reason than "a hard break": the decoder reads its five
+floats from fixed offsets and ignores trailing bytes, so prepending a
+mode byte would not raise an error on a pinned interpreter. It would
+silently mis-read the centre and radius from shifted bytes and paint the
+wrong arc. Opcodes are append-only. The cheap path
 exists; take it only with the set in hand.
 
 ## Two disciplines every backend must follow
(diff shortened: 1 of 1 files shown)


luofang34:
## Independent review — a false claim about the wire vocabulary, fixed

The reviewer fact-checked every structural claim against the code and confirmed all of them — no `SCALE`, unknown-opcode accept/count/skip semantics (decoder *and* layer validator), no format bump, the corpus/backend/admission costs, and consistency with both sibling contracts. One claim was false.

### HIGH — the paint-mode list was wrong

The section said `LINE`, `POLYLINE`, `RECT`, `CIRCLE`, and `POLYGON` carry a paint mode while `ARC` does not, making `ARC` the odd one out. **Two of those five carry no mode either.** `Cmd::Line` and `Cmd::Polyline` are stroke-only and their own doc comments say "(always stroked)"; the decoder parses their payloads with no leading mode byte, and the encoder takes no `PaintMode` argument for them.

The real split is by shape: closed-area ops (`RECT`, `CIRCLE`, `POLYGON`) carry a `PaintMode`; open-path ops (`LINE`, `POLYLINE`, `ARC`) have no interior to fill. `ARC` is consistent with its kind, not anomalous — which inverts the premise the paragraph leaned on.

Correcting it **strengthens** the conclusion: a band is the area between two radii, so what it wants was never a mode byte on a stroke-only op — it wants its own geometry. "A new opcode, with the set in hand" stands.

The error came in from issue #10's own body rather than being invented here, so I have [corrected the issue text](https://github.com/luofang34/Indicate/issues/10#issuecomment-5221769103) too — otherwise the next author re-imports it.

### MEDIUM — the opener over-claimed

It said both limits were settled while the section itself calls `ARC` "genuinely open" and prices out the version that would close it. It now separates the one settled contract limit from the one open question recorded with its cost, matching the issue's own resolution plan.

### Strengthened on the reviewer's evidence

"Changing the `ARC` payload is a hard break" was **understated**. The decoder reads five floats from fixed offsets and ignores trailing bytes, so a prepended mode byte would raise no error at all — it would silently mis-read centre and radius from shifted bytes and paint the wrong arc. Silent geometry corruption is worse than a break, and the doc now says so.

Also added, since the reviewer verified it is a real part of the price: the background-coverage check enumerates only the painting commands it knows, so a new painting opcode is invisible to it until it is added there.

### Confirmed, not assumed

No `SCALE`, skew, or matrix op anywhere in `opcode.rs`/`cmd.rs`/`encode.rs`/`decode.rs`; unknown opcodes are accepted inside a layer, counted against `MAX_LAYER_COMMANDS`, and reported via `report.unknown`, with the behavior pinned by existing tests; `SCENE_FORMAT_VERSION` is checked only against the leading byte; the raster backend, the conformance outcome encoder, and admission all tolerate an unknown opcode. No contradiction with the design-frame section above it, with `scene-layer-protocol.md`, or with the panel-side half in `panel-contract.md`.

PR #24, opened 2026-08-07
The design frame becomes an emission input (#12)

### LINKED ISSUE #12: The design frame is a constant, so resizing scales the picture instead of showing more instrument
Every shipped panel is authored against one fixed logical frame —
`PANEL_W × PANEL_H = 480×360` (`panels/src/lib.rs:35,38`), hard-coded down to
the horizon translate at 240,180 (`pfd/horizon.rs:46-48`) — and the backend
scales that frame uniformly to its surface (`backend-contract.md`, "The design
frame"). The consequence: a bigger display shows a *bigger picture*, not more
instrument. The tape shows the same knots at 4K as at 480×360; text and
bezels balloon with the window. Panels cannot do better today because they
never see a size: `DrawFn` has no frame parameter
(`registry/src/descriptor.rs:12-17`).

Resizing that shows more instrument means the logical frame itself must vary,
and the panel must lay out against it.

## The model

- `DrawFn` gains the frame: emission becomes a pure function of
  (state, config, alerts, **frame**).
- The shell derives the logical frame from the physical placement at a fixed
  density: one logical unit ≈ constant physical size. A larger placement then
  yields a larger *frame* — more tape range, more monitor lines, wider rose
  margins — while strokes, text, and symbology keep constant physical size.
  The derivation rule (physical extent ÷ density, quantized to the panel's
  step, clamped into its declared range, leftover letterboxed) becomes a
  backend-contract section; backends still never reinterpret geometry.
- Uniform whole-panel scaling stays available as the degenerate case (ask for
  the minimum frame, scale up) — a shell that wants today's behavior loses
  nothing.

The frame is a **draw parameter, not a config key**. Config blobs are
optional 256-byte schema-gated wire blobs, and the admission harness
deliberately draws the empty config — a frame carried as config would make
every size-adaptive panel unadmittable by construction
(`conformance/src/admission.rs`, empty-config discipline). The frame instead
joins state as a first-class input that admission and the digest iterate.

## The restraints every panel must declare

New descriptor fields, validated by `Registry::new`:

- `frame_min` / `frame_max` — per-axis, integer logical units. Min is the
  readability floor: AIR-OUT-004's conspicuity must hold at min, which is
  what min is *for*. Max is bounded by the budgets.
- aspect bounds — the width/height ratio range the layout supports (min/max
  alone admit shapes the layout was never designed for).
- `frame_step` — dimensions are multiples of a quantization step, keeping the
  behavior space and the evidence matrix finite.
- `canonical_frames` — the pinned evidence sizes; must include the smallest
  and largest admissible frames, with every canonical frame validated against
  all the constraints above, aspect bounds included (the per-axis min/max
  corners alone may not be admissible shapes).
- `raster_baseline` becomes per-canonical-frame.
- A content policy per panel in the panel contract (#7): which axis extends
  what — tape range versus tick density, line count, radius versus margin —
  so panels do not answer resize in gratuitously different ways.

## What moves, and how it is kept honest

- **Digest:** the frame joins the tuple. The contract block gains the
  constraint fields; scenes are drawn per canonical state × canonical frame
  (`digest.rs:58-128`). One digest move when this lands, pin-advanced as
  usual.
- **REN-03:** `frame_hashes` pins per canonical frame, still tight-stride
  RGBA8 at 1:1.
- **Admission:** the full case matrix — canonical + extreme states,
  fully-fed + per-group withheld — runs at every canonical frame. Between
  the pins, property checks over sampled in-range frames (host-side; the
  conformance crate may allocate) assert the layer contract, required
  layers, budgets, provenance honesty, and in-frame ink. The overflow
  warning ratchet becomes per-frame; newly declared ranges start at zero
  warnings.
- **Work budget:** the stable invariant is samples-per-pixel (the ~6.4/px
  ceiling in `renderer-verification.md`); the absolute `RenderWork` budget
  derives per canonical frame as ceiling × width × height and gates at each
  one.
- **Config interactions:** `SVS_VIEWPORT` defaults to the full frame and
  validates "inside the frame" (`pfd/panel_config.rs:82-95`); that
  validation gains the frame parameter.

## Migration — no big bang

1. Descriptor fields + the `DrawFn` signature + the digest format land
   together, with every shipped panel declaring the degenerate range
   min = max = canonical = 480×360. Behavior is byte-identical; evidence is
   identical apart from the digest format. One PR, one digest move.
2. Evidence plumbing iterates canonical frames — a no-op at one frame.
   Guardrail: the degenerate-range digest and the unchanged baselines pin
   the refactor.
3. Panels adopt ranges one per PR, each with its layout refactor, canonical
   frames, baselines, admission at every frame, and its content-policy
   section: the monitor first (`LINE_H = 36` fixed — more lines is the whole
   story), then the PFD tapes (vertical extension is range), then the HSI
   (radius versus margin).

## Rejected alternatives, recorded

- **Backend-side non-uniform stretch / nine-patch:** breaks every rotated
  construct (horizon, rose) and violates "backends never reinterpret
  geometry".
- **A `SCALE` opcode:** sizing is a layout decision inside the panel, not a
  transform over its output — see #10 and #11. Placement scaling stays
  the backend's single uniform mapping.

## Open decisions

- Whether the density (logical units per physical unit) is a per-shell choice
  or a contract constant with a tolerance. Leaning per-shell with a
  documented recommended value, since physical DPI varies and the family
  cannot know it.
- The step value (a coarse step like 20 units keeps the canonical matrix
  tiny; a fine step serves arbitrary windows better).

Change control: AIR-BAS-007 quantifies determinism over "display
configuration" — the frame is configuration, and the digest treats it as
such; confirm no intended-function amendment is needed, else amend first
(AIR-BAS-005). AIR-OUT-004 binds `frame_min`.

Ordering: before a second set is authored — #10's warning cuts here: a set
authored against a fixed frame bakes the assumption in. #11 consumes
this in its final phase. Roadmap in #13.

Cites: AIR-BAS-007, AIR-OUT-004, AIR-BAS-005.

luofang34:
Fixes #12.

Implements the issue's model and its migration phases 1–2. Phase 3 — panels adopting real ranges — is deliberately excluded: the issue itself says each is its own change with its own layout refactor, canonical frames, baselines, and admission at every frame.

## The model

Emission becomes a pure function of (state, config, alerts, **frame**). The frame is a **draw parameter, not a config key**, for the reason the issue gives: config blobs are optional schema-gated wire blobs and admission deliberately draws the empty config, so a frame carried as config would make every size-adaptive panel unadmittable by construction.

A descriptor now declares the range it supports rather than one constant:

| Field | Meaning |
|---|---|
| `frame_min` | the readability floor — AIR-OUT-004's conspicuity must hold here, which is what a minimum is *for* |
| `frame_max` | the ceiling the budgets allow |
| `frame_step` | quantization, keeping the behaviour space and evidence matrix finite |
| `aspect_min` / `aspect_max` | shapes the layout supports (min/max alone admit shapes it never was) |
| `canonical_frames` | the sizes evidence is pinned at |
| `raster_baselines` | one baseline per canonical frame |

## Validation — eleven new refusals, each with a test that fires it

The subtle one the issue calls out explicitly: the per-axis corners alone may not be an admissible *shape*, so every canonical frame is checked against range, step grid, **and** aspect bounds, and both ends of the range must be pinned. `a_canonical_frame_outside_the_aspect_bounds_is_refused` uses 600×360 — inside both per-axis corners, a shape 4:3 never declared.

## The digest moved once, and that is the deliberate move

`bd85b853…` → `3efb08c5…`. The per-panel contract block gained the frame constraints, and scenes are drawn per canonical state × canonical frame.

**No paint moved.** All three REN-03 raster frame hashes are byte-identical to `main` — I diffed them directly — which is exactly what separates a format move from a regression. Admission is still 121 cases with the 83-warning ratchet intact, as it must be at one canonical frame per panel. The corpus is untouched.

Re-pinned with the reason, and `CHANGELOG.md` gains a 0.2.0 entry naming all five contract values (the release-marker guard requires it).

## Byte-identical behaviour

Every shipped panel declares the degenerate range — `frame_min == frame_max ==` its old constant, one canonical frame — and the panels take their geometry **from the parameter** rather than the constant. That is what proves the plumbing carries the value rather than the panels ignoring it; at one frame the values are identical, so output is unchanged.

## Evidence

The raster attitude suite's source calls `draw_pfd`, whose signature changed, so its blob moved and its recorded digest stopped matching. All three suites were re-run (86 / 167 / 79 — unchanged counts, as behaviour is unchanged) and re-captured. The other two bound sources are untouched and their digests are unchanged.

Both HARD gates `VALID`; all 26 negative fixtures still fail correctly.

**Follow-up required:** squash-merging orphans this baseline, so a re-anchor to the merged commit lands immediately after, as with #16/#17 and #20/#21.

## Not implemented, and why

The issue sketches deriving the absolute `RenderWork` budget per canonical frame from the samples-per-pixel ceiling. That is a real part of the model but not reachable without a panel that has more than one canonical frame; it belongs with phase 3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,47 @@ from a rewritten one. Entries are newest first, and the tag's message
 repeats the same five values so `git show <tag>` answers the question
 without a checkout. `CONTRIBUTING.md` has the release steps.
 
+## [0.2.0] — 2026-08-07
+
+The design frame becomes an emission input. `DrawFn` gains a
+`DesignFrame` parameter, and `PanelDescriptor` declares the range of
+frames a shell may ask for — `frame_min`, `frame_max`, `frame_step`,
+aspect bounds, and the `canonical_frames` the evidence is pinned at —
+in place of the single `design_frame` constant. `raster_baseline`
+becomes `raster_baselines`, one per canonical frame.
+
+| Value | This release |
+|---|---|
+| State ABI | 6 |
+| Scene format | 1 |
+| Corpus | 4 |
+| Composition digest | `3efb08c55eadadc2b006ee6b006b29e4b3a3f8d4ec3ce1324f401dbc16dc85ca` |
+| Panel set | `pfd`, `hsi`, `monitor` |
+
+Panel set changed since the previous release: no.
+
+### Notes for anyone re-pinning
+
+- The composition digest **moved**, and this is the deliberate move: the
+  digest's per-panel contract block now carries the frame constraints,
+  and scenes are drawn per canonical state × canonical frame. It was
+  `bd85b853…` and is now `3efb08c5…`. No paint changed — the reference
+  rasterizer's REN-03 frame hashes are byte-identical, which is what
+  proves the move is format and not regression.
+- Every shipped panel declares a degenerate range: `frame_min` equals
+  `frame_max` equals 480×360, with one canonical frame. The only frame a
+  conforming shell may ask them for is that one, and behaviour there is
+  unchanged. What the registry refuses is a *declaration* whose bounds,
+  step, aspect, or canonical frames break its rules; deriving a frame
+  for a placement and clamping it into the declared range stays the
+  shell's job.
+- A shell calling `(panel.draw)(…)` passes the frame it chose, between
+  the alerts and the scene writer. A shell that wants exactly today's
+  behaviour asks for `frame_min` and scales, as before.
+- Out-of-repo panel sets must add the new descriptor fields; the
+  registry refuses a composition whose canonical frames do not include
+  both ends of its declared range.
+
 ## [0.1.0] — 2026-08-07
 
 First tagged revision. The contract surfaces already versioned
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -101,6 +101,7 @@ the tag is what a human owes it.
   sources by content digest and its baseline by commit: editing a
   recorded source file requires re-recording that evidence, and history
   rewrites that orphan the baseline commit are forbidden.
-- Panels are authored in the design frame their descriptor declares;
-  unclipped ink past the frame edge is counted and ratcheted by the
-  admission harness — growth is a deliberate decision, not drift.
+- Panels are authored against the frame range their descriptor declares
+  and receive the chosen frame as a draw argument; unclipped ink past
+  that frame's edge is counted and ratcheted per frame by the admission
+  harness — growth is a deliberate decision, not drift.
--- a/crates/indicate-evidence/tests/fixtures/att01-run-output-baselined.txt
+++ b/crates/indicate-evidence/tests/fixtures/att01-run-output-baselined.txt
@@ -2,9 +2,9 @@ Pilotage evidence — captured verification run (SIM / NOT FOR FLIGHT)
 scope: ATT-01
 suite: fixture band behavior
 command: cargo test -p indicate-instrument-raster
-config-digest: f660304b057bec44ce31a7602673bc1e9cd9e6b9
+config-digest: 836371a8dd7fab23231f9a17f461f562cd42433a
 tool-version: rustc 1.95.0
-run-id: local:f660304b057bec44ce31a7602673bc1e9cd9e6b9
+run-id: local:836371a8dd7fab23231f9a17f461f562cd42433a
 outcome: pass
 summary:
 test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
--- a/crates/indicate-evidence/tests/fixtures/attitude-slice.evg
+++ b/crates/indicate-evidence/tests/fixtures/attitude-slice.evg
@@ -17,12 +17,12 @@ attr test band_matches_reference
 
 node RESULT-BAND verification-result
 attr command cargo test -p indicate-instrument-raster
-attr config-digest f660304b057bec44ce31a7602673bc1e9cd9e6b9
+attr config-digest 836371a8dd7fab23231f9a17f461f562cd42433a
 attr tool-version rustc 1.95.0
 attr source-digest git-blob:f399b5de714ad59fc3c75ba830354c8e95e958fd
 attr artifact tests/fixtures/att01-run-output-baselined.txt
-attr output-digest sha256:aaf3c06593b58ffecc8a222015c313ffcabc5de9970ca7e2c1debf6eeaa8a322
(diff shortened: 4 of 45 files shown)


luofang34:
## Independent review — one CI blocker and two validation gaps, all fixed

The reviewer went beyond reading: they checked `main` into a temp worktree, built dumper crates against both trees, and diffed the **actual emitted scene bytes** for every panel × canonical state × extreme state — the digest's exact coverage. **All 18 identical.** So "no paint moved, the digest move is purely format" is proven at the byte level rather than inferred from the raster hashes.

They also confirmed the frame genuinely reaches the panels (no library code references `PANEL_W`/`PANEL_H` any more), that each of the eleven refusals is reached by its own test with ordering asserted rather than assumed (they hand-checked the fixture arithmetic — 520×380 really can only fail the step rule; 600×360 really reaches the aspect rule), that the digest binds the new constraint fields and still excludes set identity, and that every evidence value checks out with counts 86/167/79 reproduced.

### HIGH — a gate that passes on main failed here

`docs/instruments/backend-contract.md` gained a bare `AIR-OUT-004` mention. `check-instrument-requirements.sh` requires the linked form, so **CI would have gone red**. I reproduced it and it now carries its link.

### MEDIUM — two rules missing from the family this PR installs

Both reproduced by the reviewer as accepted-but-wrong declarations, both now refusals with must-fail tests:

- **A repeated `canonical_frames` entry** digested the same scene twice and ran the whole admission matrix again for it, doubling `report.cases` and every `FrameOverflow` warning for identical behavior.
- **Two `raster_baselines` at one frame** were both accepted, but the lookup takes the first match — so the second was dead code disagreeing with the live one about what is pinned. The exactly-one rule existed only in a test that iterates `BUILTIN_PANELS`, which a third-party set escapes entirely.

The repo's own rule is that a fix ships with its guardrail, and these belong to precisely the rule family this change installs.

### Contract over-claim, corrected

The backend contract said a refused declaration means "a range naming frames its own panel could not lay out against never reaches a shell". Only *canonical* frames are aspect-checked, so a valid range may still contain an in-range, on-grid frame that violates the aspect bounds. It says that plainly now — a shell told the range is fully policed would not clamp, and clamping is the shell's job.

### Noted, not changed

A NaN canonical frame is refused, but as `CanonicalFrameOutOfRange` rather than a degenerate-value variant (all float comparisons being false) — refused, just mislabelled. Baseline hex strings are not format-checked by the registry. And no test draws a *shipped* panel at a non-canonical frame, which is inherent at a degenerate range and is what phase 3 changes.

Issue #11, opened 2026-08-07
No screen-composition contract: panels cannot mix, inset, or overlap on one display


luofang34:
One registry, many shells, and still only one way to put panels on a display:
one panel per surface. Every consumer iterates the registry and renders each
panel into its own framebuffer (`tools/instrument-bench/src/main.rs:112-119`);
nothing specifies how a shell places several panels on one screen — side by
side, as an inset, or overlapping. So each shell would invent its own layout,
cross-shell identity would stop at the panel edge, and the compositor
obligations of AIR-HAZ-004 would be met ad hoc or not at all.

This issue specifies **screen composition as a contract above the scene IR**:
whole validated panel scenes, placed and stacked by declaration, with bounded
cost. No new opcode, no stream surgery. ("Composition" here is screen layout —
distinct from #6's composition of provider crates into a registry, which this
depends on.)

## Two walls the code already sets

- **Scene streams cannot be merged.** Each layer appears at most once per
  scene, strictly ascending
  (`crates/pilotage-instrument-scene/src/layer.rs:286-328`). Two panels'
  streams cannot concatenate into one legal scene, so cross-panel *band*
  interleaving — panel X's Annunciation painting above overlay panel Y's Tapes
  inside Y's rect — is rejected by construction. The unit of stacking is the
  slot. If finer interleaving is ever wanted, that is a scene-format change
  and a different issue.
- **There is no `SCALE` opcode** (`opcode.rs:7-23`), and that stays true
  (#10). A compositor cannot resize a scene by rewriting commands; slot sizing
  comes from the frame the panel is asked to emit (#12), never from
  geometry manipulation.

## The contract

A `CompositionDescriptor` in the registry crate:

- `screen: DesignFrame` — the logical screen space, under the same units
  discipline as panel frames;
- `slots: &'static [Slot]` where
  `Slot { panel: &'static str, rect: Region /* screen units */, occludes: &'static [&'static str] }`.
  **Slot index is z**, exactly as `LayerId` discriminant is z within a scene:
  later slots paint above earlier ones, and the declaration order is the
  paint order everywhere.

The backend-contract addition is one rule: paint slots in index order; per
slot, clip to the slot rect, translate to the slot origin, replay that panel's
validated scene; one global uniform mapping from screen-logical to the
surface. No offscreen surfaces, no retained graph, no per-slot rotation. A
slot that fails at runtime paints its failure presentation inside its own rect
and nowhere else.

[...]

PR #26, opened 2026-08-07
Grant the composed multi-panel surface an intended function (#11)

luofang34:
Part of #11 — its **step zero**, which the issue mandates land before any code: "AIR-HAZ-004 constrains a compositor's *failure behavior*; it does not by itself grant the composed-display intended function. If a new requirement is needed, amending `requirements.md` is this plan's first PR, before any code (AIR-BAS-005)."

Doc-only. No display behavior changes.

## The decision: a new requirement is needed

I checked the candidates the issue names, and none covers it:

- **AIR-OUT-001..004** each name a *single function's* surface — the PFD, the HSI, conventional instruments — and none contemplates several panels sharing one screen.
- **AIR-OUT-005..010** likewise scope SVS and HUD outputs.
- **AIR-HAZ-004** constrains a compositor's *failure* behaviour ("shall resolve toward exposing higher-criticality bands") but does not grant the function whose failures it governs.
- **AIR-BAS-007** makes the composition part of "display configuration" for determinism purposes, without saying a composition may exist at all.

## AIR-OUT-011 — what it grants and bounds

It grants the surface and then constrains it in the four ways the issue's validation rules will implement, so the code has a requirement to cite rather than the requirement being written to match the code:

- placement is **declared**, not computed at draw time, and a bad declaration is **refused before first paint** — a region off the screen, a panel asked for a size it does not support, a panel wholly buried under opaque panels above it, or a panel above obscuring the criticality content of one below without that obscuration being declared;
- one snapshot and one alert state per composed frame, so two panels showing the same quantity cannot disagree (the misleading-display concern AIR-BAS-007 speaks to);
- a run-time panel failure is presented **inside its own region** and suppresses no other panel;
- composition is placement only — no panel's geometry is reinterpreted, rewritten, or rescaled to fit. That is the same line the deliberately absent `SCALE` opcode draws from the other direction (#10).

## Traceability

The requirements gate enforces that every requirement is referenced, so AIR-OUT-011 is traced from `intended-functions.md` (the function matrix — what it is, its inputs, its failure presentation) and from `pssa.md`'s compositor row (the hazards it answers to). `check-instrument-requirements.sh` now reports 73 unique requirements, up from 72.

## Next

With this landed, the code phases follow: `group_regions` gaining teeth in admission, derived criticality bands, then `CompositionDescriptor` + `validate_composition` + the screen-composition digest.


luofang34:
## Independent review — sound and correctly sequenced; four gaps closed

The reviewer independently read AIR-OUT-001..010, AIR-BAS-007, AIR-HAZ-004, AIR-HAZ-009, AIR-MODE-004 and AIR-FLAG-007 and grepped the registry for composition/panel/screen/layout/overlay terms, and confirmed the gap is real: **AIR-HAZ-009 even presumes** an extra "instrument, panel, or surface" may be drawn, but only denies it independence credit — it grants nothing. So this is not churn to a safety baseline.

### The finding that mattered most

**The declared-occlusion clause had no floor.** As written, declaring an obscuration legalised permanently hiding a lower panel's warnings and failure indications — the exact outcome AIR-HAZ-004 forbids a compositor *fault* from producing, now reachable by design. Issue #11 wrote it as a "**reviewed**, visible declaration"; the requirement had kept only "declared".

Criticality content is now defined, and declaration covers ordinary symbology **only** — never warnings, never failure indications, never the simulation labelling.

That definition also closes a hole the reviewer spotted and I had missed entirely: AIR-BAS-001 and AIR-FLAG-007 require every surface to identify itself as SIM / NOT FOR FLIGHT, "rendered by the applicable surface itself". A composed screen paints nothing of its own, so the label comes from a slot — and nothing stopped a declared overlay from covering the only one.

### Also fixed

- **`intended-functions.md`'s "Source assumptions" cell contained no source assumption** — it restated validation behaviour already in two other cells. Every sibling row uses that column for the caveat that keeps the function honest. It now carries the one this function needs: co-locating panels establishes no independence, because every slot resolves from the same snapshot. A composed six-pack is precisely the false-independence hazard `fha.md` already names as FC-INST-01.
- **The function matrix asserted "every panel's own modes are unchanged by being composed"** while the requirement said nothing about modes. That belonged in the requirement, and is now in it.
- **The no-rescale rule read absolutely** and could be taken as forbidding a backend from scaling to serve a larger surface, which `backend-contract.md` explicitly permits. It now says the single uniform screen-to-surface mapping is unaffected.
- **The PSSA traceability was one-directional** — `pssa.md` states that the assessment cites implementation issues back, and #11 appeared nowhere in `docs/`. The compositor row cites it now.

### Noted for the phases that follow

The reviewer observed that FC-CMP-01–03 in `fha.md` are *within-scene* layer faults; the FHA has no failure condition for multi-panel composition (dead slot, undeclared occlusion, cross-slot snapshot divergence). Not a defect in a doc-only step zero, but an FHA amendment should precede the validation phase.

Also flagged, not changed: AIR-OUT-011 bundles several obligations under one ID, which limits traceability granularity later; and "composition" now means two things across sibling docs (this sense, and #6's registry composition).

Gates: 73 unique requirements, certification-claims clean, both HARD evidence gates VALID.

PR #27, opened 2026-08-07
Screen composition: panels mix, inset, and overlap by declaration (#11)

### LINKED ISSUE #11: No screen-composition contract: panels cannot mix, inset, or overlap on one display

luofang34:
Fixes #11. Implements phases 2–6, folds in phase 7 (available now that #12 landed); phase 1, the requirement, landed separately in #26 as the issue mandates.

## The contract

`CompositionDescriptor { screen, slots }` with `Slot { panel, rect, occludes }`. **Slot index is z**, exactly as a `LayerId` discriminant is z within a scene, so declaration order is paint order everywhere. No new opcode, no stream surgery — whole validated panel scenes, placed and stacked.

`validate_composition` is alloc-free and refuses at init. **Eleven variants, each with a test that fires it**: unregistered panel, degenerate rect, off-screen rect, a slot sized to a frame its panel does not support (phase 7), too many slots, a dead slot buried under the opaque slots above it, an undeclared occlusion, an unmeasured criticality band, and three shape/bounds refusals.

## The AIR-OUT-011 floor, and how it is enforced

The requirement is stricter than the issue draft: a declared obscuration may cover ordinary symbology but **never** warnings, failure indications, or the simulation labelling. That is enforced structurally rather than by prose — **the criticality check runs before the `occludes` list is ever consulted**, so no declaration can reach it. `declaring_the_occlusion_does_not_permit_covering_a_warning` and `declaring_the_occlusion_permits_covering_a_readout` use byte-identical declarations and differ only in what lies beneath.

Two supporting choices make the floor sound: the band is **measured from emitted ink, not declared** (a panel able to name its own warning surface could understate it), and a panel with **no** measured band is refused rather than assumed quiet.

## The finding that changed the design

Phase 2 as drafted — "every run attributed to group G lands inside G's declared region" — was implemented, measured, and **rejected**, because it refuses all three shipped panels (HSI `Heading` by 299 units, PFD `Air` by 159).

It is not a panel bug. Provenance *totality* forces every ladder rung to carry the group it measures, while `group_regions` were authored as the readout value's own box — the PFD descriptor says so outright ("the scale ladder's runs anchor at x 70 and stay outside"), and `panel-contract.md` instructs authors to draw them that way. The rule contradicted both. No region was widened and no panel changed.

**The rule is non-vacuity instead**: a declared region must hold a claimed run of its group. That is the failure worth refusing — a region pointing at empty space would make composition guard a surface where the readout is *not*. `a_region_merely_grazed_by_claimed_ink_is_refused` proves it is stronger than bare overlap. The rejected alternative and the reason are written down in the module and in the contract, not discarded.

## Evidence the contract is implementable

The reference rasterizer composes: one surface, slots in index order, each clipped and translated. **Placement was generalized in place, not forked** — the single-panel path took a placement and rendering one panel became its whole-surface case, so a slot inherits the identical snapping and clip bound. The three REN-03 hashes were re-asserted immediately after that extraction, before anything else moved.

Three composed frames pinned (side-by-side, opaque inset, `NotUsed` overlay), each rendered twice and asserted bit-identical before comparison.

Hashes prove reproducibility, not correctness, so two **property** tests carry the load: the overlay test renders the overlay alone to learn its own ink, then checks *every* pixel of the rect — where the overlay paints nothing the composed frame equals the panel beneath exactly, where it paints it is its own ink, both counts asserted non-zero so neither half passes vacuously. A wrong translate or clip fails that; a hash would have pinned the bug.

## Identity

A screen-composition digest with its own domain, covering the screen frame, the ordered slots with rects and occlusion declarations, and the scene digest beneath. Slot rects are in it from day one. The bench reproduces it: `071bd35c…`.

Per-slot configuration is **shell-supplied at draw time** and therefore not in the digest — the descriptor declares layout, the feed supplies data.

## Nothing previously pinned moved

Scene digest `3efb08c5…`, all three REN-03 panel hashes, and the corpus are unchanged — verified by diffing the pinned literals against `main`. Admission is still 121 cases / 83 warnings.

## Open, and stated rather than implied

`MAX_COMPOSITION_SLOTS = 8` is a declared ceiling, not a measured one; the issue asks for a full-screen six-pack bench before pinning it. Composed `RenderWork` is not gated against a budget that prices one panel. A slot fault spoils the whole reference frame, because a bit-exact harness painting a partial frame would pin a hash of undefined content; the in-rect failure presentation is a runtime obligation of the shell.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,11 +54,26 @@ worth knowing before the next one is cut.
   `Annunciation`/`Failure` ink bound per panel × canonical frame, which
   a composition validates obscuration against. The monitor's is `None`,
   which is a measurement rather than an omission.
+- **The criticality band now folds in the alert stack, and this is a
+  safety fix.** Admission drew every case with no alerts, so the
+  measured `Annunciation`/`Failure` bound excluded the shared alert
+  stack entirely — while a composed frame fans one `AlertOutput` to
+  every slot. A declared obscuration could therefore cover warning rows,
+  which is exactly what AIR-OUT-011 forbids and what the contract says
+  is impossible. The case matrix gains an alert axis (each case drawn
+  quiet and with a saturated stack), so admission runs 242 cases instead
+  of 121 and counts 166 frame-overflow warnings instead of 83. All three
+  `BUILTIN_CRITICALITY_BANDS` entries moved; the monitor's is no longer
+  `None`.
+- **An unwitnessed band refuses obscuration.** A band measured empty is
+  the absence of a witness, not a proof of absence, and it now refuses
+  any overlap (`CriticalityUnwitnessed`) instead of admitting it.
 - **New pinned data: three composed-frame hashes.** The reference
   rasterizer gains `render_composition`, which paints a validated
   composition into one framebuffer, and pins REN-03-style hashes for a
   side-by-side screen, an opaque inset over a PFD, and a `NotUsed`
-  overlay. The overlay's show-through is asserted as a property, not
+  overlay. The inset and overlay fixtures moved above the PFD's band
+  once alerts were folded in, so two of the three hashes are new. The overlay's show-through is asserted as a property, not
   only as a hash. `indicate-instrument-raster` gains normal
   dependencies on the registry, the state crate, and the alert model;
   the arrow points that way and no crate gained a dependency on the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,39 @@ from a rewritten one. Entries are newest first, and the tag's message
 repeats the same five values so `git show <tag>` answers the question
 without a checkout. `CONTRIBUTING.md` has the release steps.
 
+## [Unreleased]
+
+Screen composition ([`AIR-OUT-011`](docs/instruments/requirements.md)).
+None of the five values above moved, so this is not a release — but a
+consumer gains a sixth pinnable value and two new refusals, and both are
+worth knowing before the next one is cut.
+
+- **New pinnable value: the screen-composition digest.** Its own domain
+  string (`pilotage-screen-composition-digest-v1`), covering the screen
+  frame, the ordered slots (panel id, rect, `occludes`), and the scene
+  digest beneath. `tools/instrument-bench` composes a fixture screen and
+  reproduces `071bd35c…`. Per-slot configuration is shell-supplied at
+  draw time and is deliberately not in it.
+- **New ceiling: `MAX_COMPOSITION_SLOTS` = 8.** Every composed-frame
+  budget is a sum over slots, so this is what makes those sums finite.
+  The value is a declared ceiling, not a measured one: no full-screen
+  six-pack has been benched against it.
+- **`group_regions` became load-bearing.** Admission asserts
+  non-vacuity: every declared region must be populated by a visible run
+  claiming its group, somewhere in the panel's case matrix, at
+  `frame_min`. A region over blank space fails as `GroupRegionEmpty`.
+  What is deliberately *not* asserted is that all of a group's claimed
+  ink sits inside its regions — a numeral must carry a claim, so every
+  ladder rung and compass tick carries its group's, and those sit
+  outside the readout box by design. All three shipped panels satisfy
+  the rule as authored; no region and no panel changed.
+- **New pinned data: `BUILTIN_CRITICALITY_BANDS`.** The measured
+  `Annunciation`/`Failure` ink bound per panel × canonical frame, which
+  a composition validates obscuration against. The monitor's is `None`,
+  which is a measurement rather than an omission.
+- The scene digest, the corpus, and every REN-03 raster frame hash are
+  unchanged.
+
 ## [0.2.0] — 2026-08-07
 
 The design frame becomes an emission input. `DrawFn` gains a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,8 +54,17 @@ worth knowing before the next one is cut.
   `Annunciation`/`Failure` ink bound per panel × canonical frame, which
   a composition validates obscuration against. The monitor's is `None`,
   which is a measurement rather than an omission.
-- The scene digest, the corpus, and every REN-03 raster frame hash are
-  unchanged.
+- **New pinned data: three composed-frame hashes.** The reference
+  rasterizer gains `render_composition`, which paints a validated
+  composition into one framebuffer, and pins REN-03-style hashes for a
+  side-by-side screen, an opaque inset over a PFD, and a `NotUsed`
+  overlay. The overlay's show-through is asserted as a property, not
+  only as a hash. `indicate-instrument-raster` gains normal
+  dependencies on the registry, the state crate, and the alert model;
+  the arrow points that way and no crate gained a dependency on the
+  rasterizer.
+- The scene digest, the corpus, the screen-composition digest, and every
+  REN-03 per-panel frame hash are unchanged.
 
 ## [0.2.0] — 2026-08-07
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -208,6 +208,7 @@ dependencies = [
 name = "indicate-instrument-raster"
 version = "0.1.0"
 dependencies = [
+ "indicate-alerts",
  "indicate-instrument-glyphs",
  "indicate-instrument-panels",
  "indicate-instrument-registry",
--- a/crates/README.md
+++ b/crates/README.md
@@ -78,7 +78,7 @@ normal dependency of one.**
 | Crate | Role |
 |---|---|
 | `indicate-instrument-registry` | Composition of descriptors into a validated registry, and the cross-shell scene digest. |
-| `indicate-instrument-raster` | Reference software rasterizer: bit-exact pinned frame hashes, corpus authorship, target-timing evidence (`evidence/`). |
+| `indicate-instrument-raster` | Reference software rasterizer: bit-exact pinned frame hashes, the composed-screen harness, corpus authorship, target-timing evidence (`evidence/`). |
 | `indicate-instrument-conformance` | Panel admission harness (host-side, allocates; deliberately outside the `no_std` closure). |
 | `indicate-evidence` | Standard-neutral lifecycle evidence graph and gate (`evidence-gate` binary); guards `docs/instruments/evidence-graph.evg`. |
 
--- a/crates/README.md
+++ b/crates/README.md
@@ -19,9 +19,14 @@ hashed or pinned by someone downstream, and do not track what the crates
 are called. Rewriting one buys a tidier grep and costs every consumer a
 re-pin for no change in what is painted.
 
-| Survivor | Why it stays |
+A new identifier of that kind is minted in the same scheme rather than a
+fresh one, so the table below is the whole convention and not a list of
+exceptions to it.
+
+| String | Why it reads `pilotage` |
 |---|---|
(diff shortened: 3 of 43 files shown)


luofang34:
## Independent review — one CRITICAL defect, measured and fixed

The reviewer did not take the floor on trust. They built the compositions, fed alerts, and counted pixels.

### CRITICAL — "Declaring buys you readouts, never warnings" was false

Admission draws every case with `alerts: None`, so the measured criticality band never contained the **alert stack** — while `render_composition` fans one alert output to every slot at runtime, exactly as AIR-OUT-011 requires. The stack sits below every band measured without it.

Measured, not argued: `pfd` + `monitor` with `occludes: ["pfd"]` was **admitted**, and rendering it with one Warning-class alert took the PFD's alert-red pixels from **262 to 0**. Worse, **this branch's own pinned evidence fixture** placed an opaque inset at (140,296,200×60) with `occludes: &[]` and asserted it admissible — its comment said it sat "below the band", and that strip *is* the alert stack. With alerts fed: 262 red px → 124.

**Fixed.** Admission gains an **alerts axis**: every case is drawn twice, once quiet and once with a saturated stack chosen to bound every other alert output — a faulted alerting path (the only way to reach the top row), more alerts than the stack shows, and a warning among them.

Both of the reviewer's experiments were first reproduced against the unfixed tree, then kept as permanent tests. Both now refuse with `CriticalityObscured`, and the first also asserts the PFD really paints 262 warning pixels so it cannot pass by the alert never firing.

Counts moved and the movement is explained rather than absorbed: **121 → 242 cases, 83 → 166 warnings**, exactly doubled, because the two overhanging runs are the groundspeed and baro readouts, which the stack does not touch, so each overhangs on both sides of the axis. Every band now reaches the stack's bottom row, and `the_pinned_bands_contain_the_alert_stack` asserts each pin contains the stack extent **measured from rendered pixels**, so a pin cannot drift from the ink again.

### HIGH — `band: None` admitted obscuration instead of refusing it

The monitor was pinned `None` while drawing a full-frame failure X; the reviewer showed every pixel of it below a covering slot was coverable by declaration. An empty measurement is the **absence of a witness, not a proof of absence** — an unwitnessed band now fails closed (`CriticalityUnwitnessed`), and the monitor is genuinely measured now since it draws the stack. Its own channel-gated flag stays unwitnessed, recorded as an author obligation rather than left implied.

### MEDIUM — the labelling clause promised coverage we do not have

No shipped panel emits SIM / NOT FOR FLIGHT. The prose now says the floor is exactly the two bands and no wider, and that such labelling is protected only if the panel paints it into one — obligation named, clause not silently dropped.

### The fixtures moved rather than the rule bending

With the PFD band now spanning y 38..352, the inset fixture is correctly inadmissible. It moved to the narrow strip above the band and re-pinned. That a PFD-sized panel now has very little surface a slot may sit on **is the floor working**, and the fixture comment says so.

### Confirmed sound by the reviewer

The ordering claim held under attack — empty `occludes`, self-reference, a panel in two slots, an occluder itself occluded, an unpinned frame: no bypass. Ink maths correct (half-stroke inflation, conservative arc bounds, transforms, clips). The rasterizer refactor is genuinely shared — only `paint_at` paints. Robust against 64×48 surfaces under a 480×660 screen and 8 coincident slots, no panics, every write bounds-checked. Scene digest, three REN-03 hashes, and corpus byte-identical to main. Tier law holds; nothing depends on raster.

Both LOWs taken: `wrapping_add` for the coverage counter, and `Cmd::Unknown` now documents why it is safe today and what makes it unsafe as the append-only opcode set grows.

PR #28, opened 2026-08-08
Cut 0.2.0: the changelog said a release was due and none was tagged

luofang34:
Housekeeping that closes a self-inconsistency, not a new feature.

## Why

The composition digest moved when the design frame became an emission input (#12). That is one of the five values whose movement cuts a release, `CHANGELOG.md` has carried a written `0.2.0` entry ever since — and `git tag -l` shows only `v0.1.0`. So the repository was asserting a discipline it had not followed, which is exactly the drift these entries exist to prevent.

## What changed

- The screen-composition work (#11) sat under `[Unreleased]` while shipping in the **same revision** as the 0.2.0 entry above it. Tagging would have made the tag disagree with the tree the moment it was applied, so it folds into 0.2.0, where it actually ships.
- One statement in it had gone stale: it said the monitor's criticality band is `None`, "a measurement rather than an omission". The safety fix in #11 made it genuinely measured, and an unwitnessed band now *refuses* obscuration rather than permitting it.
- The alerts axis and its cost are recorded, because a consumer re-pinning across this release will see admission go from 121 cases to 242 and counted warnings from 83 to 166, and should know that is an added axis rather than a regression.

`check-release-markers.sh` still validates the 0.2.0 entry against the tree: `OK (0.2.0; ABI 6, scene 1, corpus 4, panels: pfd, hsi, monitor)`.

## After merge

I will tag `v0.2.0` on the merge commit with the five values in its message, per the release steps in `CONTRIBUTING.md` — the step that is deliberately manual because a release tags its own merge commit, which does not exist while this PR is open.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,12 +24,30 @@ from a rewritten one. Entries are newest first, and the tag's message
 repeats the same five values so `git show <tag>` answers the question
 without a checkout. `CONTRIBUTING.md` has the release steps.
 
-## [Unreleased]
+## [0.2.0] — 2026-08-07
 
-Screen composition ([`AIR-OUT-011`](docs/instruments/requirements.md)).
-None of the five values above moved, so this is not a release — but a
-consumer gains a sixth pinnable value and two new refusals, and both are
-worth knowing before the next one is cut.
+The design frame becomes an emission input. `DrawFn` gains a
+`DesignFrame` parameter, and `PanelDescriptor` declares the range of
+frames a shell may ask for — `frame_min`, `frame_max`, `frame_step`,
+aspect bounds, and the `canonical_frames` the evidence is pinned at —
+in place of the single `design_frame` constant. `raster_baseline`
+becomes `raster_baselines`, one per canonical frame.
+
+| Value | This release |
+|---|---|
+| State ABI | 6 |
+| Scene format | 1 |
+| Corpus | 4 |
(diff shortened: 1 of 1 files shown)


luofang34:
## Independent review — two duplications my own fold introduced

The reviewer verified all five contract values against the tree, every folded-in claim (`071bd35c…` reproduced by the bench, `MAX_COMPOSITION_SLOTS` = 8, 242/166, non-vacuity enforced and containment explicitly not, unwitnessed bands refusing, the monitor's band no longer `None`), and confirmed nothing else in the file is stale — including the 0.1.0 entry's claim that the rename left the digest untouched, which still holds because `SCENE_DIGEST_DOMAIN` was deliberately not renamed.

They also noted this fix was needed: on `main` the file contradicted itself, one bullet saying the monitor's band is `None` and another saying it is no longer `None`.

**Fixed in 0e91e21**, both introduced by my fold:

- The alerts-axis bullet repeated the numbers and the reason from a bullet twenty lines above, and landed *after* the section's closing summary. It also attached the warnings explanation to the case count — cases double because the axis has two values; the warnings double because the two overhanging readouts sit on both sides of it. Removed; the earlier telling stands.
- The sentence I added to the criticality-bands bullet was the unwitnessed-band bullet again minus the error name. It only had to stop being stale about the monitor, so that is all it does now.

Also caught while fixing those: the summary line claimed the screen-composition digest was **unchanged** in the release that introduces it. Corrected.

PR #29, opened 2026-08-08
Correct a rule the layer-protocol table's own rows disprove (#14)

luofang34:
Follow-up to #15. Two independent reviewers of that PR reported after it had merged, and **both flagged the same defect I shipped in it** — with matching file:line evidence, arrived at separately.

## The defect

The paragraph #15 added beside the required-profiles table said:

> A panel requires a band when it opens that band on *every* frame, not when the band always carries content.

Two of the three rows directly above it disprove it. `draw_hsi` and `draw_monitor` both open `Background` as the **first statement** of the function — no branch, no early return — while both rows list `Background` as optional, and **no shipped mask contains it at all** (verified: zero occurrences of `layer_bit(LayerId::Background)`).

And it is not a near-miss a cedeable band would explain away. Both declare `BackgroundCapability::Opaque`, and admission *refuses* an `Opaque` panel that does not lay a full-frame opaque cover in that band in every case. So for those two the band is mandatory by a **stronger** mechanism than the mask, while the mask calls it optional.

This is the drift class #14 existed to kill, reintroduced one paragraph below the table it fixed, in the region the new test cannot see. A backend author applying the stated rule to the rows above it gets a contradiction.

## The fix

The rule now scopes to the bands above `Background`, and says why `Background` is outside it: the mask is what a **host enforces before visible commit**, and requiring the one band a compositor may replace or drop would forbid the composition that band exists for. What a panel owes there is stated by `BackgroundCapability` and proven by admission. Both readings are correct — which is precisely why the paragraph had to say which question it answers.

## The same omission, still living in code

Both reviewers also found `PFD_CRITICAL` in the PFD's own tests: `[Attitude, Tapes, Annunciation]` — **no `Guidance`**, the identical stale fact #14 fixed in prose, three files away. So the band a host refuses a frame for lacking went unasserted by the test that checks the layer contract.

It is derived from `PFD_DESCRIPTOR.required_layers` now, which is the PR's own thesis applied to itself. Proven: wrapping the Guidance band in a condition now fails with `Valid missing Guidance`; with the hardcoded list the same mutation passed.

Two more stale statements in the same crate, both in the honest spirit of the issue: the PFD module doc named a `symbology` band the vocabulary does not define (it is `Guidance` — the band this issue turns on), and the crate doc offered "PFD and HSI panels" while shipping the monitor, which is the omission half of #14 restated one file up.

## Note on the header hole they also raised

One reviewer flagged that validating only `cells[0] == "Panel"` lets the column labels be swapped. That was real on `3c3d6e0` and is already closed on `main` — `9bdb147` pins all three header cells against a `HEADER` constant and parses positionally. The Codex reviewer re-verified against `9bdb147` and confirmed it fails closed.

## Gates

Full battery green; scene digest and screen-composition digest both match their pins; evidence gate VALID.

### DIFF
--- a/docs/instruments/scene-layer-protocol.md
+++ b/docs/instruments/scene-layer-protocol.md
@@ -124,12 +124,25 @@ ascending layer order. `layer_profile_doc_tests` in
 column disagrees with the descriptor it describes, so a mask change must
 arrive with its row. The section holds this one table and no other.
 
-A panel requires a band when it opens that band on *every* frame, not when the
-band always carries content. The PFD requires `Guidance` on those terms: it
-opens the band unconditionally so the layer contract holds under every
-degradation, and the flight-director command bars inside it disappear when the
-director is disengaged, invalid, or decluttered by the unusual-attitude tier.
-A backend must therefore not read an empty `Guidance` band as a missing one.
+Among the bands above `Background`, a panel requires one when it opens that
+band on *every* frame, not when the band always carries content. The PFD
+requires `Guidance` on those terms: it opens the band unconditionally so the
+layer contract holds under every degradation, and the flight-director command
+bars inside it disappear when the director is disengaged, invalid, or
+decluttered by the unusual-attitude tier. A backend must therefore not read an
+empty `Guidance` band as a missing one.
+
+`Background` is deliberately outside that rule and appears in no panel's
+required mask, which is why it sits in the optional column of every row above
+— including the HSI's and the monitor's, both of which do open it on every
+frame. The mask is what a host enforces before visible commit, and requiring
+the one band a compositor may replace or drop would forbid the composition
+that band exists for. What a panel owes in that band is stated by its
+`BackgroundCapability` instead, and proven by the admission harness: an
+`Opaque` or `Cedeable` panel must lay a full-frame opaque cover there in every
+case, which is a stronger obligation than the mask could express. So for those
+panels the band is mandatory in practice and optional in the mask, and both are
+correct.
 
 A well-framed prefix that ends at a layer boundary is a smaller structural
 scene, not proof of a complete panel frame. Missing a required layer is a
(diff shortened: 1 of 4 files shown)


luofang34:
## Independent review — it caught me making the same error while fixing it

The reviewer verified every other claim in the replacement paragraph (HSI and monitor open `Background` unconditionally; no shipped descriptor — nor the template — requires it; the host-enforcement framing matches the document; no contradiction with `panel-contract.md` or `backend-contract.md`), and found that **my replacement over-claimed in exactly the way the original did**.

I wrote that the band is "mandatory in practice" for `Opaque` **or `Cedeable`** panels. The shipped `Cedeable` panel is the PFD, and its band is genuinely absent from some frames under a synthetic-vision or no-background configuration — asserted by *this branch's own edited test* (`assert!(!bare_report.contains(LayerId::Background))`), by `an_svs_config_blob_decodes_and_still_cedes` proving a real wire blob reaches that path, and by the same document's SVS section a few sections down.

Also over-strong: "in every case". The harness draws the **empty config**, so what it proves is the band-owning default, not every configuration a shell may supply — and `background.rs`'s own module doc says exactly that.

**Fixed in 7a7bb99.** The admission claim is scoped to the cases the harness draws, "mandatory in practice" is restricted to `Opaque`, and `Cedeable` gets its own paragraph — because it is the case the optional column suits *exactly*: requiring the band would forbid the one configuration the capability exists to allow.

## The regression this PR closes was demonstrated, not inferred

The reviewer swapped the old hardcoded `[Attitude, Tapes, Annunciation]` back in and re-ran the Guidance mutation: **both tests passed**. With the derived list, dropping Guidance, Attitude, or Tapes each fails both tests.

They also confirmed excluding `Background` from `pfd_critical()` is necessary rather than cosmetic — `critical_overlay_is_byte_identical_without_background` renders with `BackgroundMode::None`, so including it would panic on a `None` range — and that an empty iterator cannot drift in one edit, since `required_layers == 0` is rejected at registry init and any mask change breaks `layer_profile_doc_tests` unless the table moves with it.

Both doc-comment fixes confirmed accurate against the emission order and `BUILTIN_PANELS`.

663 tests pass; both HARD evidence gates VALID.

PR #34, opened 2026-08-08
The descriptor answers which frames it accepts (#31)

### LINKED ISSUE #31: The descriptor publishes frame bounds but no predicate over them, so every shell writes the rule
`PanelDescriptor` publishes `frame_min`, `frame_max`, `frame_step`,
`aspect_min`, `aspect_max`, and `canonical_frames`. It publishes no predicate
over them. `grep -rE "fn (accepts|fits|permits|allows)"` across
`indicate-instrument-descriptor` and `indicate-instrument-registry` returns
nothing.

So every shell writes the rule. Migrating avionics-link to v0.2.0 meant writing
this, and it is not one line — it is six bounds plus a float tolerance on the
step:

```rust
fn accepted_frame(panel: &PanelDescriptor, width: f32, height: f32) -> Result<DesignFrame, _> {
    // finite and positive; within min and max; aspect within bounds;
    // and on the step from the minimum, within tolerance
}
```

The next shell writes a second one. They will differ — on the tolerance, on
whether the bound is inclusive, on whether the step is measured from `frame_min`
or from zero — and each will be locally green, because each shell only ever
tests its own. Two shells then disagree about which frames a panel accepts, and
the panel is the one thing that knows.

This is the same shape as the critical-layer mask before #6: the family
published the facts and left the rule to consumers, and the mask ended up
written three times. The lesson generalises — **a contract owner publishes the
predicate, not just the data it reads.**

## Ask

`PanelDescriptor::accepts(DesignFrame) -> Result<(), FrameRefusal>`, with the
step tolerance decided once, here, where the constants live. A typed refusal is
worth more than a bool: a shell that asked for a frame wants to know which
bound it violated, and the diagnostic then reads the same on every shell.

Worth doing before a second panel set, because a set that offers a real frame
range is the first time consumers will disagree about anything but the trivial
case.

luofang34:
Fixes #31.

## The rule now has one home

`PanelDescriptor::accepts(DesignFrame) -> Result<(), FrameRefusal>`, in `indicate-instrument-descriptor` where the constants live.

Typed rather than a bool, as the issue asks: a shell that asked for a frame wants to know *what to ask for instead*, and `Degenerate` / `OutOfRange` / `OffStep` / `Aspect` reads the same on every shell.

**The step tolerance is decided once and is zero** — `FRAME_STEP_TOLERANCE`, exported so a shell uses it rather than choosing its own. A step that cannot express a frame exactly is a declaration to fix, not a rounding to absorb; admitting near-misses would pin the digest and the raster baselines at frames the arithmetic cannot reproduce.

## Load-bearing on arrival, not a dead API

This was the issue's key requirement, and it is satisfied in two places, not one:

- **`Registry::new` holds every canonical frame to it.** So the sizes a panel's evidence is pinned at are exactly the sizes a shell may ask that panel for. The refusal is *mapped* back to a per-rule registry error rather than collapsed, because a declaration is fixed by knowing which bound it broke.
- **The composition layer's slot check calls it too.** `validate_composition`'s rule 3 previously had its own copy via a registry-internal `supports()`; that copy is gone.

So the three places that had to agree — the descriptor's declaration, the registry's canonical-frame validation, and composition's slot sizing — now cannot disagree, because there is one rule.

## Acceptance

- **A descriptor whose canonical frames fall outside its own bounds is refused by `Registry::new`, and the refusal names the bound.** All twelve `frame_range` tests still pass, now driven through the predicate: `CanonicalFrameOutOfRange`, `CanonicalFrameOffStep`, `CanonicalFrameAspect`.
- **The step tolerance appears exactly once in the repo.** `grep -rn "% step\|FRAME_STEP_TOLERANCE"` returns only its definition in `frame.rs`, its use two lines below, and the re-export.

Five new tests cover the predicate directly, including that each bound refuses *by name* (a 600×360 frame is in range and on the grid, so only `Aspect` can refuse it; a 500×375 frame is in range and 4:3, so only `OffStep` can) and that a non-finite or non-positive dimension is refused before any bound is consulted.

## Contract prose

`backend-contract.md` now tells shells not to re-derive the rule, names the refusal variants, states the tolerance, and says why: the panel is the one thing that knows which frames it accepts.

## Gates

fmt, clippy `-D warnings`, test `--all-targets`, doc, `thumbv7em-none-eabihf` (the predicate is in the no_std kernel), all script gates, both digests matching their pins.

No behavior change: every shipped panel declares a degenerate range, and the predicate accepts exactly the frame they already accepted.

### DIFF
--- a/crates/indicate-instrument-descriptor/src/frame.rs
+++ b/crates/indicate-instrument-descriptor/src/frame.rs
@@ -46,15 +46,22 @@ impl PanelDescriptor {
     ///
     /// This is the whole rule, and the only copy of it. A shell asks
     /// before drawing; the draw path re-checks nothing.
+    ///
+    /// Safe on a descriptor this crate never validated: a bound that is
+    /// not a number refuses every frame rather than accepting one.
     pub fn accepts(&self, frame: DesignFrame) -> Result<(), FrameRefusal> {
         if !(finite_positive(frame.width) && finite_positive(frame.height)) {
             return Err(FrameRefusal::Degenerate);
         }
-        if frame.width < self.frame_min.width
-            || frame.width > self.frame_max.width
-            || frame.height < self.frame_min.height
-            || frame.height > self.frame_max.height
-        {
+        // Phrased as "must be inside" rather than "reject if outside",
+        // so a bound that is itself not a number refuses every frame
+        // instead of accepting every frame. A shell may hold a
+        // descriptor this crate never validated.
+        let inside = frame.width >= self.frame_min.width
+            && frame.width <= self.frame_max.width
+            && frame.height >= self.frame_min.height
+            && frame.height <= self.frame_max.height;
+        if !inside {
             return Err(FrameRefusal::OutOfRange);
         }
         if !on_grid(frame.width, self.frame_min.width, self.frame_step.0)
@@ -63,7 +70,7 @@ impl PanelDescriptor {
             return Err(FrameRefusal::OffStep);
         }
         let aspect = frame.width / frame.height;
-        if aspect < self.aspect_min || aspect > self.aspect_max {
+        if !(aspect >= self.aspect_min && aspect <= self.aspect_max) {
             return Err(FrameRefusal::Aspect);
         }
         Ok(())
@@ -75,13 +82,18 @@ fn finite_positive(value: f32) -> bool {
 }
 
 /// Whether `value` is `min + k * step` for a whole `k`.
+///
+/// Both sides of a grid line are tested. At a tolerance of zero the two
+/// tests collapse into one, but a one-sided test would silently admit
+/// near-misses above a line and refuse the identical miss below it,
+/// which is not what a tolerance means.
 fn on_grid(value: f32, min: f32, step: f32) -> bool {
     let offset = value - min;
     if offset < 0.0 {
         return false;
     }
     let remainder = offset % step;
-    remainder <= FRAME_STEP_TOLERANCE
+    remainder <= FRAME_STEP_TOLERANCE || remainder >= step - FRAME_STEP_TOLERANCE
 }
 
 #[cfg(test)]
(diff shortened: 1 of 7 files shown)


luofang34:
## Independent review — sound, three low findings taken

The reviewer built a **differential harness** running the old rule against the new one over 3060 checks (10 descriptor variants × 306 frame probes including NaN, ±inf, ±0.0, negatives, `MIN_POSITIVE`, `MAX`, and off-grid-by-one-ULP): **zero divergences** on any descriptor a registry can hold. They also confirmed `x % 0.0` is IEEE fmod (NaN, not a panic) so a zero step yields `OffStep` exactly as before, and that a negative or infinite step matches the old answer too.

They verified acceptance by construction rather than by trusting the suite — 11 descriptors through `Registry::new` — and noted the existing suite had **no** below-minimum case and **no** non-finite case; both now confirmed working.

**Fixed in 3b03030:**

1. **The refactor failed *open* where the old rule failed *closed*.** I phrased the range and aspect tests as "reject if outside", and every comparison against a NaN bound is false — so a descriptor with a NaN `frame_max` or `aspect_max` accepted *every* frame, including the squat shape the aspect rule exists to refuse. Unreachable in-tree, but `accepts` is public and the whole point is that shells call it, possibly on a descriptor from a set built elsewhere. Both tests read "must be inside" now, with a test pinning it.
2. **The grid test was one-sided.** Inert at a tolerance of zero, but the moment the constant moved it would admit a near-miss just above a line and refuse the identical miss just below. Both sides are tested now, and the doc says why.
3. **Paragraph glue** in  — the new prose had swallowed a sentence belonging to the paragraph above it.

**Confirmed:** the tolerance appears exactly once (other repo hits are unrelated domains); no copy of the rule survives in registry, conformance, raster, tools, or sets; the rule moved *down* a tier, from verification into the `no_std` kernel where the constants live; nothing moved — both digests, corpus, baselines, and admission 242/166 byte-identical to main; 668 tests (main 663, +5 new); all 10 script gates and both HARD evidence gates green.

PR #37, opened 2026-08-08
A release manifest a consumer's CI can diff (#33)

### LINKED ISSUE #33: A release manifest of pinned values, and a consumer-side check that notices drift
A consumer taking a new revision has to re-pin things. Which things is
currently answerable only by reading source across several repositories.

The pinned values I can find, having just migrated one consumer:

| Value | Lives in | Who pins it |
|---|---|---|
| State ABI version | `abi::v6::VERSION` | every feeder and shell |
| Scene format version | `SCENE_FORMAT_VERSION` | every backend |
| Corpus version + sha | `corpus/scene-conformance-corpus.json` | every backend |
| Composition digest | `BUILTIN_SCENE_DIGEST` | shells that compose the builtin set |
| Screen-composition digest | `composition_digest` | shells that compose screens |
| Raster frame baselines | `raster_baselines` per descriptor | the reference rasterizer |
| Glyph pack content hash | `PANEL_GLYPHS.recorded_hash()` | backends that report which pack is loaded |
| Criticality bands | `BUILTIN_CRITICALITY_BANDS` | shells that place panels by criticality |

`CHANGELOG.md` names five of them, which is already the best index that exists
and is why the migration was tractable at all. The gap is that five is not all
of them, the table is prose, and a consumer cannot check its own pins against
it mechanically.

This is good discipline outrunning its own indexing. Every one of those values
exists because something worthwhile is being held down; the cost is that the
set has grown past what a person can hold, and there are now four pin edges
between repositories (Indicate→Pilotage, Indicate→avionics-link,
Indicate→InstrumentSceneKit for the corpus, InstrumentSceneKit→avionics-link).
Two of those four had drifted fifteen commits without anything noticing.

## Ask

A machine-readable release manifest — one file per tag, listing every pinned
value and its current content — plus a consumer-side check that compares its
own pins against it and fails on a mismatch.

The check is the half that matters. A manifest a consumer has to read by hand
is the changelog again; a manifest a consumer's CI can diff against is the
thing that notices drift before a migration does.

Worth noting that consumers have already diverged on strategy: Pilotage reads
the corpus straight out of the pinned crate checkout, while InstrumentSceneKit
vendors a copy with its own provenance record. Both are reasonable; a manifest
is what lets either of them prove it is current.

luofang34:
Fixes #33.

## What lands

`release-manifest.json` — all **eight** pinned values, machine-readable, versioned with the code so `git show v0.2.0:release-manifest.json` is that tag's manifest:

state ABI · scene format · corpus version + sha256 · composition digest · screen-composition digest (with the screen it was taken over) · every raster baseline keyed by panel and frame · glyph pack hash · criticality bands.

The changelog names five of them in prose and stays the human summary; this is the half a consumer's CI can diff.

## Generated from the constants, not grepped

`cargo xtask gen-release-manifest` reads the actual values, so the file cannot drift from the code by construction. `check-release-markers.sh` greps because it validates a *prose* table; this does not.

Stronger than that: **where the tree can recompute a pin, the generator recomputes it and refuses to emit on disagreement** — scene digest against `BUILTIN_SCENE_DIGEST`, screen-composition digest against the bench pin, glyph `content_hash()` against `recorded_hash()`. A stale pin stops the generator rather than being laundered into a machine-readable claim.

Two runs are byte-identical (verified by `cmp`, and against the committed file). Floats print at fixed precision, so the bytes never depend on the formatter's shortest-round-trip choice.

## The acceptance criterion, verified directly

> hand-editing the tree without updating the manifest must make CI red

Changing the PFD's raster baseline hash in `descriptors.rs` and running the check:

```
DRIFT: release-manifest.json disagrees with the tree it describes
DRIFT: the keys that moved: frame height panel sha256 width
    --- release-manifest.json (committed)
    +++ release-manifest.json (what this tree generates)
regenerate and commit it: cargo run -q -p xtask -- gen-release-manifest
check-release-manifest: FAILED
```

`exit=1` on the drifted tree, `exit=0` once restored. Wired into `.github/workflows/ci.yml`.

Fails closed on a missing manifest, an empty one, or a generator that fails — and ships a **selftest over eight failure modes** (`scripts/test-release-manifest.sh`), matching the discipline `check-standards-registry.sh` and `check-structure.sh` already keep. Its drift fixtures match by *shape* rather than by current value, so a pin that legitimately moves does not send an author to re-edit a fixture.

## One structural change

The bench's fixture screen moves into a small library that both the bench and the generator read, so the manifest cannot describe a composition the bench does not actually check.

## What it deliberately does not do

It states what *this* revision pins. It cannot verify a **consumer's** pins — that is each consumer's own check to write against this file, and the README says so plainly rather than implying the file does more than it does. Both consumer strategies the issue notes (Pilotage reading the corpus from the pinned checkout, InstrumentSceneKit vendoring a copy) can prove currency against it.

## Nothing pinned moved

Scene digest `3efb08c5…`, screen-composition digest `071bd35c…`, the three REN-03 baselines, corpus 4 / `1fb8e6de…`, admission 242/166 — all unchanged. The only `Cargo.lock` change is xtask's new dependency lines.

Full battery green, including both HARD evidence gates and all ten script gates.


luofang34:
## Independent review — 14 mutation experiments; one MEDIUM fixed

The reviewer checked all eight values against their sources by hand, ran fourteen mutations through the guard, and probed the vacuous-pass and determinism surfaces. All eight values are correct — including the one I flagged for scrutiny: `90.857147` **is** the same f32 as the source literal `90.85715` (both are bits `0x40B5B6DB`), not a truncation.

### MEDIUM — the ABI scan was fail-open on exactly the mode it claims to prevent

`strip_suffix(';')` required the declaration to *end* at the semicolon, so:

| `abi.rs` line | before |
|---|---|
| `pub mod v7;` | refuses — correct |
| `pub mod v7; // staged` | **emits `v6`, check passes** |

`cargo fmt` accepts the decorated form, so nothing normalised it away — and the module doc claimed the scan follows `check-release-markers.sh`'s rule, which is prefix-anchored (`grep -oE '^pub mod v[0-9]+;'`) and *does* see it. The Rust scan was strictly weaker than the shell rule it cited.

This mattered beyond the scan: the changelog hunk asserts the two guards cannot contradict each other, and here they could — staging a module and bumping the changelog marker left **both** green while the manifest certified the wrong ABI to every consumer.

Fixed: the scan reads to the semicolon instead of requiring the line to end there, tested four ways, and verified end to end — a decorated `pub mod v7;` now stops the generator with `the generator reads the state ABI from v6, but the crate declares v7`.

### Three claims corrected rather than left standing

- **The glyph "cross-check" is a tautology.** `recorded_hash()` is *defined* as the content hash of the same groups, so the two cannot disagree. The comment now calls it a shape assertion and says glyph drift is caught by regeneration instead — which the reviewer confirmed fires (`DRIFT: the keys that moved: glyphPackHash`).
- **The float docstring overclaimed.** Six decimals do not separate every distinct `f32`: below magnitude 16 an ulp is finer than 1e-6, and the manifest holds `"x": 6`. Stated honestly now.
- **Rot-prone counts** ("five … eight") in three files — my own conventions ban exactly that, and nothing pinned them.

`json.rs` now carries the tests its determinism claim rests on, since the whole diff depends on that formatter.

### Confirmed by the reviewer

The guard bites on the state ABI, scene format, glyph hash, criticality bands, both digests, the raster baselines, and the bench fixture — twelve mutations, each red. No vacuous pass survived (`{}`, single space, zero bytes, missing path, a directory, `/dev/null`, a generator that emits nothing). CI sets **no** `INDICATE_*` variables, so neither env knob can weaken it and the selftest genuinely runs. Output is byte-identical across `LC_ALL=tr_TR`, `LC_NUMERIC=de_DE`, `TZ=Asia/Kolkata`, and a different cwd; no `HashMap`, no `cargo metadata`, no `Path::display()` in the emitted document.

Recorded honestly as a limit rather than fixed: three pins are *quotations* whose correctness rests on separate tests in the same CI job — a tampered byte inside a corpus **entry** passes the manifest check and is caught by `golden_matches_reference` instead. The manifest quotes the corpus header; it does not re-hash the corpus.

Issue #32, opened 2026-08-08
No shipped panel exercises the frame range, and nothing proves a panel varies with its DesignFrame


luofang34:
v0.2.0 replaced the single `design_frame` constant with a range a panel will
emit into, and every consumer paid for it: `DrawFn` gained a parameter,
descriptors gained six fields, and each shell had to decide which frame to ask
for.

All three shipped panels declare `frame_min == frame_max == 480×360`
(`sets/indicate-instrument-panels/src/descriptors.rs`, PFD, HSI, and monitor
alike), and `CANONICAL_FRAMES` is `&[BUILTIN_FRAME]`. The range contains one
frame. Nothing exercises the machinery it was built for.

That is a defensible way to sequence a breaking change — land the API before
the panels that need it, so consumers migrate once. Two things follow from it
that are worth naming rather than discovering:

**The choosing policy is unwritten and untestable.** A host with a 1024×768
viewport has to pick a frame. Today the only answer is "the canonical one", and
that is what the migrated shell does — not because it is the right policy but
because it is the only legal value. When a panel offers a real range, three
shells will each invent a policy, and there is currently nothing that says
whether the right answer is largest-that-fits, nearest-aspect, or nearest
canonical. Same divergence risk as the missing predicate in the sibling issue,
one level up.

**Nothing proves a panel can emit at two frames.** `raster_baselines` is keyed
by frame and every panel has one entry. A panel that quietly ignores its
`DesignFrame` argument and emits fixed geometry would pass every test in the
repository, because no test asks it for a second frame. Admission does not
check it either.

## Ask

Either the first panel that actually varies with its frame, or a test that
proves the shipped ones do not have to. Concretely, the cheapest useful step:

- give one shipped panel a real range and a second canonical frame with its own
  baseline, so the whole path — descriptor, `DrawFn`, baselines, digest,
  admission — is exercised end to end at least once;
- and state the choosing policy in the panel contract, so shells do not each
  invent one.

Until then the panel contract should say plainly that a panel may treat its
`DesignFrame` as fixed, so an author does not assume the framework is doing
something for them that nothing checks.

luofang34:
Half of this landed in #38: admission now refuses a panel that declares a non-degenerate range and emits identically at both ends of it, and the panel contract states plainly that a panel **may** treat its `DesignFrame` as a constant — so an author does not assume the framework lays out for them.

[...]

PR #38, opened 2026-08-08
A panel that declares a range must use it (#32)

luofang34:
Addresses the half of #32 that is cheap today. The other half — giving a shipped panel a real range — needs a layout, new canonical frames, new baselines, and a digest move, and is deliberately **out of scope** here; it belongs with the set that needs it.

## The hole

> A panel that quietly ignores its `DesignFrame` argument and emits fixed geometry would pass every test in the repository, because no test asks it for a second frame. Admission does not check it either.

Exactly right. The matrix asks each canonical frame the same questions *separately* and never compares the answers, and every shipped panel declares one frame — so nothing ever asked one for a second.

## What lands

**Admission renders a non-degenerate panel at both ends of its range and requires the bytes to differ.** Differing bytes are a weak proof of a good layout and a sufficient proof of the thing that was otherwise unprovable: that the argument reached the geometry at all.

New typed refusal `FrameIgnored { panel, min, max }`.

**The panel contract now says plainly that a panel may treat its `DesignFrame` as a constant**, which is the other thing #32 asks for — so an author does not assume the framework is doing something for them that nothing checks. The honest way to say it is a degenerate range, and `accepts` then refuses any other size before the panel is called.

What is refused is declaring a range and not honouring it: a shell asking for the larger frame is owed more instrument, not the same picture for the backend to stretch.

## Shown, not asserted

Two fixture panels, identical but for one line — one sizes its rect from the argument, the other hard-codes 480×360:

| Fixture | Result |
|---|---|
| ignores the frame | `FrameIgnored { panel: "fixed", … }` |
| uses the frame | admitted |
| degenerate range, fixed geometry | admitted — a degenerate range *is* the declaration that it does not vary |
| the three shipped panels | admitted, unchanged |

That last one is pinned deliberately: the check is inert for the shipped set today, so a panel that gains a range **without** gaining a layout cannot land quietly.

## Nothing moved

Scene digest `3efb08c5…`, screen-composition digest `071bd35c…`, admission still 242 cases / 166 warnings, raster baselines and corpus untouched. The new check does no work for a degenerate range, which is every shipped panel.

Full battery green including the new release-manifest guard and both HARD evidence gates.

## Still open in #32, and why I left it

The **choosing policy** — largest-that-fits, nearest-aspect, nearest-canonical — is unwritten. It cannot be settled honestly while every panel offers exactly one legal frame: any policy would be untestable and would read as arbitrary. It belongs with the first panel that offers a real range, alongside its content policy.

### DIFF
--- a/crates/indicate-instrument-conformance/src/admission.rs
+++ b/crates/indicate-instrument-conformance/src/admission.rs
@@ -38,6 +38,7 @@ mod alerts;
 mod background;
 mod criticality;
 mod error;
+mod frame_varies;
 mod geometry;
 mod ink;
 mod provenance;
@@ -45,6 +46,7 @@ mod regions;
 
 use alerts::saturated_stack;
 use background::check_background;
+use frame_varies::check_frame_varies;
 use geometry::{Ctm, Rect, text_rect};
 use provenance::check_provenance;
 use regions::check_non_vacuity;
@@ -144,10 +146,12 @@ fn admit_panel(
     for frame in panel.canonical_frames {
         admit_panel_at_frame(panel, *frame, report)?;
     }
-    // A per-panel fact rather than a per-case one: the witness a region
-    // needs may only appear in one case of the matrix, so the verdict
-    // waits until every case has been drawn.
-    check_non_vacuity(panel)
+    // Per-panel facts rather than per-case ones. The witness a region
+    // needs may appear in only one case of the matrix, and whether a
+    // panel varies with its frame is a comparison across frames that no
+    // single case can make.
+    check_non_vacuity(panel)?;
+    check_frame_varies(panel)
 }
 
 fn admit_panel_at_frame(
--- a/crates/indicate-instrument-conformance/src/admission/error.rs
+++ b/crates/indicate-instrument-conformance/src/admission/error.rs
@@ -170,4 +170,17 @@ pub enum AdmissionError {
         /// search ran at.
         frame: DesignFrame,
     },
+    /// A panel declares a range of frames and emits the same bytes at
+    /// both ends of it.
+    #[error(
+        "panel {panel} emits identically at {min:?} and {max:?}, so it does not use the frame it is given"
+    )]
+    FrameIgnored {
+        /// The declaring panel.
+        panel: &'static str,
+        /// The smallest frame it accepts.
+        min: DesignFrame,
+        /// The largest frame it accepts.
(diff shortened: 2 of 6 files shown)


luofang34:
## Independent review — two confirmed false refusals, fixed

The reviewer built panels that **do** read their frame and were refused anyway. Both reproduced:

- a panel whose width tracks `frame.width` only when alerts are fed → `FrameIgnored`
- a panel whose width tracks the frame only at an airspeed produced by its own `ExtremeState` → `FrameIgnored`

The search covered only the canonical states, drawn quiet, while the matrix drives three more axes — extreme states, withholding, and the alert axis. Telling a panel it ignores an argument it uses is the worst failure mode for this check, since the author's only recourse is to distrust the harness.

**Fixed:** it searches the whole case matrix now, a strict superset of what it looked at before, so it can only refuse *fewer* panels. One witness still settles it. The alert-only panel is now a permanent test.

**Also fixed — a vacuous pass the reviewer found:** the check compared two consecutive draws and treated any byte difference as proof, so a panel carrying interior phase state (blink, scroll counter) passed on variation that had nothing to do with the frame. A difference now counts only when the smaller frame **reproduces**, making a deterministic `DrawFn` part of what is checked rather than assumed.

**Two documentation corrections:**

- The contract said asking for another size "is refused by `accepts` before you are called". It is not — `accepts` is a predicate a shell is expected to consult, and no draw path re-checks the argument (`raster/src/composition.rs`, `instrument-bench`, and `digest.rs` all call `(panel.draw)` with a caller-chosen frame). The backend contract already had the honest phrasing — what a *conforming* shell may ask — and this now matches it.
- The conformance crate's own summary still described declared regions as **containment**, which non-vacuity replaced, and counted six families where there are now seven. Both corrected.

**Scratch moved to the heap**, matching every other buffer in the crate; 128 KiB of stack was safe but looked nothing like its neighbours.

### Confirmed sound

`range_is_degenerate` is correct given a validated registry, and the reviewer verified the registry is genuinely closed — one private field, no `Default`, both constructors funnelling through `validated` — so NaN and `-0.0` bounds cannot reach `admit()`. The check is reached from `admit()` for every panel and its `Err` propagates. Nothing moved: the diff against `sets/`, the corpus, `release-manifest.json`, and the rasterizer is **zero lines**, and admission is still 242/166 because the check early-returns on every shipped panel.

PR #40, opened 2026-08-08
Take sha1 and sha2 0.11 (#2, #3)

luofang34:
Supersedes the two dependabot PRs, #2 (sha1 0.10.7 → 0.11.0) and #3 (sha2 0.10.9 → 0.11.0), rebased onto current `main` and taken together because they share a transitive dependency change.

## Why these two need more than a green build

Both compute values something downstream **pins**:

- **sha2** — the REN-03 raster frame hashes, the three composed-frame hashes, the corpus hash, and every recorded evidence artifact digest.
- **sha1** — the git blob identity of the test sources a recorded verification result binds itself to.

A major bump of a hashing crate is therefore a question about pinned values, not about compilation. If any of them moved, every consumer would go red at its next pin advance and the evidence graph would stop resolving.

## None moved — checked, not assumed

| Pin | Result |
|---|---|
| Scene digest | `3efb08c5…` matches |
| Screen-composition digest | `071bd35c…` matches |
| REN-03 panel hashes (×3) | reproduce and match |
| Composed-frame hashes (×3) | reproduce and match |
| Corpus | unchanged |
| Release manifest | regenerates identically |
| Admission | 242 cases / 166 warnings |

The **evidence gate exercises both crates at once** — sha1 on every bound source digest, sha2 on every artifact — and resolves to the same graph digest `f993634c…` with `verdict: VALID`. All 26 negative fixtures still fail closed.

That is the expected outcome: SHA-1 and SHA-256 are the same algorithms they were. What 0.11 changes is the crate API, and this tree's use of `Digest` is untouched by it — no source file needed editing.

## Dependency graph

Loses `generic-array` and `version_check`, gains `hybrid-array` — all transitive, none a wire-protocol, host, or shell crate, so the downstream-agnostic boundary is unmoved and its CI gate stays green.

## Full battery

fmt, clippy `-D warnings`, test `--all-targets`, doc, release build, `thumbv7em-none-eabihf` closure, and all six script gates including the new `check-release-manifest.sh`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -31,11 +31,11 @@ checksum = "b588b76d00fde79687d7646a9b5bdf3cc0f655e0bbd080335a95d7e96f3587da"
 
 [[package]]
 name = "block-buffer"
-version = "0.10.4"
+version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3078c7629b62d3f0439517fa394996acacc5cbc91c5a20d8c658e77abd503a71"
+checksum = "d2f6c7dbe95a6ed67ad9f18e57daf93a2f034c524b99fd2b76d18fdfeb6660aa"
 dependencies = [
- "generic-array",
+ "hybrid-array",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


luofang34:
## Independent review — no findings

The reviewer re-derived every pinned value on **both** branches in separate worktrees and diffed them, rather than reading the committed pins: scene digest, screen-composition digest, admission 242/166, the regenerated manifest (byte-identical, 1914 bytes), and the evidence graph digest `f993634c…` — all identical, with the same 690 tests and the same test-name set.

They caught something I had not stated: **manifest equality alone would be circular**, because the generator *reads* `corpusSha256` and `raster_baselines` rather than recomputing them. The non-circular sha2 exercises are the raster and conformance tests, which recompute with `sha2::Sha256` and compare against the pinned constants, plus the evidence graph digest. Those are what actually pin 0.11's output, and they agree.

**sha1 verified against real git**: `git_blob_sha1` recomputed for three bound sources and matched `git hash-object` exactly on both branches, with the `stale-source-artifact` fixture proving the path is non-vacuous.

**API safety is structural, not assumed**: `Sha256::digest(bytes).into()` → `[u8; 32]` is a compile-time length check, so a changed output length or silent truncation could not type-check. And `indicate-sha256`'s own test cross-checks sha2 0.11 against the vendored FIPS 180-4 implementation over lengths 0..=300, one-shot and streamed — two independent implementations agreeing.

**Dependency graph**: `hybrid-array` and `const-oid` in (the latter a non-optional dep of `digest 0.11`, not a choice this PR made), `generic-array` and `version_check` out. Net **one fewer build script**, no proc-macro change, and licences strictly more permissive (`generic-array` was MIT-only; both new crates are `MIT OR Apache-2.0`). MSRV floor 1.85 matches the edition-2024 floor already required. The downstream-agnostic gate run verbatim: `seen=16 ≥ floor=15`, still has teeth.

`--locked` honoured throughout — `git diff HEAD -- Cargo.lock` empty after every invocation.

All 18 script gates, both HARD evidence gates, 26/26 negative fixtures, and exception expiry in both directions.

PR #41, opened 2026-08-08
A public API addition is worth a tag (#39)

### LINKED ISSUE #39: The release trigger does not cover public API additions, so the two most consumer-facing ones are untagged
`CONTRIBUTING.md` says it plainly:

> Cut one whenever any of the five contract values moves: state ABI, scene
> format, corpus, composition digest, or the panel set.

A public API addition moves none of those. Two have landed since `v0.2.0` and
neither is reachable by tag:

- `PanelDescriptor::accepts` and `FrameRefusal` (#31 / #34) — the whole point
  of which was to stop consumers writing their own frame rule;
- `release-manifest.json` and `scripts/check-release-manifest.sh` (#33 / #37) —
  the thing a consumer's CI is meant to diff against.

So a consumer that wants either must pin a bare revision and explain in a
comment which commit it is and why. That is exactly the archaeology the tag
policy was introduced to remove (#8), reappearing through a gap in the trigger
rather than through the absence of tags.

It bites hardest on the two additions above, because both exist *for*
consumers. A predicate nobody can pin by name does not stop the second copy
from being written; a manifest nobody can pin by name does not get diffed.

## Two ways out, and they are not equivalent

**Widen the trigger.** Add "a public API addition consumers are expected to
call" to the five. Cheap, and it makes the tag mean "a revision worth pinning"
rather than "a revision whose contract values moved". The cost is more tags,
some of which carry no contract change — which the manifest and changelog
already make harmless, since both state exactly what did and did not move.

**Or say the quiet part.** Decide that tags track contract values only, and
that consumers pin bare revisions for anything else — then say so in
`CONTRIBUTING.md` and in the README's boundary section, so a consumer reading
"pin this repository by git rev" knows a tag is not always available and stops
looking for one.

Either is defensible. What is not is the current state, where the policy reads
as though a tag exists for anything worth pinning and one does not.

## Ask

Pick one and write it down. If it is the first, `v0.2.1` covering #34 and #37
is the immediate consequence, and avionics-link can then adopt `accepts` by
name instead of by revision.

luofang34:
Fixes #39. Doc-only plus a release entry; no code changes.

## The choice

The issue offers two ways out and says they are not equivalent. I took **widen the trigger**, because the alternative — documenting that tags track contract values only — would leave the two additions that exist *for consumers* permanently unpinnable by name, and both fail at their own purpose when unreachable:

- a predicate published so no shell writes its own copy does not stop the second copy from being written;
- a manifest published for a consumer's CI to diff does not get diffed.

Reaching either meant pinning a bare revision and explaining in a comment which commit it was — the archaeology tags were introduced to remove (#8), arriving through the trigger rather than through the absence of tags.

## What changes

`CONTRIBUTING.md`'s trigger gains "or when a public API addition lands that consumers are expected to call", with the reasoning recorded so the next reader knows it is deliberate.

The issue names the cost honestly and so does the doc: some tags now carry no contract change. That is harmless here precisely because the changelog entry and the release manifest both state exactly what did and did not move — a consumer sees at a glance whether it has anything to re-verify.

## 0.3.0

Cut for the two additions the issue names, plus the frame-variation check and the sha1/sha2 bump. Its table is **identical to 0.2.0's** — same ABI, scene format, corpus, composition digest, and panel set — and the entry says so plainly, so advancing a pin across it is a manifest bump and nothing else.

The re-pinning note also tells anyone who wrote their own frame-bounds check while migrating to 0.2.0 that `accepts` replaces it, and warns that a local epsilon on the step would accept frames this repository refuses — the tolerance here is zero, and it is what the descriptors' own canonical frames are validated against.

`check-release-markers.sh` validates the new entry against the tree: `OK (0.3.0; ABI 6, scene 1, corpus 4, panels: pfd, hsi, monitor)`. The manifest is unchanged and still regenerates identically, which is the point.

I will tag `v0.3.0` on the merge commit, per the release steps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,56 @@ from a rewritten one. Entries are newest first, and the tag's message
 repeats the same five values so `git show <tag>` answers the question
 without a checkout. `CONTRIBUTING.md` has the release steps.
 
+## [0.3.0] — 2026-08-08
+
+Two additions made for consumers, neither of which moves a contract
+value — which is why the release trigger now covers a public API
+addition as well as the five. A predicate nobody can pin by name does
+not stop the second copy being written, and a manifest nobody can pin by
+name does not get diffed.
+
+| Value | This release |
+|---|---|
+| State ABI | 6 |
+| Scene format | 1 |
+| Corpus | 4 |
+| Composition digest | `3efb08c55eadadc2b006ee6b006b29e4b3a3f8d4ec3ce1324f401dbc16dc85ca` |
+| Panel set | `pfd`, `hsi`, `monitor` |
+
+Panel set changed since the previous release: no.
+
+- **`PanelDescriptor::accepts(DesignFrame) -> Result<(), FrameRefusal>`.**
+  The rule over the declared frame bounds, in the crate where the
+  constants live, so a shell does not write its own. The refusal names
+  the bound it broke, and `FRAME_STEP_TOLERANCE` settles the one
+  parameter two shells would otherwise each choose. `Registry::new` and
+  composition's slot check both call it, so it cannot rot as an unused
+  API.
+- **`release-manifest.json` and `scripts/check-release-manifest.sh`.**
+  Every pinned value this revision holds down, generated from the
+  constants rather than grepped for, with CI failing when the file
+  disagrees with the tree. It states what *this* revision pins;
+  comparing that against its own is each consumer's check to write.
+- **Admission refuses a panel that declares a frame range and emits
+  identically across it.** Inert for every shipped panel, all of which
+  declare a degenerate range, and the panel contract now says outright
+  that treating the frame as a constant is allowed.
+- **sha1 and sha2 at 0.11.** No pinned value moved; the algorithms did
+  not change, only the crate API around them, and this tree's use of
+  `Digest` was untouched by it.
+
+### Notes for anyone re-pinning
+
+- Nothing to re-verify. Every value in the table above is what `v0.2.0`
+  carried, the raster baselines and criticality bands are unchanged, and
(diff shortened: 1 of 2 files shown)


luofang34:
## Independent review — one contradiction fixed

The reviewer verified every factual claim in the 0.3.0 entry against the tree and against `v0.2.0`: all five values correct, the table byte-identical to 0.2.0's including the full digest, all four bulleted additions genuinely landed in `v0.2.0..HEAD`, and `git diff v0.2.0 HEAD` touching neither `descriptors.rs`, nor the raster crate, nor the corpus — so "raster baselines and criticality bands unchanged" and "corpus untouched" both hold.

**The finding, and it was in the same file:** `CHANGELOG.md`'s preamble still stated the narrow trigger — "A release is cut whenever any of the five moves" — which contradicted the widened one and, read literally, declared that the 0.3.0 entry beneath it should not exist. The sentence introducing the table was stale for the same reason. Both fixed: the preamble now carries both triggers and says that an entry whose five values match the one below it is the second kind.

They also sharpened the re-pinning advice's justification: at `FRAME_STEP_TOLERANCE = 0.0`, `on_grid` reduces to `remainder == 0.0`, so any ε > 0 is strictly more permissive — an epsilon consumer check accepts frames `accepts` refuses, never the reverse. With every shipped panel degenerate, the live case today is an epsilon on the *bounds* comparison accepting 480.0001 where `accepts` returns `OutOfRange`. The sentence is true either way.

Confirmed no contradiction with the README's pinning section or the marker script's comments, that the script reads 0.3.0 with 0.2.0 neither orphaned nor double-counted, and that the manifest is untouched and still regenerates identically. All gates green including both HARD evidence gates.

PR #42, opened 2026-08-08
Velocity validity splits where its consumers do (#30)

### LINKED ISSUE #30: The shipped PFD paints a fabricated 0 fpm VSI: velocity validity is one bit for a three-component vector
A source that supplies horizontal velocity and no vertical speed cannot say so,
so the shipped PFD paints a live VSI needle parked at 0 fpm while the aircraft
climbs. No flag, no dashes — a fabricated primary-flight reading, which is the
one failure this whole validity model exists to prevent.

## Why it is not a feeder bug

Kinematics carries a single validity bit for the whole NED velocity vector
(`abi/v6/declared.rs:16`, "valid flags … 3 velocity"), one integrity check over
the whole array (`validate.rs:85`, `values.iter().all(is_finite)`), and one
resolved status shared by every derived signal:

```
resolve.rs:281   vsi_fpm: finite(Sig::with_status(vsi_fpm, vel_status)),
resolve.rs:~279  gs_kt:   finite(Sig::with_status(gs_kt,   vel_status)),
```

A feeder therefore has exactly two options, and both are wrong:

- **Zero the down component.** `vsi_fpm` resolves `Valid` at `0.0` and the PFD
  draws the needle. Measured downstream against a real Connext panel.
- **Make it non-finite.** `integrity.velocity` faults, `vel_status` degrades,
  and ground speed and track go down with it — both of which the source really
  does supply, and the HSI track diamond with them.

Verified against this revision: with a NaN down component, `gs_kt` and
`track_rad` stop resolving `Valid`. This is not a matter of feeding it more
carefully.

## Why v6 did not fix it

v6 made *groups* extensible — presence is meaning, sources with different group
sets drive the same panels. Within a group the granularity is unchanged: one
bit per group, and kinematics is one group holding position, horizontal
velocity, and vertical speed. The extensibility stops exactly one level above
where this source needs it.

This was reported against the closed v5 ABI as
[`luofang34/Pilotage#256`](https://github.com/luofang34/Pilotage/issues/256)
and travels here with the family.

## Ask

A status per signal, or per component, rather than per group — enough for a
source to say "I have horizontal velocity and no vertical speed" and for the
PFD to flag the VSI while the tape and the rose keep reading.

Whatever the shape, the acceptance is behavioural: a source that supplies
ground speed and track and no vertical speed must produce a PFD whose VSI does
not show a number and whose ground speed does.

Downstream keeps the zero for now — losing two real readings to flag a third is
worse — recorded as a known misstatement with a test that fails the day this is
expressible.

luofang34:
Fixes #30. Unblocked now that Pilotage has migrated — this is the ABI-breaking change the family was holding until both consumers sat on the v0.2.x baseline, so the ABI moves once and each side follows once.

## The defect was the model, not the feeder

Kinematics carried one validity bit for the whole NED velocity vector, one integrity check over the whole array, and one resolved status shared by every derived signal. A source with horizontal velocity and no vertical speed had two options and both were wrong:

- zero the down component → `vsi_fpm` resolves `Valid` at `0.0` and the PFD paints a live needle **while the aircraft climbs**;
- make it non-finite → ground speed *and* track go down with it, both of which the source really does supply.

A fabricated primary-flight reading is the one failure this validity model exists to prevent.

## The split, and where it is drawn

Horizontal and vertical, across all three layers — the declared bit (wire bit 3 keeps horizontal, bit 8 takes vertical from the spare range), the integrity check (north/east finiteness versus down), and the resolved status. Ground speed and track take horizontal; vertical speed takes its own; **a non-finite down component now faults vertical alone**.

Not per component. No panel reads north or east by itself, so a three-way split would mint statuses nothing could act on differently. The seam is where consumers actually differ, and it is exactly the seam the reported source needs.

## The acceptance, tested as stated

> a source that supplies ground speed and track and no vertical speed must produce a PFD whose VSI does not show a number and whose ground speed does

Asserted at both levels. At resolve: `gs_kt.status.shows_value()` and `!vsi_fpm.status.shows_value()`. At the emitted scene: **zero fill rects in the VSI needle strip, zero text runs at the numeral anchor, and the ground-speed box still reading `"GS 103kt"`** — on a *climbing* aircraft, which is the case that used to paint the fabricated zero.

A control declaring both axes restores the needle and its numeral, so the absence is the split working rather than the VSI being broken. Plus the converse (vertical present, horizontal absent) and a both-declared case that behaves exactly as before.

## What moved, and what did not

| | |
|---|---|
| State ABI | v6 → **v7**, replacing it |
| Composition digest | `3efb08c5…` → `9a80dbcd…` (it binds the ABI version byte) |
| Screen-composition digest | `071bd35c…` → `34dc4332…` (it binds the composition digest beneath it) |
| **Raster frame hashes** | **unchanged** — all three byte-identical |
| **Corpus** | **unchanged** |
| Admission | 242 cases / 166 warnings, unchanged |

No paint moved. That is what separates a version move from a regression, and I verified it by diffing the pinned literals against `main` rather than trusting the tests.

`release-manifest.json` regenerates with exactly three lines changed — `stateAbi`, `compositionDigest`, `screenComposition.digest` — with the raster baselines, corpus, glyph hash, and criticality bands unchanged in context. CHANGELOG gains 0.4.0.

## The re-pinning hazard, stated in the changelog

A feeder that sets only the old bit 3 now declares **horizontal velocity alone** and loses its VSI. That is fail-closed and it is the right default, but it is a behaviour change for any source that meant "both", and a migrating consumer needs to know to set bit 8.

## Evidence

Two of the three bound sources name the split field and could not compile unedited, so they were re-run and re-captured; the presentation suite is untouched and its digest is unchanged. Counts move (98 / 174 / 81) because this change *adds* tests. Both HARD gates VALID; all 26 negative fixtures still fail correctly.

**Follow-up required:** the squash orphans this baseline, so a re-anchor lands immediately after, as with every prior evidence-touching merge.


luofang34:
## Independent review — one real finding, fixed; everything else held

The reviewer did not take the acceptance on trust. They built the state themselves using the issue's *actual* feeder workaround (a **zeroed** down component, not the committed test's `-4.0`), and:

- confirmed `gs_kt.status == Valid` **by equality**, not merely `shows_value()` — so the VSI's disappearance is not a side effect of the state being degraded for some other cause;
- **reproduced the regression on `main`**: same state, and the scene contains `Rect{Fill, x:466, y:180, w:8, h:1}` — the fabricated 0 fpm needle, guaranteed non-empty by `len.max(1.0)`. Both horns of the issue reproduce there and neither does here.

They also proved the two moved digests are the version byte and nothing else, by patching `v7::VERSION` back to `6` in an otherwise fully-v7 tree: the digests came out as `main`'s exact values. And the golden fixtures differ from main's in **exactly two bytes each**.

### MEDIUM — the VSI painted nothing where every other readout dashes

This is the finding worth the review. `tapes::vsi` early-returned on an unusable value, leaving four grey ticks and no number — while airspeed, altitude, baro, and groundspeed all dash. `panel-contract.md` says it outright: *"Drawing nothing where a dash belongs is admissible, and it is a defect."*

It is pre-existing, but **this PR is what makes the case operationally normal**: before it a blank VSI meant the whole kinematics group was down and the panel was visibly degraded; now it happens on an otherwise fully green PFD, in exactly the case this issue exists to enable. A blank scale scans as zero vertical speed — the same fabrication the split was made to prevent, arriving through the drawing instead of the status.

Fixed: the VSI dashes, unclaimed like its neighbours. The composition digest moves again for it (`9a80dbcd…` → `f82d9056…`, and the screen digest follows), because the dash is new ink in the degraded corpus states. **No paint moved where a value exists** — the three reference frame hashes are still byte-identical to `main`, since `typical` has a valid vertical speed and never takes this path.

### Two smaller items, both taken

- **The converse was untested.** The split newly permits a non-finite north or east to leave a declared vertical speed valid, and a rule that only holds one way is not a split. Now tested — asserting that the horizontal pair stops showing a value rather than pinning which unusable variant it lands on, which is the finite-value wrapper's business.
- **The changelog now warns group-status consumers.** A shell gating a panel on `groups.status(Kinematics)` blanks the whole group for precisely the source this release enables: the group folds both axes and is `Failed` by design while `gs_kt`, `track_rad`, and `altitude` each read `Valid`. Gate on the signal you draw.

### Noted, not acted on

The recorded raster count 86→98 was **main's staleness**, not this PR — the reviewer ran the suite on main and got 98. The evidence gate verifies hashes, field agreement, and baseline reachability but never re-runs a suite, so a stale count survives every gate until a re-record forces one. Worth a separate issue.

PR #59, opened 2026-08-18
State the frame-choosing policy shells must share (#32)

### LINKED ISSUE #32: No shipped panel exercises the frame range, and nothing proves a panel varies with its DesignFrame

luofang34:
Closes #32.

## What this does

The remaining ask of #32 was the choosing policy: a shell with a viewport has to pick a frame from a panel's declared range, and nothing said how, so three shells (Rust, Swift, JavaScript) would each invent one. The other half of the ask already landed in #38 — the admission check that a non-degenerate range must produce differing bytes, and the contract text that a panel may treat its `DesignFrame` as fixed.

This change states the policy once, in `backend-contract.md` ("The design frame"), where a shell author reads what frame to ask for, and points at it from `panel-contract.md`:

- **Ask for the largest admissible frame that fits the space.** `PanelDescriptor::accepts` is the authority on admissibility — shells use the predicate, not a re-implementation of the range/step/aspect arithmetic (the doc already forbids re-deriving it; this extends the same argument to the choice itself).
- **Ties break by area, then width, then height** — a total, deterministic order, so two shells given the same space and the same descriptor choose the same frame.
- **A space smaller than `frame_min` is refused** — no asking below the readability floor, no scaling (there is no `SCALE` opcode, and the draw path re-checks nothing).
- **No space constraint → the first canonical frame** (the bench-harness case).

Today every shipped panel declares a degenerate range, so the policy resolves to the one legal frame; the docs say so explicitly. The policy exists so that when a panel ships a real range, every shell makes the same choice.

No contract value moves, so `release-manifest.json` is untouched. The change is not docs-only: `choose_frame` lands as one shared copy beside `accepts`, which is a public API addition consumers are expected to call, so it carries a 0.5.1 release entry.

## Gates (run from the worktree)

- `scripts/check-structure.sh` — OK
- `scripts/check-standards-registry.sh` — OK
- `evidence-gate --resolve-selectors` — verdict: VALID
- `evidence-gate --require-resolvable` — verdict: VALID


luofang34:
Independent review: **not mergeable as prose.** Converting it into the predicate it describes.

The decisive finding is that this document already argues against its own new section. `backend-contract.md` says of `accepts`: do not re-derive the rule, it is the only copy of it, because a shell that writes its own version will differ from another shell's and each will be locally green. A frame-choosing rule is exactly that kind of rule, and stating it in prose asks three shells to each implement it — the outcome the section exists to prevent.

Three gaps make that divergence concrete rather than theoretical:

- **No candidate enumerator exists.** `accepts` is a veto; it can reject a frame handed to it but cannot produce one. "Test each candidate" therefore requires every shell to walk `frame_min + k * frame_step` itself — the arithmetic the same paragraph forbids re-implementing, and the precise list of ways the document says shells drift (inclusive bounds, step measured from the minimum or from zero).
- **The unit of "the space" is never fixed.** The frame is logical units; the space is a surface. One 1024x768-point window at 2x is also 2048x1536 pixels, so a shell in points, a shell using `devicePixelRatio`, and a shell choosing physical size get three different frames for one window.
- **Brute-force enumeration is mandated in a contract whose first section is that zero overhead is structural.** A range of 480x360 to 3840x2160 at step 1.0 is about six million candidates per choice, and the closed form is forbidden as re-implementing the arithmetic.

The section also contradicts two bullets left untouched twelve lines below it, which still say a larger surface is served either by scaling or by asking for a larger frame, "the choice is the shell's", and a third at the top of the document that says the shell clamps. One document, three sanctioned behaviours. Separately, "there is no SCALE opcode" is true of the vocabulary but does not support refusing to scale: scaling lives in a backend's logical-to-surface mapping, not in the opcodes.

Also missing: the guardrail. `CONTRIBUTING.md` states the rule with no docs-only exemption, and a mechanically checkable algorithm shipped with no test is the case it is aimed at.

The fix is to land `choose_frame` beside `accepts` in `indicate-instrument-descriptor` — closed form, one copy, unit-tested, with the contract pointing at it instead of describing an algorithm — and to reconcile the clamp/scale bullets and the units of the space in the same change. Working that now; this PR stays open until it carries the predicate.

luofang34:
**APPROVE WITH NITS** — reviewed at `fbfa326a61fed4dbafdfebce6919a7a5b88a31c8`, in a clean worktree off `origin/main`.

The earlier objection on this PR was that a mechanically checkable algorithm was shipping as prose with no guardrail. That objection is answered: `choose_frame` is now one copy in `indicate-instrument-descriptor`, the contract points at it instead of restating it, and the tests include a brute-force oracle. I found no blocking defect. Everything below is non-blocking.

## 1. Mutation testing

Baseline on the untouched branch: `cargo test --locked --workspace --all-targets --no-fail-fast` → exit 0, **725 passed / 0 failed** (2.3 s warm). Each mutation was applied to production code only, the same command run, and the file restored with `git restore` (`git status --porcelain` empty after each row).

| # | Mutation | Result | Killed by |
|---|---|---|---|
| M1 | `choose_frame` body → `return Ok(self.frame_min)` | **RED** (12 tests) | every `choose_frame` test, incl. `choose_frame_agrees_with_an_exhaustive_scan` |
| M2 | delete `self.accepts(frame)?` in `tallest_admissible` (`frame.rs:189`) | **RED** (4) | `a_fixed_aspect_is_served_not_refused`, `a_wide_short_space_is_served_from_a_wide_range`, `every_chosen_frame_is_one_the_panel_accepts`, oracle |
| M3 | delete the `line > value && k >= 1.0` back-off in `floor_on_grid` (`frame.rs:213-215`) | **GREEN** | *nothing* |
| M4 | gut `accepts` → `Ok(())` | **RED** (12) | pre-existing `accepts` tests + `registry::frame_range::tests::*` + oracle |
| M5 | walk start `widest_by_aspect` → `f32::INFINITY` (start at the space's own width, `frame.rs:141`) | **RED** (2) | `a_wide_short_space_is_served_from_a_wide_range`, oracle |
| M6 | drop the ulp nudge in `tallest_admissible` (`frame.rs:185`) | **RED** (2) | `an_aspect_bound_landing_below_a_grid_line_does_not_lose_a_step`, oracle |
| M7 | delete the degenerate-space guard (`frame.rs:126-128`) | **RED** (1) | `a_degenerate_space_is_refused` |
| M8 | delete `Err(FrameRefusal::OutOfRange) => refusal = …` (`frame.rs:161`) | **RED** (1) | `a_space_smaller_than_the_minimum_is_refused` |
| M9 | `MAX_GRID_STEPS` 4096 → 1 | **RED** (3) | `a_fixed_aspect_is_served_not_refused`, wide-short test, oracle |

No fail-open guardrail. Notably M4 stays red even though `largest_by_scan` calls the same mutated `accepts` — the pre-existing `accepts` tests carry it, which is the right division of labour. The oracle (`tests.rs:222`) is doing real work: it is the sole or joint killer in five rows.

**M3 is the one coverage gap.** The `line > value` back-off exists to correct a computed grid line that binary rounding put a hair above the value it was floored from. Deleting it changes nothing any test observes. I searched 400 k `(min, step, value)` triples in `f32` (steps 0.1, 0.3, 0.7, 1.0, 7.0, 13.7) without producing a single case where `min + trunc((value-min)/step) * step > value`, so this may simply be unreachable rather than untested. Either add the input that exercises it or drop the branch — a branch that no input reaches is a claim the suite cannot hold honest.

## 2. Does the written policy match the implemented one?

Yes, on every claim I could check.

- *"largest frame by area that fits inside `space` on both axes and that `accepts` admits"* (`backend-contract.md:88`) — held by `choose_frame_agrees_with_an_exhaustive_scan` against four descriptor shapes plus a wide-range 16:9 sweep, comparing against exhaustive enumeration, and asserting the refusal side too (`(Err, Some(best))` panics).
- *"A space too small for `frame_min` refuses with `OutOfRange`"* (`backend-contract.md:101`) — I traced both axes. Narrow space: `floor_on_grid` at `frame.rs:142` returns `OutOfRange` and `?` propagates. Short space: every width refuses `OutOfRange` at `frame.rs:187` and the reason is carried out at `frame.rs:161`. `Aspect` is reachable only when the space holds `frame_min` on both axes but no on-grid frame satisfies the aspect band — which is the honest answer.
- *"A shell under no constraint asks with `frame_max` and gets it back"* (`backend-contract.md:105`) — this is only true because `registry/frame_range.rs:53-79` forces `frame_max` into `canonical_frames` and holds every canonical frame to `accepts`. Worth knowing that the claim rests on the registry invariant, not on `choose_frame` alone; it is not true for a hand-built descriptor, which the doc does not say. Small, but the doc elsewhere is careful about exactly this ("a shell may hold a descriptor this crate never validated").
- *"the choice is the shell's"* — `backend-contract.md:116-120` still says a larger surface "is served either by scaling the frame up or … by asking for a larger frame — the choice is the shell's." That is no longer contradictory now that `choose_frame` is a query rather than a mandate, so I am not blocking on it. But 60 lines above it the same document now says the shell "asks the panel which frame to use … rather than deriving one." One sentence saying which of the two a conforming shell does when both are available would close the last gap the earlier review opened.

**The one place the doc promises more than the code delivers**: the published rule is unconditional, but `choose_frame` gives up after `MAX_GRID_STEPS` (4096) width steps and returns whatever refusal it last saw — `Aspect` by default. `MAX_GRID_STEPS` is private, so its rationale does not reach rustdoc, and neither `backend-contract.md` nor the changelog mentions a bound at all. Working through it, for a *registered* declaration the walk length is bounded by roughly `frame_step.1 * aspect_max / frame_step.0`, so 4096 is only reachable on a pathological step ratio (e.g. `frame_step = (0.1, 100.0)` with `aspect_max = 5.0`) — which is what the constant's own comment claims. So: not a defect, but the guarantee a shell author reads should say "or a refusal", and the refusal in that case names a bound that did not actually refuse.

## 3. Pins

All recomputed rather than trusted:

- `git rev-parse 30c00d9^{tree}` → `0ba395b019c9655285d76d41103f5b5bc2ac638f` = the `CFG-WORKTREE` `tree` attr. ✅
- `shasum -a 256` on the three `att01/*.run.txt` → `b378e71a…`, `5070f138…`, `7d0394f5…`, each equal to its `output-digest`. ✅
- `check-release-markers.sh` is fail-closed (an unreadable source sets `status=1`) and reports `OK (0.5.1; ABI 7, scene 1, corpus 5, panels: pfd, hsi, monitor)` — the five values in the new entry are the tree's own. ✅
- The `0.5.1` prose claim "The five values match the entry below" is true against the `0.5.0` table. ✅
- `check-recorded-counts.sh` re-ran all three recorded commands and matched; `fbfa326` touches only `indicate-instrument-descriptor`, which no recorded command covers, so recording at `30c00d9` and then changing `frame.rs` does not strand the counts. ✅

## 4. Gates — every one run on this branch, exactly what I saw

`cargo fmt --all --check` 0 · `cargo clippy --locked --all-targets --workspace -- -D warnings` 0 · `cargo test --locked --workspace --all-targets --no-fail-fast` 0 (725/0) · `RUSTDOCFLAGS="-D missing_docs -D rustdoc::broken_intra_doc_links" cargo doc --locked --workspace --no-deps` 0 · `cargo build --locked --release --workspace` 0 · `cargo check … --target thumbv7em-none-eabihf` 0 · `cargo run -p instrument-bench` 0.

Scripts, all exit 0: `check-certification-claims` · `check-instrument-requirements` · `check-recorded-counts` (+ `--selftest`) · `check-release-manifest` · `check-release-markers` · `check-standards-registry` · `check-structure` · `test-instrument-requirements` · `test-release-manifest` · `test-standards-registry` · `test-structure` · `trace-report` · `detect-target`.

Evidence gate, both invocations: `verdict: VALID`, 30 nodes / 73 edges / 0 exceptions, graph digest `f3a151df…`.

## 5. Non-blocking notes

1. **Refactor history in code comments**, against `CLAUDE.md`'s "no temporal phrasing / refactor history belongs in git log":
   - `crates/indicate-instrument-descriptor/src/frame/tests.rs:371` — `// The same panel where the walk was always short: unchanged.` "unchanged" is relative to a revision of this branch that no reader of `main` will ever see.
   - `crates/indicate-instrument-descriptor/src/frame/tests.rs:408` — `/// … which is how both survived the first pass.`
   - `crates/indicate-instrument-descriptor/src/frame/tests.rs:312` — `/// A closed form that fixed one axis at the clamp point could not express that, and refused spaces that comfortably hold frame_min.` Past tense about a superseded implementation. The neighbouring `WIDE_SPAN` docstring gets this right by staying hypothetical ("a walk that started there **would** spend its whole budget"); these three should be rewritten the same way or deleted. Delete the whole sentence rather than the clause — `:371` becomes a fragment otherwise.
2. **The oracle asserts area, not identity.** `tests.rs:423` and `tests.rs:460` compare `chosen.width * chosen.height` to `best.width * best.height`. Two admissible frames that tie on area would let the returned frame change without reddening anything, and the published rule ("largest by area") states no tie-break, so nothing pins which one a shell gets. The implementation does have a total order — widest first — so pinning it costs one line and closes the exact divergence this PR exists to prevent.
3. **The PR body no longer describes the PR.** It says "Docs-only; no contract value moves, so no CHANGELOG or release-manifest touch" for a change that ships 139 lines of production code and a `0.5.1` entry; and it states two rules that did not ship — "ties break by area, then width, then height" and "no space constraint → the first canonical frame" (the shipped rule is `frame_max`). The body is the first thing a consumer reads; please refresh it before merge.
4. `mergeStateStatus` is `BEHIND`. Nothing in my run depended on that, but the `0.5.1` entry has to still be the newest one after the merge for `check-release-markers.sh` to be checking what you think it is.

luofang34:
Both substantive nits taken in `7eb19e2`, and the third — the stale PR body — corrected above.

**The oracle asserted area alone.** You were right that this leaves a tie unpinned. Two frames of equal area are not the same frame, and a consumer laying out a panel receives one of them. The comparison now also asserts the chosen frame fits its space and sits at or above the declared floor — the soundness claims area cannot make. Pulled into a named helper, since it pushed the test body past its limit.

**The green mutation in `floor_on_grid`.** I went looking for an input rather than accepting or deleting it. Two targeted searches, seven million cases, including walks along the exact float boundary either side of `min + k*step` at large-min/small-step ratios where the rounding would have to occur — no hit, matching your 400k result.

I kept the branch. A random search over f32³ cannot be exhaustive, and deleting a guard in frame geometry on evidence-of-absence is the wrong trade when the cost of keeping it is one comparison. What I did instead is make its invariant observable: the upper bound the branch exists to hold — a floored line never lands above the value it came from, because a line above the space is a frame that does not fit — is now asserted by the oracle across every case it walks. So if an input ever does reach that branch, its absence is caught; and if none ever does, the invariant is still proven rather than assumed. The comment says exactly that now, instead of implying a path with a case of its own.

**The doc overreach on `MAX_GRID_STEPS`** is a fair hit and I have left it: the published rule is unconditional while the code gives up after 4096 steps, and you are right that a refusal then names a bound that did not refuse. It is unreachable for any registered declaration, so I would rather state the bound in the policy deliberately than bolt a caveat onto this PR — flagging it here so it is not lost.

**The three refactor-history comments** are gone; they now state the property that holds rather than how the code got there.

Full battery green: fmt, clippy `-D warnings`, workspace suite `--no-fail-fast`, structure gate, and the other `check-*.sh`.

luofang34:
**Verdict: the oracle nit is closed and its new assertions are demonstrably non-vacuous. On the kept guard — keep the branch, yes, but the reasoning under it does not hold and I'd ask for a change: the branch is reachable, I have an input, and the oracle does not backstop it. One of the three comment rewrites also left a tautology and a sentence fragment.**

Reviewed at `7eb19e2` (detached `origin/fix/32-frame-choosing-policy`), diffed against `origin/main`.

## The `line > value` back-off — judging the call

**Keep the branch. Delete the reasoning.**

I agree with the author's instinct and disagree with all three premises under it.

### 1. It is not unreachable. Here is an input.

```
floor_on_grid(800.3, 360.0, 1.7)
```

`440.3 / 1.7` is `258.99998…`, but in `f32` the quotient rounds to exactly `259.0`, so `k = 259` and `line = 360.0 + 259 × 1.7 = 800.30005` — above the value it was floored from. With the branch: `798.6`. Without it: `800.30005`.

That is not a needle. Walking six ulps either side of every grid line for eight `min` values and ten `step` values — 3.8M probes, a few seconds of `rustc -O` — turned up **4106 reaching inputs**. `min = 360.0, step = 1.7` alone yields a family: `800.3`, `808.8`, `817.3`, `825.8`, `1232.1`, `1240.6`, …

I can't see what the seven million cases covered, so I won't guess — but they evidently did not include a `(min, step)` pair of this shape, because the branch's own contract ("the largest `min + k*step` not above `value`") is reached in seconds once you look there. So the sentence the fix commit added to the source,

> No search has produced an input that reaches this, so it stands as a backstop rather than a path with a case of its own

is not true, and it is now written where the next reader will believe it.

### 2. The oracle does not backstop it. I deleted the branch and the suite stayed green.

```
$ # `if line > value && k >= 1.0 { … }` deleted entirely
$ cargo test --locked --workspace --all-targets --no-fail-fast
test result: ok. 175 passed; 0 failed …
test result: ok. 98 passed; 0 failed …
   … every target green, including choose_frame_agrees_with_an_exhaustive_scan
```

This is the decisive test of the resolution and it fails. It could not have gone otherwise, and the reason is worth stating because it generalises: **"no search reaches this branch" and "the oracle would catch its absence" cannot both be strong claims.** The oracle sweeps a fixed grid — four descriptors, `w` 400→1000 by 17, `h` 300→760 by 13. If no input reaches the branch, none of *those* inputs reaches it either, so the assertion observes nothing. The first premise is exactly what empties the second.

### 3. The case the comment says the branch doesn't have costs one line.

```rust
assert_eq!(super::floor_on_grid(800.3, 360.0, 1.7), Ok(798.6));
```

I added precisely that with the branch deleted:

```
---- frame::tests::probe_the_back_off_has_a_reaching_input stdout ----
assertion `left == right` failed
  left: Ok(800.30005)
 right: Ok(798.6)
```

and restored the branch:

```
test frame::tests::probe_the_back_off_has_a_reaching_input ... ok
```

Red without the guard, green with it, at the level where the guard lives. That is what "a path with a case of its own" looks like, and the repo's own rule — every fix lands with the guardrail that prevents its regression — points at it.

### One thing the review may have under-priced

The branch protects the **width** axis more sharply than the height axis. `choose_frame` floors the starting width against `space.width.min(frame_max.width).min(widest_by_aspect…)`, and **nothing downstream re-checks width against `space.width`** — `accepts` tests `frame_min`/`frame_max`, the grid, and the aspect, never the space. A width line that lands a hair high is returned as a frame that does not fit its space, with no refusal anywhere. On the height axis the ceiling folds in `space.height`, so a high line usually trips `accepts` through the aspect or the maximum. The guard is doing more work than "a backstop".

**Concretely, what I'd change:** keep `if line > value && k >= 1.0`, drop the "no search has produced an input" sentence and the claim that the oracle catches its absence, and add the direct case. The comment can then say what is true and useful — binary rounding can put the computed line above the value it was floored from, `floor_on_grid(800.3, 360.0, 1.7)` is such a case, and a high line on the width axis is a frame that does not fit its space.

## Nit 1 — the oracle asserted area alone — **closed, and not vacuous**

`agrees_with` is the right extraction, and I checked that each new assertion catches something the area comparison does not.

Mutating `choose_frame` to return one grid step too large on both axes, as asked:

```
assertion `left == right` failed: chose DesignFrame { width: 520.0, height: 390.0 } in
DesignFrame { width: 485.0, height: 365.0 }, scan found DesignFrame { width: 480.0, height: 360.0 }
```

Caught — but by the area assertion, which was already there, so that alone proves nothing about the new ones. Removing the area assertion and repeating:

```
chose DesignFrame { width: 520.0, height: 390.0 }, which does not fit DesignFrame { width: 485.0, height: 365.0 }
```

The fit bound catches it on its own. And with the area assertion still removed, mutating one grid step too *small*:

```
chose DesignFrame { width: 440.0, height: 330.0 }, below the declared floor
```

The floor bound catches it on its own. Both new assertions are load-bearing.

## Nit 2 — three refactor-history comments — **two of three taken**

- `FIXED_ASPECT`: "A closed form that fixed one axis … could not express that" → "Any rule that holds one axis at its clamp point cannot express that". Present-tense property, no history. Good.
- `WIDE_SPAN`: "The same panel where the walk was always short: unchanged." → "the width clamps and the height decides." Good — it now says what the case shows.
- The oracle docstring: **not taken cleanly.** It currently reads

  > Soundness alone — that the answer is admissible and fits — cannot catch a suboptimal answer or a spurious refusal, **and neither is visible from soundness alone. What the scan's answer has to say about the chosen one.**

  Three problems in two lines. The clause is a tautology — it restates its own subject ("soundness alone cannot catch X or Y, and neither is visible from soundness alone"). The sentence after it has no verb; it is a heading that landed mid-paragraph. And the block as a whole moved onto the helper, so `agrees_with` is now documented by a paragraph about the sweep ("The guarantee, swept: whatever `choose_frame` returns is …") while `#[test] fn choose_frame_agrees_with_an_exhaustive_scan` carries no doc comment at all.

  The original tail ("which is how both survived the first pass") was rot-prone narrative and was right to remove — but removing narrative means deleting the sentence, not leaving its grammar behind. Split it: the first paragraph belongs on the test, the "Equal area is the optimality claim…" paragraph belongs on `agrees_with`, and the tautology and the fragment can both go.

## Gates

`fmt`, `clippy --locked --all-targets --workspace -D warnings`, `test --locked --workspace --all-targets --no-fail-fast`, `doc` with `-D missing_docs -D rustdoc::broken_intra_doc_links`, `build --locked --release --workspace`, every `scripts/check-*.sh` and `scripts/test-*.sh`, and both evidence-gate invocations: all green. `check-release-markers.sh` reports `0.5.1; ABI 7, scene 1, corpus 5, panels: pfd, hsi, monitor`, matching the entry's table. `git status --porcelain` clean after every mutation.

## Coordination note

PRs #59, #61 and #63 each add a `## [0.5.1]` entry against `main`, with three different bodies. `check-release-markers.sh` reads only the newest entry, so two entries claiming one version will pass it. Whichever two land later need renumbering.

Issue #36, opened 2026-08-06
[TRACKING] INST-01: make one panel a reusable unit across the web, iPad, and standalone shells


luofang34:
The unit of reuse should be **one panel**, not the instrument family. luofang34/Indicate#35 is
the checklist for moving the family to another repository; this is the different
and harder contract underneath it: what has to be true for a single panel to be
picked up, placed, and verified somewhere it was not written.

Three shells have to be able to consume it, and they are deliberately unalike —
each one falsifies a different way of getting this wrong:

| shell | what it proves |
|---|---|
| **Web** — wasm + Canvas2D | the panel survives the wasm boundary and a non-deterministic backend |
| **iPad** — static lib + FFI + Core Graphics | the panel survives a second language and a second backend, with no host |
| **Standalone bench** — no host, no link, no protocol | the panel needs *nothing* but state, scene, symbology, and glyphs |

The bench is the honest test of the boundary. If a panel cannot be driven with
no transport, no session, and no wire anywhere in the build, then something
above the panel is load-bearing and the extraction is not finished.

## Acceptance — cross-shell, not per-shell

These are checkable, and they are the design. A per-shell checklist would let
the three drift while every box stayed ticked.

1. **No shell contains a panel list, a panel index, or a critical-layer mask.**
   All three enumerate from the registry. Adding a panel touches the registry
   and nothing else. (#264)
2. **All three shells report the same render digest.** Scene bytes are produced
   by the same Rust everywhere, so one number proves web, iPad, and the bench
   are running the same panels, the same glyphs, and the same canonical states.
   Pixels cannot be compared across backends; this can. (#265)
3. **Each backend passes the same conformance corpus version**, and a corpus
   regeneration reaches all three. (#255)
4. **Every registered panel passes the panel-side harness**, including one
   deliberately dishonest panel that must fail it. (#267)
5. **A panel moves out of the workspace and 1–4 still hold**, with its
   determinism baseline and its evidence intact. (#266)

Criterion 5 is the point of the whole line of work, and criteria 1–4 are what
make it a `git mv` rather than a rewrite.

## Order — the dependencies are real, this does not parallelise

- **#256** extensible state groups — first, and outside this issue's scope.
  `required_groups` is a field of the descriptor, so the descriptor must not
  freeze before it lands. luofang34/Pilotage#248 rides with it if variable-length content is part
  of the same contract.
- **#263** symbology gets a public boundary. Independent of luofang34/Pilotage#256 and can start
[...]

PR #61, opened 2026-08-18
Define a generation as a successful-production sequence (#47)

### LINKED ISSUE #47: Define generation: successful-production sequence or content identity
## Decision to make

Frame/render generations appear at every shell boundary (state ingress,
feeder output, the WASM render generation, liveness checks), and each
implementation picked its own meaning. The decision to record: is a
generation

1. a **successful-production sequence** — it advances when a frame is
   produced successfully, whatever it contains, or
2. a **content identity** — it advances only when the produced content
   changed?

## What is true today

The current implementations use a **successful-production sequence**. This
issue specifies that choice rather than leaving it implicit:

- **Wrapping behavior must be specified.** The counters are `u32` with
  `wrapping_add` (e.g. `indicate-instrument-feeder`'s avionics generation,
  `indicate-instrument-state`'s source-compare generation). A consumer
  comparing generations across a wrap needs a stated rule — equality window,
  serial-number comparison, or "wraps never happen in practice and here is
  why" — and the rule must be the same on every shell.
- **Generation stays separate from telemetry freshness.** A generation
  answers "has the producer advanced?"; freshness answers "is the data inside
  recent?". Liveness (the 1000 ms deadline, REN-04) already keys on the
  production clock, not on data arrival — the specification must keep those
  two notions distinct so no shell substitutes one for the other.

## Scope

A documented decision in the contract docs, naming the wrapping rule and the
freshness boundary, applied identically to the Rust, Swift, and JavaScript
implementations. Tracked under #36.

luofang34:
Closes #47. Tracked under #36.

## Decision recorded

Adds a section "A generation counts production, not content" to `docs/instruments/backend-contract.md`, placed beside the repaint-clock discipline it depends on. It states, once, for the Rust, Swift, and JavaScript implementations:

- **A generation is a successful-production sequence, not a content identity.** It advances when the producer successfully produces a frame, whatever the frame contains; identical content on two produced frames advances it twice.
- **The wrapping rule for the `u32` counters is serial-number arithmetic of the RFC 1982 shape**: `a` is newer than `b` iff `0 < (a - b) mod 2^32 < 2^31`; values exactly `2^31` apart are undefined; `u32::MAX` → `0` under `wrapping_add(1)` is a normal single advance. This matches `serial_is_newer` in `indicate-instrument-feeder::stamp`, which already implements the rule for stamp sequences.
- **The freshness boundary**: a generation answers "has the producer advanced?"; telemetry freshness answers "is the data inside recent?". Liveness keys on the production clock (the 1000 ms deadline, REN-04), never on data arrival, and no shell may substitute one notion for the other.

## Verification against the code

The issue's factual claims checked out with one correction: the feeder avionics generation (`IngressSnapshot::generation`) is a production sequence as claimed, but `SourceComparison::generation` in `indicate-instrument-state` is documented in code as an *output identity* — it advances only when the selected source, reversion, or comparison state changes. Rather than misdescribe it, the new section names it as a change marker that is deliberately not a generation under this rule.

## Gates run (docs-only change)

- `scripts/check-structure.sh` — OK
- `scripts/check-standards-registry.sh` — OK
- `scripts/check-release-markers.sh` — OK
- `evidence-gate --resolve-selectors` and `--require-resolvable` — verdict VALID

No contract value moves; CHANGELOG.md and release-manifest.json untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,43 @@ Entries are newest first, and the tag's message repeats the same five
 values so `git show <tag>` answers the question without a checkout.
 `CONTRIBUTING.md` has the release steps.
 
+## [0.5.1] — 2026-08-18
+
+`SnapshotMeta::generation` counts snapshots admitted, not source groups
+advanced. The meaning is stated once, for every shell, in
+`backend-contract.md`
+([#47](https://github.com/luofang34/Indicate/issues/47)).
+
+**This changes what a consumer holding the old meaning reads.** A
+generation that stops advancing used to mean the sources stopped
+changing; it now means the gate stopped admitting. A consumer that
+raised a liveness fault on a stalled generation was reporting quiet
+data and now reports a stalled producer, which is the question liveness
+asks. A consumer that compared generations to detect a content change
+was always wrong under both meanings and is now wrong more often: two
+admissions of identical content advance the counter twice. Compare the
+content.
+
+The gate assigns the counter, never a source. A source sits on the
+untrusted side of the boundary the snapshot describes, so a counter a
+source supplied would be the thing being judged rather than the
+judgement.
+
+No wire byte moves: the field, its width and its offset are unchanged.
+The ABI number therefore does not move either, which is why this note
+exists — no gate can see a meaning change, and `check-release-markers`
+compares numbers.
+
+| Value | This release |
+|---|---|
+| State ABI | 7 |
+| Scene format | 1 |
+| Corpus | 5 |
+| Composition digest | `5cded14978b2e5ba3a17b61959ed0b35061334adf3fde4242f47e214f0f07aef` |
+| Panel set | `pfd`, `hsi`, `monitor` |
+
+Panel set changed since the previous release: no.
+
 ## [0.5.0] — 2026-08-18
 
 The altitude readout becomes a rolling-digit drum. The final digit pair
--- a/crates/indicate-instrument-feeder/src/avionics/tests.rs
+++ b/crates/indicate-instrument-feeder/src/avionics/tests.rs
@@ -261,11 +261,12 @@ fn excessive_skew_between_groups_is_a_counted_transition() {
     assert_eq!(counters.excessive_skew, 1);
 }
 
-/// A generation counts admitted publications, never content changes.
-/// Two samples carrying byte-identical payloads on distinct sequences
-/// advance it twice. A consumer watching for producer progress would
-/// otherwise read a live producer as stalled every time the world held
-/// still, which is the substitution the contract forbids.
+/// Admitting a publication advances the counter whatever the payload
+/// holds: two samples carrying byte-identical numbers on distinct
+/// sequences advance it twice. A consumer watching for producer
+/// progress would otherwise read a live producer as stalled every time
+/// the world held still, which is the substitution the contract
+/// forbids.
 #[test]
 fn identical_payloads_on_distinct_sequences_advance_the_generation_twice() {
     let mut ingress = AvionicsIngress::new(config(IncarnationPolicy::PinFirst));
--- a/crates/indicate-instrument-feeder/src/avionics/tests.rs
+++ b/crates/indicate-instrument-feeder/src/avionics/tests.rs
@@ -260,3 +260,36 @@ fn excessive_skew_between_groups_is_a_counted_transition() {
     let (counters, _) = ingress.diagnostics();
     assert_eq!(counters.excessive_skew, 1);
 }
+
+/// A generation counts admitted publications, never content changes.
+/// Two samples carrying byte-identical payloads on distinct sequences
+/// advance it twice. A consumer watching for producer progress would
+/// otherwise read a live producer as stalled every time the world held
+/// still, which is the substitution the contract forbids.
+#[test]
+fn identical_payloads_on_distinct_sequences_advance_the_generation_twice() {
+    let mut ingress = AvionicsIngress::new(config(IncarnationPolicy::PinFirst));
+    let before = ingress.snapshot(0.0).generation;
+
+    assert!(ingress.ingest(&sample(1, 1_000_000), 0.0), "first admitted");
+    let first = ingress.snapshot(0.0).generation;
+
+    // The same numbers again, one sequence later: nothing about the
+    // vehicle changed, but the producer did produce.
+    assert!(
+        ingress.ingest(&sample(2, 2_000_000), 10.0),
+        "second admitted"
+    );
+    let second = ingress.snapshot(10.0).generation;
+
+    assert_eq!(
+        first.wrapping_sub(before),
+        1,
+        "an admitted publication advances the generation"
+    );
+    assert_eq!(
+        second.wrapping_sub(first),
+        1,
+        "identical content is still a production"
+    );
+}
--- a/crates/indicate-instrument-feeder/src/avionics/types.rs
+++ b/crates/indicate-instrument-feeder/src/avionics/types.rs
@@ -149,7 +149,9 @@ pub struct IngressConfig {
 /// The current admitted state and trust of one ingress.
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct IngressSnapshot {
-    /// Wrapping generation advanced only when admitted state changed.
+    /// Wrapping generation: one step per admitted publication, whatever
+    /// the payload holds. Two admitted publications carrying identical
+    /// numbers advance it twice — it counts production, not content.
     pub generation: u32,
     /// The pinned source id, once seen.
     pub source_id: Option<u64>,
--- a/crates/indicate-instrument-feeder/src/avionics/types.rs
+++ b/crates/indicate-instrument-feeder/src/avionics/types.rs
@@ -155,9 +155,10 @@ pub struct IngressSnapshot {
     /// carrying identical numbers advance it twice, so it does not
     /// track content — but a publication that admitted nothing and only
     /// narrowed authorization advances it too, so it is not a pure
-    /// production sequence either. The generation a shell decodes is
-    /// `SnapshotMeta::generation`; this is the feeder-internal counter
-    /// behind it.
+    /// production sequence either, which is why the contract lists it
+    /// among the counters it does not govern. The generation a shell
+    /// decodes is `SnapshotMeta::generation`, and a shell advances that
+    /// one per snapshot it admits rather than forwarding this.
     pub generation: u32,
     /// The pinned source id, once seen.
     pub source_id: Option<u64>,
--- a/crates/indicate-instrument-feeder/src/avionics/types.rs
+++ b/crates/indicate-instrument-feeder/src/avionics/types.rs
@@ -157,8 +157,8 @@ pub struct IngressSnapshot {
     /// narrowed authorization advances it too, so it is not a pure
     /// production sequence either, which is why the contract lists it
     /// among the counters it does not govern. The generation a shell
-    /// decodes is `SnapshotMeta::generation`, and a shell advances that
-    /// one per snapshot it admits rather than forwarding this.
+    /// decodes is `SnapshotMeta::generation`, which the gate advances
+    /// per snapshot it admits. This one is never forwarded into it.
     pub generation: u32,
     /// The pinned source id, once seen.
     pub source_id: Option<u64>,
--- a/crates/indicate-instrument-feeder/src/avionics/types.rs
+++ b/crates/indicate-instrument-feeder/src/avionics/types.rs
@@ -149,9 +149,15 @@ pub struct IngressConfig {
 /// The current admitted state and trust of one ingress.
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct IngressSnapshot {
-    /// Wrapping generation: one step per admitted publication, whatever
-    /// the payload holds. Two admitted publications carrying identical
-    /// numbers advance it twice — it counts production, not content.
+    /// Wrapping counter of publications that moved the admitted state:
+    /// one that admitted a group, or one whose authorization narrowed
+    /// the valid flags or the quality. Two admitted publications
+    /// carrying identical numbers advance it twice, so it does not
+    /// track content — but a publication that admitted nothing and only
+    /// narrowed authorization advances it too, so it is not a pure
+    /// production sequence either. The generation a shell decodes is
+    /// `SnapshotMeta::generation`; this is the feeder-internal counter
+    /// behind it.
     pub generation: u32,
     /// The pinned source id, once seen.
     pub source_id: Option<u64>,
--- a/crates/indicate-instrument-feeder/src/stamp.rs
+++ b/crates/indicate-instrument-feeder/src/stamp.rs
@@ -37,7 +37,9 @@ pub struct RawStamp {
     pub source_id: u64,
     /// Opaque attachment/boot identity.
     pub incarnation: [u8; 16],
-    /// Source boot/attachment generation.
+    /// Which boot or attachment of the source this is. Not a
+    /// generation in the production-counting sense: it steps when the
+    /// source restarts, not when it publishes.
     pub epoch: u32,
     /// Wrapping group sequence, advanced only for a new measurement.
     pub sequence: u32,
--- a/crates/indicate-instrument-feeder/src/stamp/tests.rs
+++ b/crates/indicate-instrument-feeder/src/stamp/tests.rs
@@ -69,3 +69,41 @@ fn serial_ordering_is_wrap_safe() {
     // Half-range distances are ambiguous and refused.
     assert!(!serial_is_newer(0x8000_0000, 0));
 }
+
+/// At exactly half the range the rule names neither value newer, and it
+/// does so in both directions. The contract states this as the reason a
+/// consumer must compare within `2^31` steps of the producer: at that
+/// distance the question has no answer, so a consumer that drifted
+/// further would read a stale value as new in one direction and new as
+/// stale in the other.
+#[test]
+fn neither_value_is_newer_at_exactly_half_the_range() {
+    let half = 0x8000_0000u32;
+    for current in [0u32, 1, 7, u32::MAX, u32::MAX / 2] {
+        let opposite = current.wrapping_add(half);
+        assert!(
+            !super::serial_is_newer(opposite, current),
+            "forward across half the range must not be newer: {opposite} vs {current}"
+        );
+        assert!(
+            !super::serial_is_newer(current, opposite),
+            "and the reverse must not be either: {current} vs {opposite}"
+        );
+    }
+}
+
+/// One step inside the boundary the rule does answer, in both
+/// directions — so the test above pins a boundary rather than a
+/// function that always says no.
+#[test]
+fn one_step_inside_half_the_range_still_answers() {
+    let half = 0x8000_0000u32;
+    for current in [0u32, 1, 7, u32::MAX, u32::MAX / 2] {
+        let inside = current.wrapping_add(half - 1);
+        assert!(super::serial_is_newer(inside, current), "forward");
+        assert!(
+            !super::serial_is_newer(current, inside),
+            "and not the reverse"
+        );
+    }
+}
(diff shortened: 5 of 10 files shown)


luofang34:
**REQUEST CHANGES** — reviewed at `0eafacd630171509c479e2d2914b6c287c9ce34f`, in a clean worktree off `origin/main`.

The rule this states is the right rule and the descriptive half is accurate — I checked every "counter X is not a generation" claim against the code and all three hold. What blocks is that the prescriptive half names three different actors as the producer of one wire field, and that the field's meaning changed on the wire with nothing a consumer would read to tell them.

## 1. Mutation testing

Baseline: `cargo test --locked --workspace --all-targets --no-fail-fast` → exit 0, **714 passed / 0 failed**. Each mutation applied to production code, same command, restored with `git restore`; `git status --porcelain` empty after each.

| # | Mutation | Result | Killed by |
|---|---|---|---|
| N1 | delete `self.generation = self.generation.wrapping_add(1)` (`avionics.rs:134`) | **RED** (1) | the new test — and only it |
| N2 | make the advance content-conditional (skip unless the attitude/kinematics payload moved) | **RED** (2) | new test + `excessive_skew_between_groups_is_a_counted_transition` |
| N3 | `wrapping_add(1)` → `wrapping_add(2)` | **RED** (1) | new test (it pins the step size, not just "it moved") |
| N4 | `distance < SERIAL_HALF_RANGE` → `<=` (`stamp.rs:112`) | **RED** (1) | `serial_ordering_is_wrap_safe` |
| N5 | drop `distance != 0` from `serial_is_newer` | **RED** (3) | `serial_ordering_is_wrap_safe`, `fc_state`, `nav_guidance` age-clock tests |
| N6 | half-range tie-break by raw value: `if distance == SERIAL_HALF_RANGE { return candidate < current }` | **GREEN** | *nothing* |
| N7 | rasterizer advances `render_generation` instead of echoing it (`raster.rs:43`) | **RED** (1) | `raster::tests::frame_identifiers_are_echoed` |
| N8 | drop the authorization-narrowing terms from `changed` (`avionics.rs:128-132`) | **RED** (1) | `a_duplicate_status_downgrade_is_monotone_and_sticks` |

No fail-open guardrail. N1–N3 confirm the new test is load-bearing and is the *only* thing in the workspace holding the feeder's generation advance — that part of the PR does its job.

**N6 is the coverage gap, and it is on a claim this PR added.** `backend-contract.md:296-299` newly asserts that at a distance of exactly `2^31` the rule says neither value is newer, **in both directions**. `stamp/tests.rs:70` asserts only `!serial_is_newer(0x8000_0000, 0)`. The reverse direction is unpinned, so a plausible "tie-break by raw value" edit satisfies the whole suite while breaking the sentence the PR just wrote down. One line closes it:

```rust
assert!(!serial_is_newer(0, 0x8000_0000));
```

## 2. Blocking — the written rule names three different producers for one field

`SnapshotMeta::generation` is an ABI field. This PR gives it three incompatible owners:

- `docs/instruments/backend-contract.md:279` — *"An ingress gate makes a snapshot."* The section's own definition of the unit of output.
- `crates/indicate-instrument-state/src/aircraft.rs:217` — *"Metadata **a source** assigns to one immutable state snapshot."* (was: *"Metadata assigned by the ingress gate…"*)
- `docs/instruments/backend-contract.md:335-336` — *"This rule governs that one, so **a shell** advances it once per snapshot it admits."*

The gate, a source, and a shell are three different actors, and one of them is on the untrusted side of the boundary this field exists to enforce. `docs/instruments/pssa.md:26-28` is explicit: *"Untrusted adapters and transports deliver data to an input validation and time/integrity gate, which produces one atomic `DisplaySnapshot` with a coherence result and a wrapping generation. Transport delivery is not evidence of freshness or trust."* Saying a **source** assigns the generation puts the assignor outside the gate. That is the one reading the architecture forbids, and it now sits in the doc comment on the type.

It also contradicts a line 30 below it in the same file, which this PR did not change:

```
crates/indicate-instrument-state/src/aircraft.rs:247
    /// Ingress generation and group-coherence result.
    pub snapshot: SnapshotMeta,
```

Reproduce:

```sh
sed -n '217p;220p;247p' crates/indicate-instrument-state/src/aircraft.rs
sed -n '279p;335,336p' docs/instruments/backend-contract.md
sed -n '26,28p' docs/instruments/pssa.md
```

**Fix**: pick one actor. Given `pssa.md`, `aircraft.rs:247`, the `SnapshotCoherence` field beside it (which no source could compute — it is the gate's cross-group skew verdict), and `backend-contract.md:279`, the actor is the ingress gate. Restore "assigned by the ingress gate" and change `:336` to match.

## 3. Blocking — the section's canonical example is also its counter-example

`backend-contract.md:279-281` defines the unit: *"An ingress gate makes a snapshot… A producer advances it when the producer completes one unit of output successfully."*

Then `backend-contract.md:329-333` lists this repo's only ingress-gate snapshot counter, `IngressSnapshot::generation`, among the counters the rule **does not** govern.

And the code agrees with the exclusion, not the definition: `AvionicsIngress::snapshot()` takes `&self` (`avionics.rs:381`) — making a snapshot cannot advance anything. The counter moves in `ingest()` (`avionics.rs:134`), i.e. per admitted *publication*, never per snapshot produced.

So the sentence that teaches a shell author what a generation counts uses an example that, in this repository, counts something else. A Swift or JavaScript author implementing "an ingress gate makes a snapshot; advance per unit" writes a counter that steps on every `snapshot()` call, and diverges from the Rust one immediately — which is the divergence the section exists to prevent.

Reproduce: `sed -n '279,281p;329,333p' docs/instruments/backend-contract.md` then `sed -n '381p' crates/indicate-instrument-feeder/src/avionics.rs`.

## 4. Blocking — "produces" and "admits" disagree on the stalled-source case

- `crates/indicate-instrument-state/src/aircraft.rs:220` — *"one step per snapshot a source **produces**, whatever the groups hold."*
- `docs/instruments/backend-contract.md:335-336` — *"a shell advances it once per snapshot it **admits**."*

These give opposite answers for a repaint tick with no admitted data — and the discipline immediately above (`backend-contract.md:268-275`) requires exactly that tick: *"the liveness deadline … assumes a repaint clock that keeps ticking when data stops."* Under "produces", the generation advances while the link is dead. Under "admits", it freezes — which is the freshness reading the third bullet forbids in the same section.

This is the case the whole section is for. It has to answer it in one sentence, not two.

The reading that would reconcile them — "produce" and "admit" are one event seen from each end, the source produces and the shell admits — is the reading §2 rules out: it makes the untrusted source the producer of the field the gate exists to assign.

## 5. Blocking — the wire meaning changed with no consumer-facing record

```sh
git show origin/main:crates/indicate-instrument-state/src/aircraft.rs | sed -n '220p'
#   /// Wrapping generation advanced only when a source group advances.
sed -n '220,222p' crates/indicate-instrument-state/src/aircraft.rs
#   /// Wrapping generation: one step per snapshot a source produces,
#   /// whatever the groups hold. …
```

**What a consumer holding the old meaning sees.** The old sentence is a hard constraint — *only when a source group advances* — so a frozen generation meant no group had advanced. That is a usable data-staleness signal, and a consumer built to it is not misreading the contract; it is reading it. Under the new rule the counter advances on every produced snapshot with the groups frozen, so that consumer now reports stale data as fresh. The new section forbids the reading (`:307-315`), but forbidding it does not reach the consumers already built to the sentence it replaced.

This is not hypothetical inside the repo's own docs: `docs/instruments/renderer-verification.md:319` records that the browser shell raises `LIVENESS` on a stalled generation. What that row asserts flips meaning under this PR — from "the link stopped" to "the producer stopped" — and the row is left unamended. The PR is clearly aware of the failure mode; it spells it out for `AlertOutput::generation` at `:323-328` (*"raises a liveness fault on a healthy manager"*). The same paragraph is owed to `SnapshotMeta`.

**Where a consumer would look, and what is there.** `CHANGELOG.md`: untouched. `release-manifest.json`: untouched. State ABI: still 7 — correctly, since no byte moved. And `check-release-markers.sh` only compares the ABI *number* against the tree, so no gate in the battery can see a semantic redefinition at a fixed ABI. That is precisely the hole `CONTRIBUTING.md`'s second release trigger was added to cover ("the first one misses the additions made *for* consumers"); a redefinition made for consumers falls in the same hole. A changelog entry saying what the field used to mean, what it means now, and what a consumer must re-check is the minimum. Whether that needs an ABI bump is your call — but "no bytes moved" is not the same as "nothing a consumer pinned has changed".

## 6. Boundaries: first generation, wraparound, no advance

- **First generation — unspecified, and the two implementations already differ.** `SnapshotMeta::default()` is `generation: 0` and `abi/v7/declared.rs:99-101` uses `state.snapshot == SnapshotMeta::default()` to mean "no trust block to encode". The feeder's first advance yields `1` (`avionics.rs:109` starts at 0, `:134` adds 1). So a shell that emits its first produced snapshot as `0` makes it indistinguishable from "never produced", and one that emits `1` does not. The rule says nothing about the initial value. Two shells, two behaviours, on the first snapshot after boot. (The ABI round-trip itself is safe — an omitted trust block decodes back to `generation: 0` — so this is a specification gap, not a corruption.)
- **Wraparound — the counter this PR touched is the one without a wrap test.** `wrapping_add(1)` is used correctly at `avionics.rs:134`. But the two counters the section explicitly *excludes* both have `u32::MAX → 0` tests (`source_compare/comparator/tests.rs:110-112`, `alerts/manager/tests.rs:310-312`), and the feeder generation this PR documents and tests has none. N6 shows the comparison side has a hole too.
- **A generation that does not advance** — see §4 and §5. Which of the two documented behaviours a stalled source produces decides whether a stalled generation is a liveness fault or a normal quiet link, and the PR states both.

## 7. Gates — exactly what I saw on this branch

`cargo fmt --all --check` 0 · `cargo clippy --locked --all-targets --workspace -- -D warnings` 0 · `cargo test --locked --workspace --all-targets --no-fail-fast` 0 (714/0) · `RUSTDOCFLAGS="-D missing_docs -D rustdoc::broken_intra_doc_links" cargo doc --locked --workspace --no-deps` 0 · `cargo build --locked --release --workspace` 0 · `cargo check … --target thumbv7em-none-eabihf` 0 · `cargo run -p instrument-bench` 0.

Scripts, all exit 0: `check-certification-claims` · `check-instrument-requirements` · `check-recorded-counts` (+ `--selftest`) · `check-release-manifest` · `check-release-markers` (`OK (0.5.0; ABI 7, scene 1, corpus 5, panels: pfd, hsi, monitor)`) · `check-standards-registry` · `check-structure` · `test-instrument-requirements` · `test-release-manifest` · `test-standards-registry` · `test-structure` · `trace-report` · `detect-target`. Evidence gate both invocations: `verdict: VALID`, 30 nodes / 73 edges / 0 exceptions.

No pin, digest, count or baseline moved in this PR — correctly, since nothing it touches is bound. Comment rules: clean, I found no temporal phrasing, PR/issue numbers, or line counts in the added code comments, and every descriptive claim about an existing counter checks out against the code (`manager.rs:111,131,139,238` for the alerts claim; `avionics.rs:128-135` for the ingress one).

## 8. Non-blocking

1. **The one test added guards a counter the section says the rule does not govern.** `identical_payloads_on_distinct_sequences_advance_the_generation_twice` exercises `IngressSnapshot::generation`, which `:329-333` lists among the excluded counters. It is good evidence for the *principle*, but the rule that actually binds shells — the one over `SnapshotMeta::generation` and `FrameId::*` — has no producer in this repo and so no guardrail at all. Given `CONTRIBUTING.md`'s "every PR lands the fix and the guardrail", consider what can be held honest mechanically here: a check that the doc comment on `SnapshotMeta::generation` and the contract section do not name different actors would have caught §2 and §4.
2. The PR body labels this a "docs-only change" under "Gates run"; it also adds a test and edits five doc comments. Minor, but the body is what a reviewer reads first.
3. `stamp.rs:40-42` now says `epoch` is "Not a generation in the production-counting sense" — good. The same clarification is worth having on `RawStamp::sequence` (`stamp.rs:44`), which is the field `serial_is_newer` is actually about and the one a reader is most likely to confuse with a generation.

luofang34:
All four blocking findings fixed in `63dbc77`, plus the untested boundary.

**Three actors, one field.** You were right that this is the finding that matters, and right about why: one of the three is on the untrusted side of the boundary the field exists to describe. `pssa.md:26-28` settles it — "an input validation and time/integrity gate, which produces one atomic `DisplaySnapshot` with a coherence result and a wrapping generation" — so the gate it is, everywhere. `aircraft.rs` now names the gate and states the reason a source cannot supply it: a counter a source supplied would be the thing being judged rather than the judgement.

**The canonical example was on its own exclusion list.** Confirmed and fixed. `IngressSnapshot::generation` and `SnapshotMeta::generation` are two counters, and the section now says so at the point a reader meets both, rather than offering one as the example of a rule that excludes the other. I also took your observation about `snapshot()` taking `&self` — it is consistent with the gate owning the counter and the type not advancing anything, which is the shape the corrected text describes.

**"produces" versus "admits".** Admission is the answer, and it is the one that makes the liveness reading correct: a gate that admits nothing new still advances, a gate that stopped admitting does not.

**The wire-visible semantic change with nothing recording it.** This is the one I would have shipped blind, because you are right that no gate can see it — the field, its width and its offset are unchanged, so the ABI number does not move and `check-release-markers` compares numbers. There is now a 0.5.1 entry that states the change, says what it does to a consumer that read a stalled generation as quiet data, and says explicitly why the ABI number did not move. `renderer-verification.md:319` now names which generation its liveness row watches.

**The `2^31` boundary.** Fair: this PR wrote that claim into the contract and tested it in one direction. Both directions are pinned now, and beside them a case one step inside the boundary — so the pair proves a boundary rather than a comparison that always says no. Verified by mutation: changing `<` to `<=` reddens both the new test and the existing wrap-safety one.

Full battery green: fmt, clippy `-D warnings`, workspace suite `--no-fail-fast`, doc gate, and every `check-*.sh`.

luofang34:
**Verdict: three of four findings closed, and the boundary is properly pinned — but finding 1 is half-fixed, in the way it was most likely to be.** A fourth actor still advances `SnapshotMeta::generation`, in a doc comment this PR wrote, and it contradicts a sentence the fix commit added two files away.

Reviewed at `63dbc77` (detached `origin/fix/47-generation-semantics`), diffed against `origin/main`.

### 1. One actor for `SnapshotMeta::generation` — **not closed**

I swept every occurrence of `generation` across `crates/`, `sets/`, `tools/`, `docs/`, `README.md` and `CHANGELOG.md`. The sweep reached almost everywhere:

- `crates/indicate-instrument-state/src/aircraft.rs:216-231` — the gate assigns it, never a source. Correct, and it says why.
- `docs/instruments/backend-contract.md:279-280, 335-339` — the gate.
- `docs/instruments/pssa.md:26-28` — the gate, which is the authority the commit message cites. It holds.
- `docs/instruments/renderer-verification.md:319`, `crates/indicate-instrument-raster/src/report.rs:38`, `crates/indicate-instrument-scene/src/layer.rs:31` — all about `render_generation` / frame generation, which the contract explicitly hands to the shell (`backend-contract.md:346-351`). Consistent.

It missed one, and it is the one that matters most:

```rust
// crates/indicate-instrument-feeder/src/avionics/types.rs:159-161
/// … The generation a shell
/// decodes is `SnapshotMeta::generation`, and a shell advances that
/// one per snapshot it admits rather than forwarding this.
```

`a shell advances that one per snapshot it admits`. The fix commit's own contract text, added in the same commit, exists to forbid exactly that reading:

> The wire generation a shell decodes is `SnapshotMeta::generation`. This rule governs that one, and the gate advances it once per snapshot the gate admits. **A shell hosts the gate; it does not supply the counter**, and a source never does — `backend-contract.md:335-339`

Hosting and supplying is the distinction that sentence was written to draw, and `types.rs` collapses it. The provenance makes it worse rather than better. `git show 63dbc77 -- docs/instruments/backend-contract.md` shows the pre-fix line read *"so a shell advances it once per snapshot it admits"* — the shell was one of the three actors under review. That wording was deliberately mirrored into `types.rs` by `0eafacd`, whose own message says so: *"the contract says instead what a shell must do: advance the wire generation once per snapshot it admits, rather than forwarding the feeder's."* `63dbc77` then corrected the contract to say the gate and left the mirror behind. The sweep stopped at the file the finding was raised against.

This is not a wording preference. `IngressSnapshot::generation` is the counter a reader is most likely to confuse with the wire one — that is why this comment exists at all — so the sentence disambiguating them is the last place that can afford to name the wrong actor. One sentence closes it: *"the gate advances that one per snapshot it admits, and a shell must not forward this counter into it."*

Everything else in that comment is right, including the reason `IngressSnapshot::generation` is excluded from the rule.

### 2. The canonical example on its own exclusion list — **closed**

`backend-contract.md:279-280` now names the unit of output as "an input validation and time/integrity gate admits a snapshot", and `341-344` states plainly that `IngressSnapshot::generation` and `SnapshotMeta::generation` are two counters, not one, with the reason (the feeder's advances on an authorization change, which the rule excludes) and the obligation (a shell must not forward one into the other). The governed counter is no longer a member of the list of counters the rule does not govern.

### 3. "produces" vs "admits" — **closed**

Admission, consistently: `backend-contract.md:279` ("admits a snapshot"), `aircraft.rs:226-230` ("one step per snapshot the gate admits… A gate that admits nothing new still advances it; a gate that stopped admitting does not"), `CHANGELOG` ("counts snapshots admitted, not source groups advanced"). The stalled-source case now has one answer, and it is the one a liveness check can act on. `pssa.md`'s "produces" is not in conflict — the gate's unit of output *is* an admitted snapshot, which is the framing the section opens with.

### 4. Nothing recorded the wire-visible meaning change — **closed**

The `0.5.1` entry states what a consumer holding the old meaning reads, in both directions that matter: the liveness consumer (was reporting quiet data, now reports a stalled producer) and the content-change consumer (was always wrong, now wrong more often, compare the content). It also states why no gate can catch this — the field, width and offset are unchanged, so the ABI number does not move and `check-release-markers` compares numbers. That is the honest reason the note is owed.

`check-release-markers.sh` passes and I checked the five values against the tree myself: ABI 7 (`abi/v7.rs`), scene 1, corpus 5, digest `5cded149…`, panels `pfd, hsi, monitor`. `renderer-verification.md:319` now reads `LIVENESS on a stalled render generation`, which names the counter.

### The `2^31` boundary — **closed, and non-vacuous in both directions**

Mutated `serial_is_newer` from `distance < SERIAL_HALF_RANGE` to `<=`:

```
test stamp::tests::serial_ordering_is_wrap_safe ... FAILED
test stamp::tests::neither_value_is_newer_at_exactly_half_the_range ... FAILED
    forward across half the range must not be newer: 2147483648 vs 0
test stamp::tests::one_step_inside_half_the_range_still_answers ... ok
```

Both redden. And the inside case is not decoration — mutating to `distance < SERIAL_HALF_RANGE - 1` reddens `one_step_inside_half_the_range_still_answers` **alone**, so the pair pins a boundary rather than a predicate that always says no. That is exactly the property the docstring claims.

One honest note on "both directions": at *exactly* half the range `serial_distance` is symmetric — `a.wrapping_sub(b)` and `b.wrapping_sub(a)` are both `2^31` — so the reverse assertion in `neither_value_is_newer_at_exactly_half_the_range` evaluates the same computed distance as the forward one and cannot fail independently of it under a distance-based implementation. It still earns its place (it would catch an implementation that compared its two arguments asymmetrically, a signed comparison being the obvious one), but the genuinely independent direction is the one in `one_step_inside_half_the_range_still_answers`, where the two distances differ (`2^31-1` against `2^31+1`). No change wanted; just don't read the boundary test as two independent probes.

### Gates

`fmt`, `clippy --locked --all-targets --workspace -D warnings`, `test --locked --workspace --all-targets --no-fail-fast`, `doc` with `-D missing_docs -D rustdoc::broken_intra_doc_links`, `build --locked --release --workspace`, every `scripts/check-*.sh` and `scripts/test-*.sh`, and both evidence-gate invocations: all green. `git status --porcelain` clean after every mutation.

### One coordination note

PR #63 (`fix/46-display-reason-registry`) also adds a `## [0.5.1]` entry against `main`, with a different body. Whichever lands second needs renumbering — `check-release-markers.sh` reads only the newest entry, so it will not notice two entries claiming one version.

PR #63, opened 2026-08-18
One display-reason registry for every shell (#46)

### LINKED ISSUE #46: Define one append-only DisplayReason registry for Rust, Swift, and JavaScript
## Decision to make

Three shells — Rust, Swift, and JavaScript — each report why a display is not
showing current, valid panel imagery. If each shell enumerates its own reason
set, the sets drift: a reason one shell can name becomes unrepresentable in
another, and a cross-shell comparison of display health compares vocabularies
instead of states.

Define **one append-only `DisplayReason` registry** that all three languages
draw from:

- One authoritative enumeration lives in this repository, alongside the other
  cross-shell contract values (the Rust side already has `DisplayFault` in
  `indicate-instrument-symbology` as the anchor).
- Append-only, like the opcode vocabulary: new reasons are added, never
  renumbered or removed, so an older shell degrades gracefully on a reason it
  does not know.
- Swift and JavaScript consume the same registry — generated or mirrored with
  a drift check — rather than re-declaring it.

## Open questions

- Where the registry lives (a crate constant set vs. a generated table) and
  how the Swift and JavaScript mirrors are checked against it.
- What an unknown reason maps to on an older shell.

Tracked under #36, the cross-shell tracker; not a prerequisite for closing
#35 or #36 unless the decision moves a pinned digest.

luofang34:
Closes #46

## Decision taken

- **Where the registry lives.** The authoritative `DisplayReason` registry is the
  `DisplayFault` enum in `indicate-alerts`
  (`crates/indicate-alerts/src/condition.rs`). One correction to the issue text:
  the enum lives in `indicate-alerts`, not `indicate-instrument-symbology` —
  symbology only renders annunciation labels for it. Each reason has a stable
  one-byte code, packed into the cross-shell `AlertId` as `0x0500 | code`
  (family `0x05` is the display family), so the wire identity already exists
  and is now pinned by test.
- **Append-only rule.** Stated in the new contract doc
  `docs/instruments/display-reason-registry.md`, mirroring how the scene
  opcode vocabulary states its rule (ADR-0017): new reasons take the next free
  code; codes are never renumbered, reused, or removed.
- **Unknown-reason mapping.** Fail-closed: an older shell that receives a code
  it does not know maps it to a generic display-not-current presentation at
  warning level, never to healthy — consistent with the Unknown-sentinel
  convention (VAL-01). On the Rust side, `DisplayFault::from_code` and
  `class_of` return `None` for unknown codes, and the doc states that `None`
  is not a healthy state.
- **Swift/JS mirror obligation.** The mirrors live in their downstream
  repositories. The doc states the obligation: each mirror declares the same
  reasons with the same codes, declares nothing the registry does not define,
  and each downstream repo carries a drift check against the registry table,
  under the same pin discipline as the scene-conformance corpus. No generator
  is built here, per the issue's scope guidance.

## What changed

- `crates/indicate-alerts/src/condition.rs` — `DisplayFault` is documented as
  the append-only cross-shell registry; added `DisplayFault::ALL` (the
  registry in ascending code order); made `code()` and `from_code()` public
  with fail-closed decoding documented. No behavior change.
- `crates/indicate-alerts/src/condition/tests.rs` — pinning tests (exact
  variant/code/`AlertId` mapping, fail-closed decode of codes 0 and 6–255) and
  registry doc tests that parse the contract table and fail the build when the
  table and the code disagree, after the `layer_profile_doc_tests` pattern.
- `docs/instruments/display-reason-registry.md` — new contract doc: the
  registry table, the append-only rule, the unknown-reason mapping, and the
  Swift/JS mirror + drift-check obligation (ASD-STE100).

## Contract values

None of the five contract values moved (state ABI, scene format, corpus,
composition digest, panel set); `instrument-bench` reports the composition
digest matches its pin. `CHANGELOG.md` and `release-manifest.json` are
untouched. Note: this lands a public API addition (`DisplayFault::ALL`,
`code()`, `from_code()`), which per CONTRIBUTING.md makes the next merge a
release-tag candidate.

## Gates run

- `cargo fmt --all --check`
- `cargo clippy --locked --all-targets -- -D warnings`
- `cargo test --locked --all-targets` (all suites pass, incl. the 4 new tests)
- `RUSTDOCFLAGS="-D missing_docs -D rustdoc::broken_intra_doc_links" cargo doc --locked --no-deps`
- `cargo build --locked --release`
- `scripts/check-structure.sh`, `check-release-manifest.sh`,
  `check-release-markers.sh`, `check-instrument-requirements.sh`,
  `check-certification-claims.sh`, `check-standards-registry.sh`,
  `trace-report.sh`, `detect-target.sh`
- thumbv7em-none-eabihf closure check (all 12 closure crates)
- `cargo run --locked -q -p instrument-bench` (digest matches pin; 242
  admission cases pass)
- both `evidence-gate` invocations: `VALID`


luofang34:
**REQUEST CHANGES** — reviewed at `736a625cf9b39be9e85cff603856df592cee79d2`, in a clean worktree off `origin/main`.

The registry itself is the best-built thing I have reviewed in this repo today. `check_display_reason_completeness` refuses on every malformed input I could construct, each with its own message, and the Rust suite pins codes, identities, classes, and the document table. Two things block, both small: the selftest that guards the new check is fail-open under one extra condition, and a public API addition lands with no changelog entry against `CONTRIBUTING.md`'s second release trigger.

## 1. Mutation testing

Baseline: `cargo test --locked --workspace --all-targets --no-fail-fast` → exit 0, **718 passed / 0 failed**. `check-structure.sh` → 0. `test-structure.sh` → 0, **`structure-selftest: OK (14 cases)`** with both merged-in cases present. Each mutation applied, the mandated command (or the named gate) run, restored with `git restore`; `git status --porcelain` empty after each.

| # | Mutation | Result | Caught by |
|---|---|---|---|
| P1 | `check_display_reason_completeness()` → `return 0` | **RED** | `test-structure.sh` case "a reason outside `DisplayFault::ALL` refused" (`FAILED (1 of 14)`) |
| P2 | **the real attack**: add `ProbeReason` with `slot() => 2` (colliding with `CommandBufferCorrupt`) + a `class()` arm, left out of `ALL` — a version that *compiles* | `cargo test -p indicate-alerts` → **exit 0, GREEN**; `check-structure.sh` → **exit 1** | only the shell check: `FORBIDDEN: DisplayFault has 6 slots but ALL holds 5…` |
| P3 | two variants share a slot, both in `ALL` (`BackendLost => 2`) | **RED** (5) | `every_reason_holds_its_own_slot_and_code`, `display_reason_codes_are_pinned`, `identities_are_unique`, `class_of_agrees_with_condition_class`, `the_documented_registry_matches_the_code` |
| P4 | malformed input: wildcard arm `_ => 2,` hiding a 6th variant | **REFUSED** | `FORBIDDEN: the DisplayFault \`slot\` match has 6 arms but only 5 name a variant…(fail-closed)` |
| P5 | malformed input: `slot` renamed to `slot_of` — the check cannot read it | **REFUSED** | `FORBIDDEN: condition.rs has no readable \`slot\` match to count…(fail-closed)` |
| P6 | malformed input: a second `pub const ALL: [Self; 3]` | **REFUSED** | `FORBIDDEN: condition.rs declares 2 \`ALL\` lengths…(fail-closed)` |
| P7 | doc table: `BackendLost` class `Caution` → `Warning` | **RED** (1) | `the_documented_registry_matches_the_code` |
| P8 | doc table: last row deleted | **RED** (2) | + `the_registry_section_holds_exactly_one_table` |
| P9 | `from_code` admits code 6 | **RED** (1) | `display_reason_decoding_is_fail_closed` |
| P10 | `ALL` reordered (first two swapped) | **RED** (3) | pinning + slot + doc tests |
| P11 | `ALL` removed entirely | **exit 1, but silently** — see §7.1 | — |
| P12 | `FAMILY_DISPLAY` `0x05` → `0x07` | **RED** (5) | identity/class/doc/manager tests |
| P13 | a second table smuggled into the registry section | **RED** (2) | `the_registry_section_holds_exactly_one_table`, `the_documented_registry_matches_the_code` |
| **P14** | `check_display_reason_completeness()` → `return 0`, **plus** any unrelated finding present | **GREEN — selftest reports `ok`** | *nothing* — see §2 |

P2 is the important row and it vindicates the design: the Rust suite passes with a sixth reason silently sharing `CommandBufferCorrupt`'s code, identity, label and class, and only the counting check in `check-structure.sh` sees it. That is precisely the hole the PR says it is closing, and it closes it. P4–P6 and P11 confirm the check refuses rather than no-ops on every unreadable shape I could build.

## 2. Blocking — the new selftest case is fail-open when a second finding is present

`scripts/test-structure.sh:148-171` asserts only that `check-structure.sh` *exited nonzero*, not that it refused for this reason. So any other finding in the tree satisfies the case while the check it names does nothing.

Reproduction (both steps, then restore):

```sh
# 1. gut the check the case claims to cover
#    scripts/check-structure.sh:197  ->  check_display_reason_completeness() { return 0; ... }
# 2. give the gate an unrelated finding under the probe: pad
#    crates/indicate-alerts/src/condition.rs from 401 to 498 lines, so the
#    probe's +3 lines crosses the 500-line limit at check-structure.sh:65
bash scripts/check-structure.sh   # exit 0 — clean before the probe
bash scripts/test-structure.sh
#   ok: a reason outside DisplayFault::ALL refused
#   structure-selftest: OK (14 cases)
#   exit 0
```

The case reports `ok` for a check that returns immediately. Gutting the check on its own does red the case (P1), so this is not unconditional — but a guardrail that holds only while the rest of the tree is spotless is not the guardrail `CONTRIBUTING.md` asks this PR to land, and the co-fault is cheap: the probe edits `condition.rs`, and the registry is append-only, so this file grows with every reason added.

The fix is one line, and this same file already uses the right pattern at `test-structure.sh:110` (`expect_acceptance` greps for `"may not reach"`):

```sh
if INDICATE_STRUCTURE_SELFTEST_CHILD=1 bash scripts/check-structure.sh 2>&1 \
    | grep -q 'a variant outside ALL takes a code no test can see'; then
    echo "ok: a reason outside DisplayFault::ALL refused"
    passed=$((passed + 1))
else
    echo "REGRESSION: a reason outside ALL was accepted; it can take a code no test sees" >&2
    failed=$((failed + 1))
fi
```

*Non-blocking, pre-existing*: the seven `expect_refusal` cases and the three `expect_term_refusal` cases have the same shape and the same exposure. Not this PR's to fix, but worth an issue — the file already knows the better pattern.

## 3. Blocking — a public API addition with no changelog entry

`DisplayFault::ALL`, `DisplayFault::code()` and `DisplayFault::from_code()` become public in `crates/indicate-alerts/src/condition.rs:214-262`. `CONTRIBUTING.md` states the trigger without qualification:

> Cut one whenever any of the five contract values moves … **or when a public API addition lands that consumers are expected to call.**

and gives its reason in the next paragraph — that the first trigger *misses the additions made for consumers*, and that a predicate "published so nobody writes their own copy does not stop the second copy from being written if it cannot be pinned by name."

This PR is that paragraph's example. Its own registry document says so at `docs/instruments/display-reason-registry.md:79-84`: *"No mechanism in this repository makes a stale mirror fail. An append moves none of the five contract values, so a pinned consumer reports no error across it. Each downstream repository must therefore carry its own drift check against this table."* A downstream drift check has to pin this repository at something. Without an entry and a tag, that something is a bare revision with a comment explaining which commit it is — the archaeology the trigger exists to remove.

The PR body says this "makes the next merge a release-tag candidate." `CONTRIBUTING.md` puts the changelog entry in the PR (step 2) and the tag after the merge (steps 3–4); deferring the entry to a later change leaves the addition unrecorded in between. The sibling PR #59 cut `0.5.1` for this same trigger today.

No gate can catch this: `check-release-markers.sh` compares only the *newest* entry against the tree, and passes here reporting `OK (0.5.0; ABI 7, scene 1, corpus 5, panels: pfd, hsi, monitor)`. That is why it is a review finding.

## 4. The merge from `main` — union verified, nothing dropped

`736a625` has parents `a2c85b6` (PR side) and `830e9a6` (main). The resolved `scripts/test-structure.sh` is a strict superset of both:

```sh
git diff a2c85b6 HEAD -- scripts/test-structure.sh | grep -c '^-[^-]'   # 0
git diff 830e9a6 HEAD -- scripts/test-structure.sh | grep -c '^-[^-]'   # 0
```

Zero deletions against either parent: main's `.worktrees` prune case and this PR's display-reason case are both present and intact. `bash scripts/test-structure.sh` confirms it at runtime — `structure-selftest: OK (14 cases)`, with both `ok: a reason outside DisplayFault::ALL refused` and `ok: content under .worktrees is not walked` in the output.

I checked `scripts/check-structure.sh` the same way, since both sides touched it too: 0 deletions vs `830e9a6` (this PR's 65-line check is whole), and the 3 changed lines vs `a2c85b6` are exactly main's three `find` prunes being adopted. And `git diff a2c85b6 HEAD -- crates/indicate-alerts docs/instruments/display-reason-registry.md docs/instruments/README.md` is empty — the PR side came through the merge untouched.

## 5. Pins

Nothing this PR authors moves a pin, and the merge did not disturb the ones it inherited: `git diff 830e9a6 HEAD --stat` lists only the six PR files, so `CHANGELOG.md`, `release-manifest.json`, the evidence artifacts and `evidence-graph.evg` are byte-identical to `main`. Recomputed anyway rather than assumed: `instrument-bench` reports `scene digest: 5cded149… (matches pin)` and `screen-composition digest: f87c3b4b… (matches pin)`; `check-release-manifest.sh` regenerates and diffs clean; `check-recorded-counts.sh` re-runs all three recorded commands and matches; both evidence-gate invocations are `VALID` (30 nodes / 73 edges / 0 exceptions).

## 6. Gates — exactly what I saw on this branch

`cargo fmt --all --check` 0 · `cargo clippy --locked --all-targets --workspace -- -D warnings` 0 · `cargo test --locked --workspace --all-targets --no-fail-fast` 0 (718/0) · `RUSTDOCFLAGS="-D missing_docs -D rustdoc::broken_intra_doc_links" cargo doc --locked --workspace --no-deps` 0 · `cargo build --locked --release --workspace` 0 · `cargo check … --target thumbv7em-none-eabihf` 0 · `cargo run -p instrument-bench` 0.

Scripts, all exit 0: `check-certification-claims` · `check-instrument-requirements` · `check-recorded-counts` (+ `--selftest`) · `check-release-manifest` · `check-release-markers` · `check-standards-registry` · `check-structure` · `test-instrument-requirements` · `test-release-manifest` · `test-standards-registry` · `test-structure` (14 cases) · `trace-report` · `detect-target`.

Comment and doc rules: clean. No temporal phrasing, PR/issue numbers, line counts, or fragments in the added comments; I swept the whole merge diff for them.

## 7. Non-blocking

1. **The zero-`ALL` message can never print.** `scripts/check-structure.sh:19` sets `set -euo pipefail`, and `declarations=$(grep -c 'pub const ALL: \[Self; [0-9]*\]' "$file")` at `:205` is a bare assignment, so when `ALL` is absent `grep` exits 1 and `set -e` kills the script *before* the `if`. P11 confirms it: exit 1 with **zero bytes of output**, and `check_crate_naming`, `check_tier_law` and `check_crate_map` never run. The verdict is still fail-closed, so this is diagnostics, not correctness — but the branch is carefully worded for a case it cannot reach. `declarations=$(grep -c … || true)` restores it.
2. **The retirement rule does not say which cell.** `display-reason-registry.md:51-53` says to "keep its row and its code, and write `retired` in its row", but `the_documented_registry_matches_the_code` fixes cells 0–3 to code / variant / identity / class and checks only that cell 4 is non-empty. So `retired` can only go in **Meaning**, and the variant must stay in `ALL` or `the_registry_section_holds_exactly_one_table` fails on the row count. Say that in the doc — a future retirement will otherwise discover it by reddening the build.
3. **The PR body is stale after the merge.** It reports "242 admission cases pass"; the merged tree reports `admission: 280 cases pass, 196 counted warnings`. Also worth updating the gate list, which predates the merge.
4. `docs/instruments/display-reason-registry.md:74-77` states the mirror obligation clearly and `:79-84` is admirably honest that nothing here enforces it. Consider linking the downstream drift-check requirement from `docs/instruments/README.md:53` too, so the obligation is discoverable from the index and not only from inside the contract.

luofang34:
**Verdict: the finding is closed.** The P14 probe now reddens the case it names, it reddens for the right reason, and the `0.5.1` entry agrees with the tree. One non-blocking observation below.

Reviewed at `c1b2a37` (detached `origin/fix/46-display-reason-registry`), diffed against `origin/main`.

### P14 probe re-run: gut + pad

Gutted `check_display_reason_completeness` to a no-op body and padded `crates/indicate-alerts/src/condition.rs` to 521 lines, exactly the scenario that previously printed `ok`:

```
REGRESSION: a reason outside ALL was accepted; it can take a code no test sees
REGRESSION: the gate walked into a linked worktree and reported on a checkout
structure-selftest: FAILED (2 of 14 cases)
```

Red. The second `REGRESSION` is expected collateral — the worktree-prune case asserts the gate *passes*, and the padded file fails it — not a separate finding.

### Red for the right reason

Gutting the completeness check **without** padding — the case where the gate exits zero and nothing else refuses, which is precisely what the old exit-status match slept through:

```
REGRESSION: a reason outside ALL was accepted; it can take a code no test sees
ok: content under .worktrees is not walked
structure-selftest: FAILED (1 of 14 cases)
```

One red, and it is the right one. I also confirmed the string being matched is the completeness check's own, not a message some other check could supply: with the gate intact and the probe variant planted, the only line it emits is

```
FORBIDDEN: DisplayFault has 6 slots but ALL holds 5; a variant outside ALL takes a code no test can see, …
```

which is the sole occurrence of `takes a code no test can see` in the script, and it lives in the `declared -ne variants` branch the probe targets. The three fail-closed branches above it carry distinct messages, so the case cannot be satisfied by a refusal it did not ask for.

Capturing rather than piping is also right, and load-bearing: under `pipefail` a pipeline would have carried the gate's own non-zero status regardless of the message.

### Unmodified tree

`structure-selftest: OK (14 cases)` / `check-structure: OK`, both from the standalone run and from the copy `check-structure.sh` invokes internally.

### `0.5.1` release entry

`check-release-markers.sh` passes and prints what it computed:

```
check-release-markers: OK (0.5.1; ABI 7, scene 1, corpus 5, panels: pfd, hsi, monitor)
```

I read the four remaining sources directly rather than take the script's word: `abi/v7.rs` `VERSION = 7`, `SCENE_FORMAT_VERSION = 1`, `"corpusVersion": 5`, `BUILTIN_SCENE_DIGEST = 5cded14978b2e5ba3a17b61959ed0b35061334adf3fde4242f47e214f0f07aef`. All five table cells agree, and `Panel set changed since the previous release: no` matches the `0.5.0` row. The entry is owed under the second release trigger — `ALL`, `code()`, `from_code()` are a public addition consumers are expected to call — and it is here.

### Gates

`fmt`, `clippy --locked --all-targets --workspace -D warnings`, `test --locked --workspace --all-targets --no-fail-fast`, `doc` with `-D missing_docs -D rustdoc::broken_intra_doc_links`, `build --locked --release --workspace`, and every `scripts/check-*.sh` and `scripts/test-*.sh`: all green. `git status --porcelain` clean after every probe.

### One nit, not blocking

The case now matches the message and discards the status (`… 2>&1 || true`). That closes the reviewed finding — the old form accepted any refusal, the new one accepts only this one — but it is the mirror image of the same shape, and it is reachable: dropping `status=1` from the completeness check while keeping its `echo` leaves the gate printing `FORBIDDEN: DisplayFault has 6 slots but ALL holds 5 …` and then exiting **zero**, and the selftest still reports `structure-selftest: OK (14 cases)`. CI goes green on a registry violation the gate just described out loud.

Asserting both halves closes it. I ran this and it behaves:

```sh
probe_status=0
probe_output="$(INDICATE_STRUCTURE_SELFTEST_CHILD=1 bash scripts/check-structure.sh 2>&1)" || probe_status=$?
if [ "$probe_status" -ne 0 ] && printf '%s' "$probe_output" | grep -q "takes a code no test can see"; then
```

Red against the `status=1`-dropped mutation, green on the unmodified tree. Worth taking, but it is a strengthening of a case that already does what review asked of it — not a reason to hold the PR.

Issue #57, opened 2026-08-18
No flap or trim indication exists; the airframe-config vocabulary is only a reserved id


luofang34:
Nothing displays flap or trim position. The only occurrence on any surface is an advisory monitor-text fixture string — "GEAR DOWN-LOCKED FLAPS 25 TRIM 4" (`sets/indicate-instrument-panels/src/descriptors/extreme_states.rs:138`) — free text on the advisory channel, which the monitor contract is explicit is not a flight instrument.

## The registry decision comes first

The group registry reserves 0x0E for "engine (planned)" (`crates/indicate-instrument-state/src/group_id.rs:17-26`). Flap and trim are airframe configuration, not engine. Decide whether 0x0E's charter covers actuator/config state or whether configuration takes its own id from 0x12+ — the registry is append-only and doc-first, so this is a doc edit before it is a variant.

## State

- Flap: position ratio, plus optionally the selected detent. Sensed and selected must not be conflated — the baro precedent: a selection and an applied value are compared and flagged, never merged (`aircraft.rs:118-121`).
- Trim: per-axis ratio, elevator first. The takeoff band is airframe profile/config data, not wire data.
- Every field optional: a vehicle without the sensor renders `Missing`, not a substitute (ADR-0017, `aircraft.rs:27-28`).

## Surface

These are conventional-instrument surfaces — the "Conventional instruments" row of the intended-function matrix (comparison/reversion, `docs/instruments/intended-functions.md`) — not PFD elements; the PFD's 480x360 is fully allocated. The shape the panel contract anticipates is a small config panel (or one per instrument) as a registry entry in a new set: a set outside `BUILTIN_PANELS` may leave `raster_baselines` empty until rasterizer coverage exists, so the panel can land before its pixel pins.

## Config interplay

A flap indicator that marks its speed-limit band reads the same airframe values that feed the speed tape's `v_speeds` config. Reuse that config schema pattern; do not duplicate the values in a second key.

## Assurance surface

This adds a new intended function, and that documentation is part of the implementation, not a follow-up: rows in intended-functions.md and requirements.md, and an FHA/PSSA entry — a misleading configuration indication is its own hazard class, distinct from misleading primary flight data.

## Guardrails, same PRs

[...]

Issue #55, opened 2026-08-18
The CDI's nav source is identified by hue alone


luofang34:
Which source drives the CDI is encoded as hue alone — magenta vs green (`sets/indicate-instrument-panels/src/hsi/cdi.rs:12-17`) — plus the single letter V/G on the vertical-deviation scale (`hsi/boxes.rs:165-171`). Receiver identity appears nowhere: Nav1 and Nav2 wear the same green, and the only textual source label on the HSI is heading's (`hsi.rs:137-144`). A hue-only distinction also fails under color-vision deficiency; the theme validator defends the magenta/green separation distance (`theme.rs:41-43`), but a text annunciation is the actual fix, not a stronger hue.

## First step needs no ABI change

`NavSource` already distinguishes Gps / Nav1 / Nav2 (`crates/indicate-instrument-state/src/aircraft.rs:38-53`). Render "GPS" / "NAV1" / "NAV2" near the rose, in the source color, gated exactly as the CDI is (source selected, nav shows a value, course converts into the rose reference — `hsi.rs:82-88`) so the label and the needle appear and disappear together. The labels are uppercase plus digits — inside `PANEL_VOCABULARY` as it stands.

## Follow-on, ABI

Distinguishing what a receiver is tuned to (VOR vs LOC) needs the facility type in state — a Nav layout addition. Keep it out of the first PR: receiver identity is free today, facility type is a wire change that belongs in the coordinated ABI batch.

## Guardrails, same PRs

A switch-together test extending the source_tests discipline — the label text, the CDI color, and the CRS-box color must change in the same frame; corpus entries for each source; a group region covering the label so its text carries a claim; baseline re-pin with its reason.

Issue #54, opened 2026-08-18
The CDI scale is fixed and unannunciated


luofang34:
The CDI's lateral scale is a constant: ±2 dots = ±50 m (`LATERAL_M_PER_DOT`, `crates/indicate-instrument-feeder/src/nav_display.rs:17`), vertical ±2.5 dots = ±20 m (`nav_display.rs:20`), chosen as a terminal-area scale. Nothing on the glass says what full scale means, and nothing in the state lets a guidance source vary it. When guidance later flies phases that want different scales (enroute vs approach), the same two dots will silently mean different distances — the "same picture, different meaning" failure the reference annunciations elsewhere on these panels exist to prevent.

## State

The scale mode must be declared by the guidance source, not inferred: a mode enum on the Nav group (a layout move) — Enroute / Terminal / Approach plus `Unknown` — with `Unknown` failing the group, never a silently wrong scale. Deriving a label from a meters-per-dot number on the display side would invent meaning the source did not declare; the enum is the declaration.

## Feeder

`nav_display.rs` maps mode → meters-per-dot through a table instead of two constants. Sign conventions unchanged. An unknown mode removes the display the way an unusable solution quality does today.

## Drawing

Mode text near the CDI in the source color (`source_color`, `sets/indicate-instrument-panels/src/hsi/cdi.rs:12-17`), gated exactly as the CDI is (`hsi.rs:82-88`) so the label and the needle appear and disappear together.

**The glyph constraint shapes the vocabulary.** `PANEL_VOCABULARY` has no `+` and no `/` (panel-contract.md, "Symbology you may not invent"), so richer per-approach labels of the LNAV+V / L/VNAV family are unrepresentable and would fail admission as `GlyphCoverage` until the glyph pack grows. Start with ENR / TERM / APR — uppercase, in vocabulary — and treat extended labels as a follow-on that lands together with its glyph-pack change.

## Guardrails, same PRs

Mode-to-scale table tests; fail-closed unknown-mode decode test; a switch-together test (the scale in use and the annunciated mode must change in the same frame — the source_tests discipline); corpus entries per mode; ABI round-trip; baseline re-pin with its reason.

luofang34:
Landed on `main` in #64 (merge commit 46b6e44), which carried the whole state-ABI v8 allocation batch. Closed here rather than automatically: the pull request body listed the issues after a single `Closes` keyword, which only closes the first.

Issue #53, opened 2026-08-18
The HSI rose carries no bearing pointers


luofang34:
The rose rim carries the heading bug (`sets/indicate-instrument-panels/src/hsi/rose.rs:111-134`) and the ground-track diamond (`rose.rs:137-157`), and the CDI serves exactly one selected source (`NavData.source`, `crates/indicate-instrument-state/src/aircraft.rs:70-93`). There is no way to display a bearing to a second facility: no pointer needles, no pointer-source legends.

## State

Bearing pointers are bearings from up to two receivers concurrently, independent of which source drives the CDI. The single-source `NavData` cannot express that, so this is a new group (append-only registry, `group_id.rs:10-26`, ids from 0x12): per pointer, a source id (None / Nav1 / Nav2 / Gps plus `Unknown` failing closed), the bearing in radians with its `HeadingReference`, and per-pointer validity. A bearing converts into the rose reference through the same gate as the heading bug — an unconvertible reference suppresses the pointer and flags it, never draws it raw (the pattern at `hsi/boxes.rs:111-114`).

## Drawing

Pointer 1 as the single-line needle with open head, pointer 2 as the double-line form, at a radius inside the rose (~120 px) so both read under the CDI. Both rotate with the card, including under a track-up basis (`hsi.rs:66-74`). Legend boxes at the lower corners identify each pointer's source.

Color is a real decision: the selection cyan is chartered as "bugs, selected values, baro" (`crates/indicate-instrument-symbology/src/palette.rs:45-46`); a pointer is pilot-selected but displays a sensed bearing. Either extend that charter deliberately or add a dedicated token — `Theme::all()`'s exhaustive destructuring forces a new token into validation either way.

## Geometry constraint

The lower corners are taken: the CRS box at (2, 322, 112, 36) and the heading-select box at (366, 322, 112, 36) (`hsi/boxes.rs:74-141`), with the vertical-deviation scale occupying x ≈ 444-460 (`boxes.rs:144-173`). Legends must share, stack, or claim that space conditionally — state the layout rule in the panel rather than discovering it in overlapping pixels.

## Descriptor

The HSI's `required_groups` lists seven groups today (`sets/indicate-instrument-panels/src/descriptors.rs:204-212`). Decide whether the pointer group joins as required-and-usually-Missing, or whether the descriptor vocabulary needs an optional-consumption notion. Either way the group regions must cover the legends so their text carries claims.

## Failure semantics

[...]

PR #64, opened 2026-08-18
Bump the state ABI to v8 as the batch coordination point (#58)

### LINKED ISSUE #50: No surface annunciates AP/FD modes, and no AP target values exist in state
The PFD annunciates the flight director's own mode — FD ATT / FD NAV / FD APR, magenta, at (240, 44) (`sets/indicate-instrument-panels/src/pfd.rs:241-256`) — and nothing else. There is no AP/FD engagement annunciation, no active or armed lateral/vertical mode display, and the state carries no autopilot target values at all: `Selections` (`crates/indicate-instrument-state/src/aircraft.rs:97-122`) holds the heading bug, the selected altitude with its reference identity, and the selected baro — no selected airspeed, no selected vertical speed, no altitude-capture target. A pilot flying against automation cannot see what it is holding or armed to capture.

## What implementing it needs

**State.** A new group (the registry is append-only, `crates/indicate-instrument-state/src/group_id.rs:10-26`, next standard ids from 0x12): engagement (Off / FD / AP plus `Unknown` failing closed, the `NavSource::Unknown` pattern), active lateral mode, armed lateral, active vertical, armed vertical — each an enum whose unknown wire value fails the group — and the targets: airspeed m/s, vertical speed m/s, altitude. The altitude target needs the same reference-identity trio as `altitude_sel` — class, origin, geoid model (`aircraft.rs:104-117`) — or a target expressed against a different reference renders as comparable when it is not.

**Stamped vs declared is the first decision.** `Selections` lives in the ageless declared codec ("pilot/UI state ... carry no age", `crates/indicate-instrument-state/src/abi/v7/declared.rs:1-6`). AP engagement and modes from a real flight computer are telemetry: a stale "AP engaged" annunciation is exactly the failure the freshness discipline exists to prevent, and `fc_state.rs` is the in-repo precedent — a stamped lane with pinned source identity and a staleness threshold whose heartbeat loss surfaces as stale. Modes and engagement likely belong on a stamped lane; the target values could defensibly stay declared UI state. This decides which codec and which resolve path owns each field, so it comes before the layout.

**Feeder.** No shipped posture publishes flight-director data today (`crates/indicate-instrument-state/src/director.rs:17-19`), and the same will initially be true here. Landing the contract before the data is the established pattern; the tracking issue keeps the unfed state visible rather than letting it read as done.

**Surface.** Two options. (a) A new strip panel — the monitor panel exists to prove a panel is a registry entry, not a shell change (`sets/indicate-instrument-panels/src/monitor.rs:1-3`); a strip declares its own canonical frame and aspect, and joining `BUILTIN_PANELS` (`descriptors.rs:345-346`) moves the composition digest once, for a stated reason. (b) A band inside the PFD, which re-pins the PFD baselines and competes with the FD annunciation already at (240, 44). Option (a) keeps composition a shell choice.

**Symbology.** Active/armed mode text needs color tokens. Reusing `radio_nav` green for "active mode" would overload source identity — the validator keeps `gps`/`radio_nav` at least 64 apart precisely so source class survives theming (`crates/indicate-instrument-symbology/src/theme.rs:30-33`, `:41-43`). Likely shape: a dedicated active-mode token plus `safety::ANNUNCIATION_WHITE` for armed; `Theme::all()`'s exhaustive destructuring forces a new token into validation. Labels must fit the glyph vocabulary — uppercase is covered (AP, FD, HDG, ROL, ALT, VS, PIT), the airspeed unit is `kt`, and a vertical-speed unit must be `FPM`, because the lowercase set is only `k` and `t` (`PANEL_VOCABULARY`, panel-contract.md "Symbology you may not invent").

**Honest status.** Missing renders blank, never last-known modes; Stale wears the amber flag; Failed the red X; every target numeral carries its claim through a declared group region.

## Guardrails, same PRs

Fail-closed decode tests for every enum; armed-vs-active and mode-switch draw tests; corpus entries for shown / stale / failed; a pinned baseline (or an honestly empty one until rasterizer coverage exists, as the contract permits for a new set); the digest re-pin with its reason in the commit; and the docs row — intended-functions.md withholds command-guidance credit without a separate intended function, so this lands with its own function row and AIR-IN entry rather than borrowing the flight director's.

### LINKED ISSUE #58: [TRACKING] Panel feature gaps: AP modes and targets, trend, TAS, bearing pointers, CDI annunciations, altitude drum, flap/trim
A feature audit of the shipped panel set — `pfd`, `hsi`, `monitor` (`BUILTIN_PANELS`, `sets/indicate-instrument-panels/src/descriptors.rs:345-346`) — against the instrument functions a pilot-facing display of this class is expected to carry. Each gap has its own issue with the implementation shape worked out; this issue owns the ordering and the one decision they share.

## The gaps

- [ ] #50 AP/FD mode annunciation and AP target values (selected airspeed, vertical speed, altitude capture)
- [ ] #51 Airspeed trend vector on the speed tape
- [ ] #52 True airspeed in `AirData` and on the glass
- [ ] #53 Bearing pointers and their source legends on the HSI
- [ ] #54 CDI scale modes, declared and annunciated
- [ ] #55 CDI nav-source text annunciation
- [ ] #56 Rolling-digit altitude readout
- [ ] #57 Flap and trim indication

## Sequencing

**1. Batch the ABI moves.** Six of the eight need wire changes — #50 (new group), #51 (Dynamics field + trust valid bit), #52 (Air field), #53 (new group), #54 (Nav field), and #57 (new group) — plus #55's follow-on facility-type field. A group addition is designed to be cheap in-repo (the enum is the single registration point; exhaustive matches turn a new variant into compile errors until handled, `group_id.rs:1-9`), but every ABI move costs each consumer at the wasm, FFI, and JS-writer boundaries, and the state ABI version is one of the five pinned release values. One coordinated ABI revision carrying all agreed layouts beats six serial bumps. The stamped-vs-declared decision in #50 gates the batch: it decides which codec two of the new vocabularies live in.

**2. Independent drawing work can land anytime.** #55's first step (receiver identity — `NavSource` is already in state) and #56 (the scene format already has `clip_rect`) need no wire change and no coordination.

**3. Derivations and postures after the vocabulary lands.** #51's feeder derivation follows its Dynamics field. #50 will initially be contract-live-data-missing, the same posture the flight director is in today (`crates/indicate-instrument-state/src/director.rs:17-19`); this issue keeps that unfed state visible so it cannot read as done.

**4. New surfaces last.** #50's strip panel and #57's config panel(s) each move the composition digest when they join a shipped set, and each introduces a new canonical frame shape — which sharpens #32's open question: the frame-choosing policy should be written before a second frame shape exists, not after three shells improvise one.

## Discipline

One issue per PR; each PR carries its guardrail in the same change — fail-closed decode tests, corpus entries, admission coverage, baseline re-pins with the reason in the commit message; digests are re-pinned by the author of the change that moves them.

Out of scope for every issue above: transport, host postures, and shells. The boundary README draws that line; these issues stay on the panel side of it.

luofang34:
Closes #58. Closes #50, #51, #52, #53, #54, #57, #79.

## What this is

The state ABI moves to v7 → v8, and this branch carries the whole allocation batch that issue #58 coordinates. It opened as scaffolding — a version bump with the layouts recorded but not implemented. Every slice has since landed onto it, so the number now names exactly one wire format that is fully implemented, drawn, and gated.

**Merge this with a merge commit, not a squash.** The branch declares an ATT-01 evidence baseline of its own; a squash rewrites that commit and orphans it. `scripts/check-baseline-survives-merge.sh` says so on this page (advisory by design).

## The wire

| id | group | lane | bytes | issue |
|----|-------|------|-------|-------|
| 0x12 | BearingPointers | stamped | 20 | #53 |
| 0x13 | AirframeConfig | stamped | 24 | #57 |
| 0x14 | ApModes | stamped | 12 | #50 |
| 0x15 | ApTargets | declared | 20 | #50 |

| group | append | bytes | issue |
|-------|--------|-------|-------|
| Air (0x03) | `tas_mps` f32, NaN-absent | 12 → 16 | #52 |
| Dynamics (0x0B) | `ias_trend` f32, NaN-absent + Trust valid bit 9 | 16 → 20 | #51 |
| Nav (0x04) | `scale` u8 | 42 → 43 | #54 (`facility_type` follows with its consumer, #55) |

Each append goes after the trailing `age_ms`, per the stamped-lane growth policy. An older decoder accepts the longer payload and counts the tail. A writer that stamps version 8 and keeps writing the old length has the whole frame refused, not that group — the failure looks like total signal loss rather than one short group, which is the point.

`GroupId::index` is an exhaustive match, not `tag - 1`: 0x0E–0x11 stay reserved while 0x12–0x15 are allocated, so the arithmetic no longer holds. A new variant fails to compile until it has a slot.

## What moves on the glass

- The PFD gains a true-airspeed box at the head of the speed tape, and the tape starts 25 units lower to clear it.
- The HSI gains two bearing needles, and two labels beside the rose: which receiver drives the CDI, and what a dot is worth. Both draw under the needle's own gate.
- A new autoflight annunciator joins the built-in set: what the automation holds now, what it is armed to hold next, and the values it flies toward.
- A configuration panel ships in its own set, `config`, which a shell composes when the airframe has the sensors. It does not move `BUILTIN_PANELS`.

An unknown deflection scale fails the whole nav group rather than drawing at a guessed one. An altitude target whose reference identity does not match the displayed datum is withheld rather than drawn.

## The merge of `main`

`main` released 0.5.1 through 0.5.3 while this branch sat at 0.5.0, and both sides changed the same three things. The merge (f46e633) resolves 15 conflicts:

- **The altitude readout takes this branch's geometry and `main`'s drum.** This branch split `tapes.rs` into `airspeed.rs` + `altitude.rs` and moved the tape bottom; `main` kept one file and replaced the readout with the rolling-digit drum. Either side taken whole would have deleted the other's work.
- **Both nav labels survive.** The scale label is this branch's, the receiver label is `main`'s. A test holds them clear of the rose rim and of each other.
- **The canonical corpus keeps both new states**, `ias-without-tas` and `nav2-source`. Six canonical states raise admission to 418 cases for the built-in set, 48 for the config set and 36 for the template set, and the PFD's frame-overflow count to 256.
- **The release entry becomes 0.6.0.** Two entries under one version leave the older one unchecked, which `check-release-markers.sh` now refuses.

Every pinned value was recomputed by running the suite, never hand-resolved: the composition digest, the screen-composition digest, the PFD and HSI raster baselines, the three composed-frame hashes, and `release-manifest.json` (regenerated by `cargo xtask gen-release-manifest`).

The scene conformance corpus moves to version 6. The two entries that replay the built-in PFD re-pin, because the panel above them emits different bytes for the same altitude. No entry is added or removed, and every hand-built scene keeps its bytes.

## What landed on the branch after the merge

- **The receiver label is pinned to the gate the merge gave it.** The two nav labels and the needle share one condition, and the deflection scale is part of it. Only the needle's disappearance was pinned; hoisting the label back above the scale check left the suite green while the panel painted a receiver name beside a needle that is not there. The assertion fails against that hoist.
- **#80 lands #79's configuration-panel rows**: the intended-functions and requirements entries, the scene-layer-protocol row, and the one-crate-two-`PanelSet`s answer. `layer_profile_doc_tests` now derives its panel list from `ALL_SETS` rather than `BUILTIN_PANELS`, so a set outside the built-ins can no longer ship an undocumented layer profile.

## Evidence

ATT-01 is re-anchored to `67c8eda`, the merge that brought the batch onto #80's branch. All three results and CFG-WORKTREE name one baseline; each `source-digest` is the blob of the test source its run executed, and each `output-digest` is the sha256 of the artifact it names.

## Gates

Green locally on 1.95.0 and in CI: `fmt --check`, `clippy --all-targets -D warnings`, `test --all-targets --no-fail-fast` (793 passed, 0 failed), `doc` with `-D missing_docs -D rustdoc::broken_intra_doc_links`, `build --release`, and every `scripts/check-*.sh`.


luofang34:
Independent review: **do not merge to main yet.** Keeping this branch as the batch base instead.

The finding that decides it: the three golden frames are byte-identical to v7 after the version byte. This revision changes no wire layout, so cutting 0.5.0 from it would publish an ABI number that names one format now and a different, incompatible one after the stacked slices land. `min_len` aborts the whole frame, not the offending group, so a consumer that pinned this rev, changed its version byte to 8, and kept emitting a 12-byte Air group would have every frame rejected once #52 raises that floor to 16 — while both ends, and `release-manifest.json`, still say "state ABI 8". The version number would have stopped identifying a format, which is its only job. The 0.4.0/v7 entry is the contrast: it carried its layout change in the same release.

The batching decision in #58 is sound and stands — one coordinated revision beats six serial bumps. Only the publication order changes: this branch stays open as the integration base, #52, #51, #53, #54 and #57 land **onto it**, and v8 is cut once, when it names exactly one format with real layout changes in it. A reserved-but-unused id costs nothing on the wire in the meantime.

Three fixes to make on this branch before that release:

- `index_is_not_wire_tag_arithmetic` passes under the very implementation it is named for (`tag - 1` gives the same answer for every contiguous id), so the mapping is unpinned. It needs a case that fails on the arithmetic — which arrives naturally with the first sparse id.
- Nothing ties the `state-abi-vN` fixture names to the ABI version. A v9 bump would satisfy every existing guard while `gen-state-fixture` overwrote a file still named v8. This PR is the rename, so the rename guardrail belongs in it.
- The module docs state "version 8 changes no wire layout" and the id table says "allocated by the v8 batch"; both rot as the slices land, and the PR-number and version-pin markers are the kind this repo bans in code comments.

CHANGELOG: 0.5.0 goes to the altitude-drum change (#65), which moves the corpus and the composition digest under an unchanged ABI 7. This branch's entry renumbers when the batch is ready to cut.

luofang34:
**Status, and the order this has to land in.**

The batch is complete: all six allocations merged (#76 bearing pointers 0x12, #75 airframe configuration 0x13, #74 nav deflection scale, #72 airspeed trend, #78 autoflight 0x14/0x15, plus #66 true airspeed). `GroupId::COUNT` is 17.

Two things stand between here and merging this one.

**1. #80 lands on this branch first.** It carries the documentation issue #57 required and #75 shipped without — `AIR-IN-016`, its intended-function row, the layer-protocol row, the one-crate-many-sets amendment, and a tightened group region. It is under verification now.

**2. Then `main` merges into this branch.** The batch is 59 commits ahead and **68 behind** — `main` has moved a long way underneath it, including a clippy fix (#83) that unblocked CI for every open PR, and three releases (#59, #61, #63) now numbered 0.5.1 through 0.5.3.

I previewed that merge with `git merge-tree`: **15 conflicts**.

```
CHANGELOG.md
crates/indicate-instrument-conformance/src/admission/tests.rs
crates/indicate-instrument-descriptor/src/states.rs
crates/indicate-instrument-raster/src/composition/tests.rs
docs/instruments/evidence-artifacts/att01/{panel,presentation,raster}.run.txt
docs/instruments/evidence-graph.evg
release-manifest.json
sets/indicate-instrument-panels/src/descriptors.rs
sets/indicate-instrument-panels/src/hsi.rs
sets/indicate-instrument-panels/src/hsi/cdi.rs
sets/indicate-instrument-panels/src/hsi/source_tests.rs
sets/indicate-instrument-panels/src/pfd/tapes.rs
tools/instrument-bench/src/screen.rs
```

Most are pinned values — digests, counts, golden frames, evidence artifacts. Those get recomputed by running the tests, never hand-resolved: this session already had one union merge silently drop an extreme-state registration, caught only because the admission ratchet came out 312 instead of 330.

**This must merge with a merge commit.** The batch carries exactly the evidence baselines a squash or rebase would orphan, which is the failure `scripts/check-baseline-survives-merge.sh` was added in #69 to warn about — it will now annotate this pull request rather than exiting green.

Then this closes the #58 batch story: one ABI revision carrying every group the batch allocated, so v8 names exactly one wire format.

Issue #44, opened 2026-08-08
A recorded pass count can go stale without any gate noticing


luofang34:
Found while reviewing #42. The recorded raster count in `docs/instruments/evidence-artifacts/att01/raster.run.txt` said 86; running `cargo test -p indicate-instrument-raster` on `main` at the time printed **98**. The record had drifted twelve tests and every gate was green.

## Why nothing caught it

The evidence gate is thorough about everything except the one thing it cannot do. It verifies that:

- the artifact hashes to its declared `output-digest`;
- its parsed fields agree with the result node's attrs;
- the declared baseline is reachable from HEAD;
- every case selector resolves against the tree;
- the bound source's blob matches its `source-digest`.

**It never re-runs the suite.** So a record stays internally consistent — correct hash, agreeing fields, reachable baseline — while its `summary:` describes a run that no longer resembles what the tree produces.

The `source-digest` check is what usually saves this: edit a bound source and the mismatch forces a re-record. But tests can be *added* to a suite without touching the file the graph names, and then nothing forces anything. That is exactly what happened here.

## Why it matters more than a wrong number

The count is the only part of the record that says what the run actually did. Everything else is identity. A record claiming 86 passes against a suite of 98 is a record of a run nobody performed, and the whole point of the artifact is that it is a statement about a real execution — the graph refuses a hand-authored one (`source-as-artifact`, `unexecuted-result`) precisely because that distinction is load-bearing.

## Possible shapes, none obviously right

- **Re-run in the gate.** Honest but expensive, and it makes the gate depend on a toolchain and a build, which it currently does not — it reads files and asks git.
- **Record the suite's test count as a separate pinned attr** and check it against a cheap enumeration (`cargo test -- --list`). Cheaper, still needs a build.
- **Bind the whole suite's sources rather than one locator per case**, so adding a test to the crate moves a digest. Closest to the existing mechanism, but the graph's case→locator model is per-case by design.
- **Accept it and say so** — document that a recorded count is only as fresh as its last re-record, and that the gate proves identity rather than currency.

The last one is the cheapest and might be the honest answer, in the same spirit as `check-standards-registry.sh` saying plainly that it proves internal consistency and not external freshness. But it should be a decision, not a gap.

Not urgent: the counts are correct as of `v0.4.0`, because #42's re-record refreshed all three.

PR #69, opened 2026-08-18
Merge evidence-touching changes without rewriting their baseline

luofang34:
Closes the policy question raised by #67.

A recorded run names the commit it was produced against. The gate refuses a baseline it cannot reach from HEAD, correctly — a fresh clone cannot fetch one. A squash merge replaces exactly that commit, so every evidence-touching squash leaves `main` red until someone re-anchors it. Independent review of #67 counted **eleven** such re-anchor commits in this repository's history; I had estimated three.

The alternative I floated first is worse than unavailable. Anchoring the record at an older commit that survives the squash buys reachability by making the record untrue: the bound sources still resolve at that commit, so both the evidence gate and the new freshness check pass, while the record claims a run against a tree that produces a different count. #44's own lane demonstrates it — the panel suite moved 81 to 84 without touching any bound source. Green and false, with nothing able to see it.

Rebase-and-merge orphans identically. Relaxing reachability contradicts a deliberate design that CI asserts with a negative fixture. So the choice is exactly two: the merge style preserves the lane commit, or the baseline stops being a commit.

This takes the first. #65 was merged this way as the test: `main` stayed green, with no re-anchor needed.

No guardrail ships with this because none is available — GitHub's merge-method restriction is a repository setting, not a file in the tree. If the setting is tightened to match, this paragraph becomes the record of why.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -235,6 +235,18 @@ jobs:
           bash scripts/check-recorded-counts.sh --selftest
           bash scripts/check-recorded-counts.sh
 
+      - name: baselines that a squash would orphan (advisory)
+        # Advisory by construction: a pull request that re-records
+        # evidence declares a baseline that is not yet on main, so this
+        # reports on every legitimate evidence change. Its value is that
+        # the requirement appears on the pull request page at the moment
+        # someone picks a merge button, rather than only in
+        # CONTRIBUTING.md. The hard reachability gate cannot cover this:
+        # a pull request build checks out the synthetic merge ref, where
+        # the lane commit is already an ancestor.
+        continue-on-error: true
+        run: bash scripts/check-baseline-survives-merge.sh
+
       - name: Negative fixtures must all FAIL the gate
         run: |
           set -u
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -235,7 +235,14 @@ jobs:
           bash scripts/check-recorded-counts.sh --selftest
           bash scripts/check-recorded-counts.sh
 
-      - name: baselines that a squash would orphan (advisory)
+      - name: the baseline advisory can still fail
+        # A hard gate, deliberately outside the advisory step below: the
+        # advisory carries `continue-on-error`, so a selftest run inside
+        # it could go red without anyone noticing — which is the exact
+        # failure the selftest exists to catch.
+        run: bash scripts/test-baseline-survives-merge.sh
+
+      - name: baselines that a squash or rebase would orphan (advisory)
         # Advisory by construction: a pull request that re-records
         # evidence declares a baseline that is not yet on main, so this
         # reports on every legitimate evidence change. Its value is that
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,6 +39,8 @@ cargo run --locked -q -p indicate-evidence --bin evidence-gate -- \
   --graph docs/instruments/evidence-graph.evg --repo-root . --require-resolvable
 bash scripts/check-recorded-counts.sh --selftest
 bash scripts/check-recorded-counts.sh
+bash scripts/test-baseline-survives-merge.sh
+bash scripts/check-baseline-survives-merge.sh
 ```
 
 The evidence gate binds recorded test sources by content digest: editing
@@ -147,14 +149,18 @@ the tag is what a human owes it.
   advisory step on every pull request: the step exits non-zero when it
   has something to say, so the requirement is on the page when someone
   picks a merge button.
-- **Do not re-anchor the record at an older commit instead.** When the
+- **Do not move the record back to an older commit instead.** When the
   branch changed a bound source, the gate refuses loudly, because the
   older baseline never contained that source. When it changed only a
-  count, both gates pass while the record claims a run against a tree
-  that produces a different count. The first is noisy and the second is
-  silent, and neither is a way to keep the record true.
-- Re-anchoring after the merge does work, and this repository has done
-  it repeatedly. It leaves `main` red until the re-anchor lands.
+  count, the reachability check and the artifact check both pass while
+  the record claims a run against a tree that produces a different
+  count. The first is noisy and the second is silent, and neither keeps
+  the record true.
+- Re-anchoring the record after the merge does work, and this
+  repository has done it repeatedly. It leaves `main` red until the
+  re-anchor lands. Merge style is what preserves the commit a record
+  names; changing what a baseline *is* would be a change to the gate,
+  not to merge policy.
 - Panels are authored against the frame range their descriptor declares
   and receive the chosen frame as a draw argument; unclipped ink past
   that frame's edge is counted and ratcheted per frame by the admission
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -138,13 +138,15 @@ the tag is what a human owes it.
   recorded source file requires re-recording that evidence, and history
   rewrites that orphan the baseline commit are forbidden.
 - **Use a merge commit when a branch changes a `config-digest` in the
-  evidence graph. Do not use a squash merge.** The test is mechanical:
-  `git diff` shows whether the branch moved one. A squash replaces the
-  commit that the record names, and the gate then refuses the record as
-  unreachable, correctly, because a fresh clone cannot fetch it.
-  `scripts/check-baseline-survives-merge.sh` reports which baselines a
-  squash would orphan, and CI runs it as an advisory step on every pull
-  request.
+  evidence graph. Do not squash and do not rebase.** The test is
+  mechanical: `git diff` shows whether the branch moved one. Squash and
+  rebase both replace the commit that the record names, and the gate
+  then refuses the record as unreachable, correctly, because a fresh
+  clone cannot fetch it. `scripts/check-baseline-survives-merge.sh`
+  reports which baselines either would orphan, and CI runs it as an
+  advisory step on every pull request: the step exits non-zero when it
+  has something to say, so the requirement is on the page when someone
+  picks a merge button.
 - **Do not re-anchor the record at an older commit instead.** When the
   branch changed a bound source, the gate refuses loudly, because the
   older baseline never contained that source. When it changed only a
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -137,16 +137,22 @@ the tag is what a human owes it.
   sources by content digest and its baseline by commit: editing a
   recorded source file requires re-recording that evidence, and history
   rewrites that orphan the baseline commit are forbidden.
-- **Merge a pull request that touches recorded evidence with a merge
-  commit, not a squash.** A squash replaces the commit the run was
-  produced against, and the gate then refuses the record as unreachable
-  — correctly, because a fresh clone cannot fetch it. Re-anchoring
-  afterwards works but leaves `main` red until someone notices, and this
-  repository has paid that eleven times. Anchoring the record at an
-  older commit that survives the squash is not the way out: the bound
-  sources still resolve there, so the gate passes while the record
-  claims a run against a tree that produces different counts. Green and
-  false is worse than red and true.
+- **Use a merge commit when a branch changes a `config-digest` in the
+  evidence graph. Do not use a squash merge.** The test is mechanical:
+  `git diff` shows whether the branch moved one. A squash replaces the
+  commit that the record names, and the gate then refuses the record as
+  unreachable, correctly, because a fresh clone cannot fetch it.
+  `scripts/check-baseline-survives-merge.sh` reports which baselines a
+  squash would orphan, and CI runs it as an advisory step on every pull
+  request.
+- **Do not re-anchor the record at an older commit instead.** When the
+  branch changed a bound source, the gate refuses loudly, because the
+  older baseline never contained that source. When it changed only a
+  count, both gates pass while the record claims a run against a tree
+  that produces a different count. The first is noisy and the second is
+  silent, and neither is a way to keep the record true.
+- Re-anchoring after the merge does work, and this repository has done
+  it repeatedly. It leaves `main` red until the re-anchor lands.
 - Panels are authored against the frame range their descriptor declares
   and receive the chosen frame as a draw argument; unclipped ink past
   that frame's edge is counted and ratcheted per frame by the admission
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -137,6 +137,16 @@ the tag is what a human owes it.
   sources by content digest and its baseline by commit: editing a
   recorded source file requires re-recording that evidence, and history
   rewrites that orphan the baseline commit are forbidden.
+- **Merge a pull request that touches recorded evidence with a merge
+  commit, not a squash.** A squash replaces the commit the run was
+  produced against, and the gate then refuses the record as unreachable
+  — correctly, because a fresh clone cannot fetch it. Re-anchoring
+  afterwards works but leaves `main` red until someone notices, and this
+  repository has paid that eleven times. Anchoring the record at an
+  older commit that survives the squash is not the way out: the bound
+  sources still resolve there, so the gate passes while the record
+  claims a run against a tree that produces different counts. Green and
+  false is worse than red and true.
 - Panels are authored against the frame range their descriptor declares
   and receive the chosen frame as a draw argument; unclipped ink past
   that frame's edge is counted and ratcheted per frame by the admission
--- a/scripts/check-baseline-survives-merge.sh
+++ b/scripts/check-baseline-survives-merge.sh
@@ -1,0 +1,55 @@
+#!/usr/bin/env bash
+# Advisory: says whether this branch's recorded baselines would survive a
+# squash merge (#69).
+#
+# A recorded run names the commit it was produced against, and the gate
+# refuses a baseline it cannot reach from HEAD — a fresh clone cannot
+# fetch one. A squash merge replaces exactly that commit. So a branch
+# whose baseline is not already on the base branch must merge with a
+# merge commit, or main goes red the moment it lands.
+#
+# This cannot be a hard gate, and the reason is worth stating: on a pull
+# request build the checkout is the synthetic merge ref, where the lane
+# commit IS an ancestor, so the hard gate passes before the merge and
+# fails only after it. That asymmetry is why every episode of this was
+# found late. The point of an advisory step is to put the requirement on
+# the pull request page at the moment someone picks a merge button.
+set -euo pipefail
+
+root_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+cd "$root_dir"
+
+graph="${INDICATE_EVIDENCE_GRAPH:-docs/instruments/evidence-graph.evg}"
+base="${INDICATE_MERGE_BASE:-origin/main}"
+
+[ -f "$graph" ] || {
+    echo "check-baseline-survives-merge: no graph at $graph" >&2
+    exit 0
+}
+
+if ! git rev-parse --verify --quiet "$base" >/dev/null; then
+    echo "check-baseline-survives-merge: $base is not fetched; skipping"
+    exit 0
+fi
+
+branch_local=0
+for digest in $(sed -n 's/^attr config-digest \([0-9a-f]\{40\}\)$/\1/p' "$graph" | sort -u); do
+    if ! git cat-file -e "$digest^{commit}" 2>/dev/null; then
+        echo "UNKNOWN: baseline $digest is not an object in this clone" >&2
+        continue
+    fi
+    if git merge-base --is-ancestor "$digest" "$base" 2>/dev/null; then
+        echo "ok: baseline $digest is already on $base"
+    else
+        echo "BRANCH-LOCAL: baseline $digest exists only on this branch" >&2
+        branch_local=$((branch_local + 1))
+    fi
+done
+
+if [ "$branch_local" -ne 0 ]; then
+    echo "check-baseline-survives-merge: $branch_local baseline(s) would be orphaned by a squash." >&2
+    echo "Merge this pull request with a merge commit. See CONTRIBUTING.md." >&2
+    exit 0
+fi
+
+echo "check-baseline-survives-merge: every baseline is already on $base"
--- a/scripts/check-baseline-survives-merge.sh
+++ b/scripts/check-baseline-survives-merge.sh
@@ -1,19 +1,28 @@
 #!/usr/bin/env bash
 # Advisory: says whether this branch's recorded baselines would survive a
-# squash merge (#69).
+# merge that rewrites its commits.
 #
 # A recorded run names the commit it was produced against, and the gate
 # refuses a baseline it cannot reach from HEAD — a fresh clone cannot
-# fetch one. A squash merge replaces exactly that commit. So a branch
-# whose baseline is not already on the base branch must merge with a
-# merge commit, or main goes red the moment it lands.
+# fetch one. A squash merge replaces exactly that commit, and so does a
+# rebase merge. So a branch whose baseline is not already on the base
+# branch must merge with a merge commit, or main goes red the moment it
+# lands.
 #
 # This cannot be a hard gate, and the reason is worth stating: on a pull
 # request build the checkout is the synthetic merge ref, where the lane
-# commit IS an ancestor, so the hard gate passes before the merge and
+# commit IS an ancestor, so a hard gate passes before the merge and
 # fails only after it. That asymmetry is why every episode of this was
 # found late. The point of an advisory step is to put the requirement on
-# the pull request page at the moment someone picks a merge button.
+# the pull request page at the moment someone picks a merge button —
+# which requires this script to exit non-zero when it has something to
+# say. The workflow step carries `continue-on-error`, so a non-zero exit
+# marks the step without failing the build.
+#
+# Every path other than "checked, and every baseline is on the base"
+# exits non-zero. A check that cannot run and says nothing is
+# indistinguishable from a check that ran and found nothing, and the
+# second is the one a reviewer will assume.
 set -euo pipefail
 
 root_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
@@ -22,20 +31,32 @@ cd "$root_dir"
 graph="${INDICATE_EVIDENCE_GRAPH:-docs/instruments/evidence-graph.evg}"
 base="${INDICATE_MERGE_BASE:-origin/main}"
 
-[ -f "$graph" ] || {
-    echo "check-baseline-survives-merge: no graph at $graph" >&2
-    exit 0
-}
+if [ ! -r "$graph" ]; then
+    echo "UNCHECKED: no readable evidence graph at $graph" >&2
+    exit 1
+fi
 
 if ! git rev-parse --verify --quiet "$base" >/dev/null; then
-    echo "check-baseline-survives-merge: $base is not fetched; skipping"
-    exit 0
+    echo "UNCHECKED: $base is not fetched, so no baseline can be placed" >&2
+    exit 1
+fi
+
+digests="$(sed -n 's/^attr config-digest \([0-9a-f]\{40\}\)$/\1/p' "$graph" | sort -u)" || {
+    echo "UNCHECKED: could not read baselines from $graph" >&2
+    exit 1
+}
+
+if [ -z "$digests" ]; then
+    echo "UNCHECKED: $graph declares no baseline; a graph with none is not a clean graph" >&2
+    exit 1
 fi
 
 branch_local=0
-for digest in $(sed -n 's/^attr config-digest \([0-9a-f]\{40\}\)$/\1/p' "$graph" | sort -u); do
+unknown=0
+for digest in $digests; do
     if ! git cat-file -e "$digest^{commit}" 2>/dev/null; then
         echo "UNKNOWN: baseline $digest is not an object in this clone" >&2
+        unknown=$((unknown + 1))
         continue
     fi
     if git merge-base --is-ancestor "$digest" "$base" 2>/dev/null; then
@@ -46,10 +67,16 @@ for digest in $(sed -n 's/^attr config-digest \([0-9a-f]\{40\}\)$/\1/p' "$graph"
     fi
 done
 
+if [ "$unknown" -ne 0 ]; then
+    echo "check-baseline-survives-merge: $unknown baseline(s) could not be placed." >&2
+    exit 1
+fi
+
 if [ "$branch_local" -ne 0 ]; then
-    echo "check-baseline-survives-merge: $branch_local baseline(s) would be orphaned by a squash." >&2
-    echo "Merge this pull request with a merge commit. See CONTRIBUTING.md." >&2
-    exit 0
+    echo "check-baseline-survives-merge: $branch_local baseline(s) would be orphaned by a" >&2
+    echo "squash or rebase merge. Merge this pull request with a merge commit." >&2
+    echo "See CONTRIBUTING.md." >&2
+    exit 1
 fi
 
 echo "check-baseline-survives-merge: every baseline is already on $base"
--- a/scripts/check-baseline-survives-merge.sh
+++ b/scripts/check-baseline-survives-merge.sh
@@ -53,7 +53,9 @@ digests="$(sed -n 's/^attr config-digest *\([0-9a-fA-F]\{7,40\}\) *$/\1/p' "$gra
 # Distinct declared values, so this compares like with like: a gap
 # between the two counts means the graph declares a baseline in a form
 # this script did not parse, which is the shape that fails open.
-declared="$(sed -n 's/^attr config-digest *//p' "$graph" | sed 's/ *$//' | sort -u | grep -c .)"
+declared="$(sed -n 's/^attr config-digest *//p' "$graph" | sed 's/ *$//' | sort -u \
+    | grep -c . || true)"
+declared="${declared:-0}"
 if [ -z "$digests" ]; then
     echo "UNCHECKED: $graph declares no baseline this script can parse" >&2
     exit 1
--- a/scripts/check-baseline-survives-merge.sh
+++ b/scripts/check-baseline-survives-merge.sh
@@ -41,19 +41,29 @@ if ! git rev-parse --verify --quiet "$base" >/dev/null; then
     exit 1
 fi
 
-digests="$(sed -n 's/^attr config-digest \([0-9a-f]\{40\}\)$/\1/p' "$graph" | sort -u)" || {
+# Every form the gate accepts must be parsed here, or the guardrail is
+# narrower than the thing it guards and skips the difference in silence.
+# The gate hands the attribute to git after a trim, and git resolves
+# abbreviated and upper-case object ids, so this does too.
+digests="$(sed -n 's/^attr config-digest *\([0-9a-fA-F]\{7,40\}\) *$/\1/p' "$graph" | sort -u)" || {
     echo "UNCHECKED: could not read baselines from $graph" >&2
     exit 1
 }
 
+# Distinct declared values, so this compares like with like: a gap
+# between the two counts means the graph declares a baseline in a form
+# this script did not parse, which is the shape that fails open.
+declared="$(sed -n 's/^attr config-digest *//p' "$graph" | sed 's/ *$//' | sort -u | grep -c .)"
 if [ -z "$digests" ]; then
-    echo "UNCHECKED: $graph declares no baseline; a graph with none is not a clean graph" >&2
+    echo "UNCHECKED: $graph declares no baseline this script can parse" >&2
     exit 1
 fi
 
 branch_local=0
 unknown=0
+checked=0
 for digest in $digests; do
+    checked=$((checked + 1))
     if ! git cat-file -e "$digest^{commit}" 2>/dev/null; then
         echo "UNKNOWN: baseline $digest is not an object in this clone" >&2
         unknown=$((unknown + 1))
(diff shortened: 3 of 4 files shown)


luofang34:
**REQUEST CHANGES** — the rule in CONTRIBUTING.md is right and the script detects the real case, but every exit path is `0`, so the step is green in exactly the situation it exists to announce; and three separate malfunctions (unparseable graph, unreadable graph, unknown object) end in the script printing its *success* line.

Reviewed at `3623121`, worktree detached on the PR head. Full battery below.

---

## Blocking

### 1. The check cannot be seen. Every path exits 0, so `continue-on-error: true` is dead code.

`scripts/check-baseline-survives-merge.sh:52` returns `0` after printing `BRANCH-LOCAL`. I built the real failure case — a commit that exists in the object store and is not an ancestor of `origin/main`, which is exactly the shape of a re-recorded evidence branch:

```console
$ NEW=$(git commit-tree 'HEAD^{tree}' -p HEAD -m probe)   # 907bd19…, on no branch
$ sed -i '' "s/e3a9afa6b1d68db4a4ff083079a45d298569dd10/$NEW/g" docs/instruments/evidence-graph.evg
$ bash scripts/check-baseline-survives-merge.sh; echo "exit=$?"
BRANCH-LOCAL: baseline 907bd19bed885bd7ac197f2f44f341a66206ecd5 exists only on this branch
check-baseline-survives-merge: 1 baseline(s) would be orphaned by a squash.
Merge this pull request with a merge commit. See CONTRIBUTING.md.
exit=0
```

The detection is correct. The delivery is not. `.github/workflows/ci.yml:247` sets `continue-on-error: true`, which only matters for a step that can fail — this one cannot. The result is a step that reports success on the branch it was written to warn about, and the header's stated purpose (`:15-16`, "put the requirement on the pull request page at the moment someone picks a merge button") is not met: a green step's log is not on the pull request page. Nobody opens the log of a check that passed.

Two ways to actually deliver it, both compatible with "advisory":

- `exit 1` at `:52`. `continue-on-error: true` then does the job it is already written to do — job stays green, the step does not, and the failure is visible without opening anything.
- Emit `echo "::warning::…"` and/or append to `$GITHUB_STEP_SUMMARY`. Annotations surface on the checks tab whether or not the step failed.

As it stands the workflow and the script disagree about which layer makes this advisory, and the redundancy costs the entire signal.

### 2. Zero digests parsed is reported as "every baseline is already on origin/main".

`scripts/check-baseline-survives-merge.sh:36` builds the work list with `sed -n 's/^attr config-digest \([0-9a-f]\{40\}\)$/\1/p'`. When that yields nothing the loop body never runs, `branch_local` stays `0`, and `:55` prints the all-clear. The script has no way to distinguish "I checked every baseline and they are fine" from "I found no baselines".

Malformed digest — one hex character short, which is what a hand-edit typo produces:

```console
$ sed -i '' 's/…dd10$/…dd1/' docs/instruments/evidence-graph.evg   # all three, 39 chars
$ bash scripts/check-baseline-survives-merge.sh; echo "exit=$?"
check-baseline-survives-merge: every baseline is already on origin/main
exit=0
```

Graph present but unreadable — `sed` fails, and the script reports success on top of the error:

```console
$ chmod 000 docs/instruments/evidence-graph.evg
$ bash scripts/check-baseline-survives-merge.sh; echo "exit=$?"
sed: docs/instruments/evidence-graph.evg: Permission denied
check-baseline-survives-merge: every baseline is already on origin/main
exit=0
```

`set -euo pipefail` does not help here: `errexit` does not fire on a failing command substitution in a `for` word list.

This is the same failure shape this repository has already shipped once — a completeness check that failed open when a `sed` returned the wrong number of lines. The fix is a counter: parse into a variable, and if the count is zero, say `UNCHECKED: no config-digest found in <graph>` and fail closed. The same counter protects against the format drifting (a 64-hex sha256 digest, an indented `attr`, a trailing comment, CRLF — every one of those silently becomes "all clear" today).

### 3. `UNKNOWN` is never counted, and the final line contradicts it.

`:37-40` prints `UNKNOWN` and `continue`s without touching any counter, so `:55` then asserts the opposite:

```console
$ sed -i '' 's/e3a9…dd10/0123456789abcdef0123456789abcdef01234567/g' docs/instruments/evidence-graph.evg
$ bash scripts/check-baseline-survives-merge.sh; echo "exit=$?"
UNKNOWN: baseline 0123456789abcdef0123456789abcdef01234567 is not an object in this clone
check-baseline-survives-merge: every baseline is already on origin/main
exit=0
```

A baseline that is not an object in the clone is precisely the post-squash state this check exists to predict, and here it is being summarised as fine. Track `unknown` alongside `branch_local` and make the closing line conditional on both being zero.

### 4. No selftest, and the PR's own CI run never executes the branch that matters.

There is no `scripts/test-baseline-survives-merge.sh`, and no `--selftest` flag. Every other guard in `scripts/` proves itself on every run — I watched them do it in this battery:

```
release-manifest-selftest: OK (8 cases)
instrument-requirements-selftest: OK (6 cases)
standards-registry-selftest: OK (9 cases)
recorded-counts-selftest: OK (4 cases)
```

This PR does not move a `config-digest`, so the only path its own CI run exercises is `ok:`. Findings 1 through 3 are all in code that no automated run has ever entered. CONTRIBUTING.md:59-61 — this PR's own file — requires the guardrail in the same PR as the fix. A selftest here is cheap and needs no fixtures: `git commit-tree` a dangling commit as above, point a temporary graph at it via the `INDICATE_EVIDENCE_GRAPH` hook the script already provides, and assert the exit code and the `BRANCH-LOCAL` line; then the empty-parse and unknown-object cases the same way.

---

## Non-blocking

- **`scripts/check-baseline-survives-merge.sh:3` — `(#69)`.** A pull-request number in a source comment is the rot-prone marker the project bans; it scatters and goes stale when history is rewritten, which is a slightly ironic failure mode for this particular script. The sentence reads fine without it.

- **CONTRIBUTING.md:140-147 forbids squash only; rebase-and-merge is enabled on this repository and breaks identically.**
  ```console
  $ gh api repos/luofang34/Indicate --jq '{allow_merge_commit,allow_squash_merge,allow_rebase_merge}'
  {"allow_merge_commit":true,"allow_rebase_merge":true,"allow_squash_merge":true}
  ```
  Rebase-and-merge replays the branch as new commits and orphans the baseline exactly as a squash does. The script's own message gets this right ("Merge this pull request with a merge commit"); the rule bullet should say the same thing positively rather than naming one of the two ways to break it. Turning off both non-merge buttons in repository settings would be the durable version of this rule.

- **The local battery in CONTRIBUTING.md:15-42 was not updated.** The new script is not in the list, so a contributor running the documented battery never runs it, and the sentence at :12-13 ("CI runs this set plus the downstream-agnostic closure check") is now one item short.

- **`:25-28` and `:30-33` exit 0 on "cannot check".** Both print a message, which is better than silence, but both still leave a green step. `:31` also misreports a broken git as a missing ref: with `GIT_DIR` pointing at nothing, `git rev-parse` writes `fatal: not a git repository` and the script concludes `origin/main is not fetched; skipping`. Same counter as finding 2 fixes all of these.

- **Prose.** `:154-155` "Re-anchoring after the merge does work, and this repository has done it repeatedly" is repository narrative rather than a rule, and belongs in the commit message; the bullet's actual content is the second sentence. `:152-153` "The first is noisy and the second is silent, and neither is a way to keep the record true" packs three statements into one sentence. Script header `:14` "That asymmetry is why every episode of this was found late" is the same. The rule bullets themselves (`:140-144`, `:148-151`) are accurate and I have no correction to the technical content.

- **shellcheck SC2013** on `:36` (word-splitting a `sed` pipeline). Info-level, shellcheck is not a CI gate here, and `check-structure.sh:346` carries an info finding too — but the `while read` form is also what makes the counter in finding 2 natural to write.

---

- **CI wiring is correct.** The step is in `quality-gates`, whose checkout sets `fetch-depth: 0` (`.github/workflows/ci.yml:15-21`), so `origin/main` is a real ref there — corroborated by the hard reachability gate at `:218` depending on the same history. The rationale for comparing against `origin/main` rather than `HEAD` is right: on the synthetic merge ref the branch commit is already an ancestor of `HEAD`, and `origin/main` is not affected by that.
- **The attribute read is the authoritative one.** `config-digest` is what the gate treats as the baseline (`crates/indicate-evidence/src/gate/baseline.rs:29`), and the gate separately requires it to equal the justifying configuration-item's `attr commit`, so reading one field cannot drift from the other.
- **Positive case is honest**, not a vacuous pass: `ok: baseline e3a9afa6… is already on origin/main` names the digest it resolved.

## Battery, run on `3623121` with a clean worktree

| command | result |
| --- | --- |
| `cargo fmt --all --check` | exit 0 |
| `cargo clippy --locked --all-targets --workspace -- -D warnings` | exit 0, no warnings |
| `cargo test --locked --workspace --all-targets --no-fail-fast` | 21 suites, all `ok`, 0 failed |
| `RUSTDOCFLAGS="-D missing_docs -D rustdoc::broken_intra_doc_links" cargo doc --locked --workspace --no-deps` | exit 0 |
| `cargo build --locked --release --workspace` | exit 0 |
| `check-structure.sh` | OK (12 selftest cases) |
| `check-release-manifest.sh` | OK (8 selftest cases) |
| `check-release-markers.sh` | OK (0.5.0; ABI 7, scene 1, corpus 5) |
| `check-instrument-requirements.sh` | OK (6 selftest cases, 73 requirements) |
| `check-standards-registry.sh` | OK (9 selftest cases) |
| `check-recorded-counts.sh --selftest` / `check-recorded-counts.sh` | OK (4 cases) / OK (3 records) |
| `check-baseline-survives-merge.sh` | OK — and see finding 1 for what that word is worth here |
| `evidence-gate --require-resolvable` | `verdict: VALID` |

All graph mutations were reverted; `git status` clean.

luofang34:
Took all the blocking findings in `a99b2a5`. Your first point was the one that mattered: the step was green on exactly the branch it exists to warn about, which made `continue-on-error` dead code and the stated purpose — putting the requirement on the page when someone picks a merge button — undelivered.

**Every path returned 0.** Now only one does: checked, and every baseline is on the base. `BRANCH-LOCAL`, `UNKNOWN`, an unparsable graph, an unreadable graph, and an unfetched base each exit non-zero with the reason. `continue-on-error` then does the job it was written for — the step is marked, the build is not failed.

**The three fail-open paths.** All three confirmed and closed. The `UNKNOWN` case was the worst of them: it printed the finding and then contradicted it in the summary line, which is a check arguing with itself. It is counted now.

**No selftest.** Six cases, one per path, asserting on exit status rather than text. I verified it catches the original defects by reintroducing them: restoring `exit 0` on the `BRANCH-LOCAL` path reddens `a branch-local baseline is refused`, and deleting the `unknown` counter reddens `an unplaceable baseline is refused`.

It runs as its own hard gate **outside** the advisory step. A selftest inside a `continue-on-error` step can go red without anyone noticing, which is the same failure one level up — and given what you just found, not a hypothetical one.

**Rebase-and-merge.** You are right that the rule named squash alone while rebase orphans the baseline identically and is enabled on the repo. Both are named, in CONTRIBUTING and in the step name.

Also took the `(#69)` breadcrumb in the script header — a banned rot-prone marker, and mine.

Left for its own change: CONTRIBUTING's local-battery list. It is a real gap, and adding one script to a list that is stale in other ways would paper over the rest; the list wants one pass with something checking it.

Verified: `bash scripts/test-baseline-survives-merge.sh` → OK (6 cases); the check itself → `every baseline is already on origin/main`, exit 0; the full workspace suite and every other `check-*.sh` green.

luofang34:
**APPROVE WITH NITS** — all four blocking findings are closed, and I confirmed the selftest is load-bearing by reintroducing each defect myself: six mutations, six distinct red cases, no overlap. One new non-blocking finding in the selftest, and one carried-over nit that this PR made more important.

Re-reviewed at `a99b2a5`.

---

## Blocking findings — all closed

### 1. Every path exits 0 → **CLOSED**

Seven probes, exit status captured on each. Only the clean case returns 0:

```console
$ bash scripts/check-baseline-survives-merge.sh                       # clean tree
ok: baseline e3a9afa6… is already on origin/main
check-baseline-survives-merge: every baseline is already on origin/main
exit=0

$ NEW=$(git commit-tree 'HEAD^{tree}' -p HEAD -m probe2)              # c2d7de6…, on no branch
$ sed -i '' "s/e3a9afa6…/$NEW/g" docs/instruments/evidence-graph.evg
$ bash scripts/check-baseline-survives-merge.sh
BRANCH-LOCAL: baseline c2d7de6de695fe2ecb26ac65ab335e92ce01dcb1 exists only on this branch
check-baseline-survives-merge: 1 baseline(s) would be orphaned by a
squash or rebase merge. Merge this pull request with a merge commit.
exit=1
```

| probe | message | exit |
| --- | --- | --- |
| clean tree | `every baseline is already on origin/main` | 0 |
| dangling commit as baseline | `BRANCH-LOCAL` | **1** |
| digest one character short | `UNCHECKED: … declares no baseline` | **1** |
| 40-hex naming no object | `UNKNOWN` + `could not be placed` | **1** |
| graph `chmod 000` | `UNCHECKED: no readable evidence graph` | **1** |
| graph absent | `UNCHECKED: no readable evidence graph` | **1** |
| `INDICATE_MERGE_BASE=origin/does-not-exist` | `UNCHECKED: … is not fetched` | **1** |
| `GIT_DIR=/nonexistent/.git` | `UNCHECKED: … is not fetched` | **1** |

The header comment at `:22-28` now states the invariant directly — "Every path other than *checked, and every baseline is on the base* exits non-zero" — and the code matches it at `:36`, `:41`, `:46`, `:51`, `:72`, `:79`. `continue-on-error: true` is no longer dead: it is now the thing that keeps the job green while the step goes red, which is what "advisory" should have meant from the start.

### 2. Zero digests parsed reported as success → **CLOSED**

`:44-52` splits the two cases the old code conflated. The parse is captured (`digests=…`) with its own failure branch, and an empty parse is its own refusal:

```console
$ bash scripts/check-baseline-survives-merge.sh   # all three digests truncated to 39 chars
UNCHECKED: docs/instruments/evidence-graph.evg declares no baseline; a graph with none is not a clean graph
exit=1
```

The unreadable-graph variant that previously printed `sed: Permission denied` followed by the all-clear now stops at `:34-37` before `sed` runs at all.

### 3. `UNKNOWN` uncounted and contradicted → **CLOSED**

```console
$ bash scripts/check-baseline-survives-merge.sh   # digest replaced with 0123456789abcdef…
UNKNOWN: baseline 0123456789abcdef0123456789abcdef01234567 is not an object in this clone
check-baseline-survives-merge: 1 baseline(s) could not be placed.
exit=1
```

The all-clear line at `:84` is now reachable only when both counters are zero.

### 4. No selftest → **CLOSED**, and the selftest is real

`scripts/test-baseline-survives-merge.sh` passes 6/6 here. I did not take that on trust — I reintroduced each defect into `check-baseline-survives-merge.sh` one at a time and re-ran the selftest. Each mutation reddens exactly one case, and no case is carried by another:

| mutation to the check | case that went red |
| --- | --- |
| `:79` `exit 1` → `exit 0` (branch-local) | `REGRESSION: a branch-local baseline is refused — wanted exit 1, got 0` |
| `:72` `exit 1` → `exit 0` (unknown) | `REGRESSION: an unplaceable baseline is refused` |
| `:51` `exit 1` → `exit 0` (empty parse) | `REGRESSION: a graph with no parsable baseline is refused` |
| `:36` `exit 1` → `exit 0` (unreadable graph) | `REGRESSION: an unreadable graph is refused` |
| `:41` `exit 1` → `exit 0` (unfetched base) | `REGRESSION: an unfetched base is refused` |
| `exit 1` appended, so the clean path fails | `REGRESSION: a baseline on the base is accepted — wanted exit 0, got 1` |

Every one reported `FAILED (1 of 6 cases)` and exited 1. That is a selftest, not a decoration.

**CI wiring is right.** `.github/workflows/ci.yml:238-243` runs the selftest as a plain step in `quality-gates` with no `continue-on-error`, ahead of the advisory; `:245-255` keeps `continue-on-error: true` on the advisory alone. The only other `continue-on-error` in the file is the pre-existing trace step at `:204`, and the job carries none. So a selftest regression fails the build and the advisory does not — which is the arrangement the step comment claims, verified rather than asserted.

### Also closed

- **`(#69)` breadcrumb** — gone.
- **Rebase-and-merge** — named in CONTRIBUTING.md:141, in the step name at `:245`, in the header at `:8-10`, and in the refusal message itself. Matches the repository settings (`allow_rebase_merge: true`).
- **shellcheck** — now clean on both scripts (`shellcheck scripts/check-baseline-survives-merge.sh scripts/test-baseline-survives-merge.sh` → exit 0). SC2013 went with the `digests` variable.

---

## New, non-blocking: the selftest's branch-local case can skip itself and still count as passed

`scripts/test-baseline-survives-merge.sh:55-61`:

```sh
if [ "$head_commit" != "$on_base" ] && ! git merge-base --is-ancestor "$head_commit" "$on_base"; then
    …expect 1 "a branch-local baseline is refused"
else
    echo "ok: skipped the branch-local case — this checkout is on the base"
    passed=$((passed + 1))
fi
```

When `HEAD` is the base, the case does not run, prints `ok:`, increments `passed`, and the summary reports six. With the fail-open from finding 1 reintroduced at `:79`:

```console
$ sed -i '' '79s/exit 1/exit 0/' scripts/check-baseline-survives-merge.sh
$ INDICATE_MERGE_BASE=HEAD bash scripts/test-baseline-survives-merge.sh
ok: a baseline on the base is accepted
ok: skipped the branch-local case — this checkout is on the base
ok: an unplaceable baseline is refused
ok: a graph with no parsable baseline is refused
ok: an unreadable graph is refused
ok: an unfetched base is refused
baseline-survives-merge-selftest: OK (6 cases)
exit=0
```

Green, six cases claimed, five run, and the one regression the whole file exists to catch is the one that was skipped. `HEAD == origin/main` is the shape of a push-to-`main` build, and this workflow triggers `on: push:`.

I am not treating this as blocking, because the configuration that gates a merge does run the case: on a pull request build `HEAD` is the synthetic merge ref, and on a feature-branch push `HEAD` is the branch tip — neither is an ancestor of `origin/main`, and I confirmed all six mutations go red in that configuration. A regression cannot reach `main` through the front door unseen.

It is still worth fixing, for two reasons. The reporting is dishonest by this repository's own standard — every other script here is scrupulous about saying what it did *not* prove, and `OK (6 cases)` when five ran is the opposite. And the skip is unnecessary: the branch-local case does not need `HEAD` to be off the base, because `git commit-tree` synthesizes a commit that is on no branch at all, which is a stronger construction than the current one and works unconditionally:

```console
$ PROBE=$(git commit-tree 'HEAD^{tree}' -p HEAD -m selftest-probe)
$ git merge-base --is-ancestor "$PROBE" HEAD || echo "not an ancestor of HEAD, even though HEAD is the base"
not an ancestor of HEAD, even though HEAD is the base
```

Two lines, and `:55-61` collapses to the `then` branch with no condition. It is also exactly how I built the failure case in both rounds of this review, so it is known to reach the `BRANCH-LOCAL` path.

## Other non-blocking

- **The local battery in CONTRIBUTING.md:15-42 still lists neither script, and this now matters more than it did.** `test-baseline-survives-merge.sh` is a hard CI gate: it can fail the build, and a contributor who runs the documented battery before pushing will never have run it. The advisory is a judgement call to include — the hard gate is not.

- **`:39-42` still misdiagnoses a broken git as an unfetched ref.** With `GIT_DIR` pointing at nothing, `git rev-parse` prints `fatal: not a git repository` and the script concludes `origin/main is not fetched`. It fails closed now, so this is wording only — but "is not fetched" sends the reader to the wrong fix.

- **The selftest asserts exit status only**, and five of six cases expect the same `1`. A check that exited 1 unconditionally would pass five of them; only `a baseline on the base is accepted` separates them. Matching the leading token of the message (`UNCHECKED:` / `UNKNOWN:` / `BRANCH-LOCAL:`) would pin which path ran without making the test brittle about wording. Minor — the exit status is the contract the workflow consumes, and the file says so deliberately at `:5-7`.

- **Prose, unchanged from round one and still non-blocking**: `check-baseline-survives-merge.sh:15` "That asymmetry is why every episode of this was found late" and CONTRIBUTING.md:156-157 "this repository has done it repeatedly" are repository narrative rather than rules; CONTRIBUTING.md:154 packs three statements into one sentence.

## Battery, run on `a99b2a5` with a clean tree

| command | result |
| --- | --- |
| `cargo fmt --all --check` | exit 0 |
| `cargo clippy --locked --all-targets --workspace -- -D warnings` | exit 0, no warnings |
| `cargo test --locked --workspace --all-targets --no-fail-fast` | 21 suites `ok`, 0 failed |
| `RUSTDOCFLAGS="-D missing_docs -D rustdoc::broken_intra_doc_links" cargo doc --locked --workspace --no-deps` | exit 0 |
| `cargo build --locked --release --workspace` | exit 0 |
| `check-structure.sh` | OK (12 selftest cases) |
| `check-release-manifest.sh` | OK (8 selftest cases) |
| `check-release-markers.sh` | OK (0.5.0; ABI 7, scene 1, corpus 5) |
| `check-instrument-requirements.sh` | OK (6 selftest cases, 73 requirements) |
| `check-standards-registry.sh` | OK (9 selftest cases) |
| `check-recorded-counts.sh --selftest` / `check-recorded-counts.sh` | OK (4 cases) / OK (3 records) |
| `check-baseline-survives-merge.sh` | OK, exit 0 — and this time that word is earned |
| `test-baseline-survives-merge.sh` | OK (6 cases) |
| `shellcheck` on both new scripts | exit 0 |

All mutations reverted; `git status` clean.

luofang34:
**APPROVE WITH NITS** — all six blocking findings are closed, verified by construction rather than by reading. Five nits below, none blocking; the strongest (N1) is a silent-death path the script's own header argues against.

Verified at `50b4a3e`, detached on `origin/docs/merge-style-preserves-evidence-baseline` (both PR branches are checked out in other worktrees, so I reviewed the remote ref directly).

## Prior findings

**1. `exit 0` on `BRANCH-LOCAL` — CLOSED.** Built the case rather than trusting the selftest: `HEAD` (`50b4a3e`) is a lane commit and `git merge-base --is-ancestor HEAD origin/main` returns 1, so a graph declaring it is genuinely branch-local. The check exits **1**, and the annotation is on **stdout** where GitHub reads workflow commands:

```
::warning file=.../branchlocal.evg::1 baseline(s) would be orphaned by a squash or rebase merge; merge this pull request with a merge commit (see CONTRIBUTING.md)
```

The `$GITHUB_STEP_SUMMARY` append also lands — I pointed it at a real file and it contained the bolded line afterwards. Both halves of the claim hold; neither was taken on faith.

**2. Every exit path — CLOSED.** `BRANCH-LOCAL`, `UNKNOWN`, unparsable, unreadable graph, unfetched base, and count mismatch each exit non-zero; the only exit 0 I could produce was "checked, and every baseline is on the base." Against the real `docs/instruments/evidence-graph.evg`: `all 1 declared baseline(s) are on origin/main`, exit 0 — correct, since `e3a9afa` *is* on `origin/main`.

**3. Parser breadth — CLOSED.** Abbreviated (12-char), upper-case 40-char, and trailing-space forms are all **placed, not skipped** (exit 0 with a real `ok: baseline …` line each). No false all-clear from any mixed graph:

| graph | exit | via |
|---|---|---|
| on-base + branch-local | 1 | `BRANCH-LOCAL` + annotation |
| on-base + unknown object | 1 | `UNKNOWN` counter |
| on-base + `not-a-hex-value` | 1 | `declares 2 distinct baseline(s) but only 1 parsed` |
| on-base + TAB-separated digest | 1 | same mismatch — the loose `sed` sees it, the strict one does not, and the counter catches exactly that |
| 64-char sha256-shaped digest | 1 | `declares no baseline this script can parse` |

The TAB case is the one I most expected to fail open, and it does not. The declared-vs-parsed counter is doing real work, not decorating.

**4. Selftest is a hard step — CLOSED.** Parsed `ci.yml` with a YAML loader rather than eyeballing indentation: `the baseline advisory can still fail` → runs `test-baseline-survives-merge.sh`, `continue-on-error` **absent** (hard, fails the build); `baselines that a squash or rebase would orphan (advisory)` → `continue-on-error: True` (does not). Triggers are `push` and `pull_request`, both unfiltered, so the annotation reaches PR pages.

**Reintroducing each original defect turns the selftest red** — actual output, not prediction:

- restore `exit 0` on `BRANCH-LOCAL` → `REGRESSION: a branch-local baseline is refused — wanted exit 1, got 0` (1 of 10)
- narrow the regex to `[0-9a-f]{40}` → 3 of 10: abbreviated, upper-case, trailing space
- drop the `unknown` counter → 3 of 10: unplaceable, "no parsable baseline", mixed

**5. Both scripts in the local battery — CLOSED** (`CONTRIBUTING.md` lines 42–43).

**6. `re-anchor` / "both gates" — CLOSED.** The prohibition now reads "Do not move the record back to an older commit instead," leaving `re-anchor` with the single meaning the following bullet uses. `grep -rn "both gates"` returns nothing on the branch; it is now "the reachability check and the artifact check both pass," which names them.

## New findings (nits)

**N1 — a graph with zero `config-digest` lines exits 1 and prints nothing at all.** Not one byte on stdout or stderr. `declared="$(… | grep -c .)"` — `grep -c` exits 1 when it counts zero, and under `set -euo pipefail` that assignment kills the script before the `-z "$digests"` message on line 57 can run. It fails *closed*, so the advisory is not fail-open. But the script's own header says a check that cannot run and says nothing is indistinguishable from one that ran and found nothing, and this is precisely that path. `| grep -c . || true`, or hoisting the `-z` check above the `declared` assignment, restores the message. No selftest case covers it.

**N2 — selftest case 4 no longer tests what its comment claims.** The comment says the 39-character digest means "the pattern matches nothing and the loop never runs." Since the parser widened to `{7,40}`, 39 characters now match: the check reports `UNKNOWN: baseline 0123… is not an object in this clone` and exits through the unknown counter, making case 4 a duplicate of case 3. Confirmed both by running the check on that exact graph and by running the strict `sed` against it (it prints a parse). Consequence: the `declares no baseline this script can parse` branch has **no** coverage. A 6-hex-character value, or the 64-char sha256 shape I probed, would restore the intent — the comment is stale relative to the fix that widened the parser.

**N3 — the `checked != declared` branch has no selftest case.** It is the mechanism the third fix added, and it is the one thing standing between a novel digest form and a false all-clear. My probes reach it two ways, so the code is right; the guardrail-for-the-guardrail is what is missing. One case — an on-base digest plus a non-hex declaration — closes N2 and N3 together.

**N4 — the branch-local case self-skips and still counts as passed.** When the checkout sits on the base, the `else` prints `ok: skipped …` and increments `passed`, so the selftest reports `OK (10 cases)` either way. On a push build to `main` the `exit 0` mutation would therefore go undetected. PR builds do cover it — the synthetic merge ref is not an ancestor of `origin/main` — so this is a nit, but a throwaway commit would make the case unconditional instead of positional.

**N5 — cosmetic.** `graph_with "$work/short.evg" >/dev/null` (line 70) is vestigial: `graph_with` writes to the file itself, and line 71 overwrites it wholesale. Both the redirect and the call are no-ops.

## Battery

All green on this branch, run in this worktree:

```
cargo test --locked --workspace --all-targets --no-fail-fast   exit 0
check-structure.sh                OK
test-structure.sh                 OK (12 cases)
check-baseline-survives-merge.sh  all 1 declared baseline(s) are on origin/main
test-baseline-survives-merge.sh   OK (10 cases)
check-release-manifest.sh         OK      test-release-manifest.sh        OK (8 cases)
check-instrument-requirements.sh  OK (73) test-instrument-requirements.sh OK (6 cases)
check-certification-claims.sh     OK      check-standards-registry.sh     OK
test-standards-registry.sh        OK (9 cases)  check-release-markers.sh  OK
check-recorded-counts.sh          OK      trace-report.sh / detect-target.sh  OK
```

`shellcheck --severity=style` is clean on both new scripts. Every mutation was reverted with `git restore`; `git status --porcelain` is empty.

luofang34:
Both nits taken in `5f2b81e`, and the first one deserved to be blocking.

**The silent death.** You are right that this is the script's own stated failure mode turned on itself: `grep -c` exits non-zero on a zero count, `set -e` killed the assignment, and the path exited 1 with no output at all — indistinguishable from a crash, which is exactly what the header argues against. Reproduced it, fixed it, and the message now prints before the exit.

**The weakened selftest case.** Also right, and I had not noticed the interaction: widening the pattern to 7–40 characters turned the "one character short" case into a valid 39-character digest, so it silently became a duplicate of the unplaceable case and left two branches uncovered. There are now cases for a graph that declares nothing at all, a value the pattern cannot read, and one of each in the same graph — plus two that assert the script *speaks*, not only that it exits. Selftest is at 14 cases.

Worth recording, because it is the same mistake one level up: my first version of the "does it speak" case piped the check into `grep -q`, and this selftest runs under `pipefail`, so the pipeline carried the check's own non-zero status and reported a silent death that was not there. It captures the output instead. A test that lies about a failure is worse than no test, and I nearly shipped one while fixing a script for lying about a success.

I also verified the new cases catch the original defects rather than assuming: restoring `grep -c` without the guard reddens `a graph declaring no baseline says why before it exits`, and it was green before I added that case — so the coverage is real, not decorative.

Thank you for the TAB-separated digest probe. That one had not occurred to me, and it is a good demonstration that the declared-versus-parsed counter earns its place independently of the pattern: it catches forms the pattern was never widened for.

`shellcheck` clean on both scripts; structure gate, the workspace suite and every other `check-*.sh` green.

luofang34:
**Approve.** Both nits are closed. The zero-count path now speaks before it exits, the selftest is at 14 cases, and I could not find a case that cannot fail — I mutated the check eleven ways singly and five ways in pairs to try. Three non-blocking notes at the end, two of them from mutations that reddened nothing.

Reviewed at `5f2b81e` against `origin/main` (`830e9a6`), detached, in a scratch worktree.

## Nit 1 — a graph with zero `config-digest` lines exited 1 printing nothing

**Closed.** Verified live rather than by reading:

```
$ printf 'node RESULT-PROBE verification-result\n' > no-attr.evg
$ INDICATE_EVIDENCE_GRAPH=no-attr.evg bash scripts/check-baseline-survives-merge.sh
UNCHECKED: .../no-attr.evg declares no baseline this script can parse
exit=1
```

It speaks and it exits non-zero. A wholly empty file behaves the same way. The `|| true` plus `declared="${declared:-0}"` lets the zero count through so the `[ -z "$digests" ]` branch is reached, which is the branch that carries the message.

**Reintroducing `grep -c` without its guard reddens the selftest**, as asked. I put back the exact pre-fix line — `declared="$(… | grep -c .)"`, no `|| true`, no fallback:

```
baseline-survives-merge-selftest: FAILED (1 of 14 cases)
REGRESSION: the no-baseline path exits silently
```

Note *which* case caught it. `a graph declaring no baseline is refused` stayed green, because `set -e` killing the assignment still exits 1 — the exit status is identical before and after the fix. Only the case that asserts the script *speaks* can tell the two apart. That is the case the last commit added, and it is the only thing standing between this defect and a clean selftest.

## Nit 2 — case 4 had stopped testing what its comment claimed

**Closed.** The 39-character digest is gone; the case now feeds `not-a-digest`, which the `\{7,40\}` pattern genuinely cannot read. The no-parsable-baseline branch and the `checked != declared` counter both have cover now, and two of the new cases assert on output rather than exit status.

**Selftest is at 14 cases**, all passing. `HEAD` is not an ancestor of `origin/main` in a detached checkout, so the branch-local case ran for real rather than taking its skip branch — no "skipped" line in the output.

## Mutation matrix

I removed one guard at a time from `check-baseline-survives-merge.sh` and recorded which cases reddened. A case that survives every mutation is the defect this PR is about.

| mutation | cases reddened |
| --- | --- |
| `grep -c` without its guard (the pre-fix line) | *the no-baseline path exits silently* |
| `UNCHECKED: … declares no baseline` message deleted, `exit 1` kept | *the no-baseline path exits silently* · *the unparsable-graph path exits silently* |
| empty-digests branch removed | *a graph declaring no baseline is refused* |
| declared-vs-parsed counter removed | *a baseline the pattern cannot read refuses the whole run* |
| parser narrowed back to 40 lower-case hex | *an abbreviated baseline is placed* · *an upper-case baseline is placed* |
| trailing-space tolerance dropped | *a trailing space does not hide a baseline* |
| unknown-count refusal removed | *an unplaceable baseline is refused* · *a graph mixing a placeable and an unplaceable baseline is refused* |
| branch-local refusal removed | *a branch-local baseline is refused* |
| ancestor test forced to "no" | *a baseline on the base is accepted* · the three other accept cases |
| ancestor test forced to "yes" | *a branch-local baseline is refused* |
| unknown-object branch removed | nothing — see finding 1 |
| unfetched-base guard removed | nothing — see finding 1 |
| unreadable-graph guard removed | nothing — redundant, see below |
| empty-digests branch **and** counter removed | *a graph with no parsable baseline is refused* (+2) |
| branch-local refusal **and** unfetched-base guard removed | *an unfetched base is refused* (+1) |

**Every one of the nine cases added by the last two commits fails under a targeted mutation.** Eight fall to a single removal. The message-deletion row is the direct answer to "does each speaks case fail when the thing it covers is removed": delete the `UNCHECKED` line and keep `exit 1`, and both speaks cases redden while both companion `expect 1` cases stay green — which is exactly the split the two kinds of case exist to make.

The one that needs a double is *a graph with no parsable baseline is refused*: the `-z` branch and the declared-vs-parsed counter both catch `not-a-digest`, so either one alone still refuses it.

That overlap is worth naming so it is not mistaken for the defect being fixed. The old case 4 could *never* fail: its input was parsable, so it drove the same code path as the unplaceable case and asserted nothing the other case did not already assert. The new case drives a genuinely different input and is falsifiable — it just takes two guards down instead of one. Redundancy, not vacancy.

**One case I could not falsify at all: *an unreadable graph is refused*.** Three independent mechanisms each refuse an absent file — the `[ ! -r "$graph" ]` guard, the `|| { … exit 1; }` on the `sed` pipeline, and `set -e` itself. Removing the first leaves the selftest green because the second catches it. I am counting this as defense in depth rather than a finding: the input is distinct and the assertion is on real behaviour.

## Non-blocking findings

**1. Two guards can be deleted with the selftest still green, and the message the contributor reads becomes wrong.**

Removing the `git cat-file -e` unknown-object branch reddens nothing. Both cases that reach it assert only the exit status, and the fallthrough is `BRANCH-LOCAL`, which also exits 1. What changes is what gets said:

```
pristine:  UNKNOWN: baseline 0123…4567 is not an object in this clone
           check-baseline-survives-merge: 1 baseline(s) could not be placed.

mutated:   ::warning file=…::1 baseline(s) would be orphaned by a squash or rebase
           merge; merge this pull request with a merge commit (see CONTRIBUTING.md)
```

Removing the unfetched-base guard is the same shape, and worse, because the baseline in question is genuinely on the base:

```
pristine:  UNCHECKED: refs/heads/no-such-base is not fetched, so no baseline can be placed

mutated:   ::warning file=…::1 baseline(s) would be orphaned by a squash or rebase
           merge; merge this pull request with a merge commit (see CONTRIBUTING.md)
```

Both mutants post a merge-button instruction to the Checks tab about a commit that does not exist, or about one that is already on `main`. The script's header states the distinction that is being lost: "A check that cannot run and says nothing is indistinguishable from a check that ran and found nothing." The annotation surface added in `50b4a3e` turns that from a log line into something a contributor acts on, so UNCHECKED-versus-BRANCH-LOCAL is now the check's load-bearing output — and nothing tests it. The technique is already in the file: two cases capture output and assert it is non-empty. Asserting `UNKNOWN` and `UNCHECKED` appear on those two paths is a few lines and closes both.

**2. The two "says why" cases detect silence, not wrongness.** They assert the output is non-empty, which is precisely the defect they were written for and they do catch it. But under the `-z`-plus-counter double removal the script prints `all 0 declared baseline(s) are on origin/main` — the opposite of why — and both stay green. Their companion `expect 1` cases catch that case, so the pairs are sound as written; I mention it only because grepping for the expected substring costs nothing and would make each case self-sufficient. It is the same one-line change finding 1 asks for.

**3. The selftest is the hard gate, and it dies at exit 128 on a raw `git` error if its base ref is missing.** `test-baseline-survives-merge.sh` calls `git rev-parse "${INDICATE_MERGE_BASE:-origin/main}"` under `set -euo pipefail` before any case runs:

```
$ INDICATE_MERGE_BASE=refs/heads/definitely-not-a-ref bash scripts/test-baseline-survives-merge.sh
fatal: ambiguous argument 'refs/heads/definitely-not-a-ref': unknown revision or path not in the working tree.
exit=128
```

Zero cases run and the diagnosis is git's, not the selftest's. The advisory it guards handles exactly this input deliberately (`UNCHECKED: $base is not fetched, so no baseline can be placed`) — the selftest does not, and the selftest is the one that blocks the build. Not reachable in CI as configured, see below, so this is a robustness note rather than a live break.

**CI wiring checks out.** `.github/workflows/ci.yml` has a single `quality-gates` job whose `actions/checkout@v7` carries `fetch-depth: 0`, and both new steps live in it, so `refs/remotes/origin/main` is present when they run. The hard `test-baseline-survives-merge.sh` step sitting outside the `continue-on-error` advisory is the right call — a selftest inside a step that cannot fail is a selftest nobody reads. On a PR build the synthetic merge ref is a *descendant* of `origin/main`, not an ancestor, so the branch-local case runs there too rather than taking its skip branch.

## Gates

All from the detached PR head; `git status --porcelain` empty before and after every mutation.

- `bash scripts/check-structure.sh` — OK (12 selftest cases) · `bash scripts/test-structure.sh` — OK
- `bash scripts/check-baseline-survives-merge.sh` — OK, places 1 declared baseline on `origin/main`
- `bash scripts/test-baseline-survives-merge.sh` — OK (14 cases)
- `cargo test --locked --workspace --all-targets --no-fail-fast` — exit 0
- `check-certification-claims` OK · `check-instrument-requirements` OK (73 requirements) · `test-instrument-requirements` OK · `check-release-manifest` OK (8 cases) · `test-release-manifest` OK · `check-release-markers` OK · `check-standards-registry` OK (9 cases) · `test-standards-registry` OK · `check-recorded-counts --selftest` OK (4 cases) · `check-recorded-counts` OK (3 records) · `trace-report` OK · `detect-target` OK
- `evidence-gate --resolve-selectors` and `--require-resolvable` — VALID, 30 nodes / 73 edges
- `cargo run -p instrument-bench` — both digests match pin, 280 admission cases pass
- `shellcheck scripts/check-baseline-survives-merge.sh scripts/test-baseline-survives-merge.sh` — clean, no findings at any severity

PR #70, opened 2026-08-18
Show where the airspeed is going, not just where it is (#51)

### LINKED ISSUE #51: The speed tape shows no airspeed trend, and no trend quantity exists in state
The speed tape draws ticks, bands, and the IAS readout (`sets/indicate-instrument-panels/src/pfd/tapes.rs`), and the theme already names a token for this element — `gps` is documented as "GPS guidance, trends, and rate cues" (`crates/indicate-instrument-symbology/src/theme.rs:79-80`) — but no trend element consumes it: there is no trend bar, and no trend quantity anywhere in the state vocabulary.

The conventional presentation: a vertical bar from the readout centerline whose tip marks the airspeed predicted ~6 s ahead. It collapses at constant speed and grows with acceleration, which is what makes an approaching band edge readable before the needle gets there.

## Where the quantity comes from — the decision

1. **Wire field**: the source supplies d(IAS)/dt in the Air group (a layout move). Honest about provenance; costs an ABI move and every source must implement it.
2. **Feeder derivation**: difference successive accepted IAS samples the way `turn.rs` differences heading — same-stream only, acquisition-clock dt window, wrap-safe ordering (`crates/indicate-instrument-feeder/src/turn.rs`). This is DYN-01's discipline for derived rates.
3. **Presentation-side derivation**: cross-frame state exists in the presentation layer already (the chevron latches with hysteresis, `crates/indicate-instrument-state/src/presentation.rs:320-338`) — but a derivative taken there is a rate over render-clock deltas, which the DYN-01 design exists to avoid.

Option 2 fits the architecture. The natural owner of the resolved value is the Dynamics group — turn and slip already live there (`crates/indicate-instrument-state/src/dynamics.rs`) — with a valid bit from the spare trust range (bits 9–15, `abi/v7/declared.rs:15-18`), so an unfed trend resolves `Missing` by construction rather than by a producer remembering to flag it.

## Drawing

A bar at the tape's inner edge (x ≈ 90–94), rooted at the vertical center, in `Theme.gps` (the token's stated purpose), clamped to the tape extent; drawn only when IAS and trend both show a value. Missing draws nothing — absence, not a zero-length bar. Any smoothing or quantization must be a deterministic function of the resolved value, never wall-clock animation, or backends diverge and the baselines stop meaning anything.

## Declutter

The unusual-attitude tier removes the speed-tape bands and the turn-rate cue through the one priority table (`pfd.rs:86-93`). A rate cue belongs in that same tier — add it to the table, not as an ad-hoc gate, so the declutter tests keep covering it.

## Guardrails, same PRs

Derivation unit tests mirroring `turn.rs` (dt window bounds, duplicate re-declares, reordered samples ignored, same-stream check); a dyn-style draw test proving the bar clamps while the resolved value does not; corpus entries for accelerating / steady / missing; the trend valid bit asserted through resolve tests; baseline re-pin with its reason.

luofang34:
Closes #51. **Targets `abi/v8-batch`** — the second slice of the coordinated ABI revision (#58).

The speed tape showed the airspeed and nothing about its rate, so an approaching band edge was only visible once the needle reached it. The theme has named a token for this cue since it was written — `gps` is documented as "GPS guidance, trends, and rate cues" — and nothing consumed it.

## The rate is measured, not differenced

`ias_trend` is a wire field on the Dynamics group, NaN-absent, declared by Trust valid bit 9. A display that differenced its own frames would be measuring its own repaint clock, which is the thing DYN-01 exists to keep out of the presentation layer — the same reason turn rate is a declared sample with a declared basis rather than a body rate.

The bar runs from the pointer line to where the airspeed will be six seconds ahead at the current rate, clamped to the tape's ends because past them it would point at a speed the tape is not showing. An absent rate draws nothing: a zero-length bar would say the airspeed is steady, which is a different statement from not knowing. A live rate beside a dashed readout draws nothing either.

It joins the one declutter priority table, beside the speed bands and the turn cue.

## A coverage gap this uncovered

**Every shared canonical state sits at an unusual attitude.** So the speed bands, the turn cue, and now the trend bar are decluttered out of every pinned case — the bands and the turn cue have had no pinned coverage since they were written, and I only found it because the composition digest did not move when it should have.

`level-accelerating` is a new PFD extreme state at a level attitude with a live trend, turn and slip. It is the first pinned case that is not decluttered, and it draws all three. That is why the admission counts move by more than the field alone would explain: 242 to 260 cases, 166 to 196 warnings, the added warnings being the same two overflowing boxes in the one state that now paints them undecluttered.

## Guardrails

- The bar's length is the rate times the look-ahead, and its direction follows the sign.
- An absent rate draws no bar; neither does a live rate beside an unreadable airspeed.
- The new pinned state carries the cue into the raster baseline and the composition digest.

## Pins

The Dynamics payload grows 16 to 20 bytes, so the ABI goldens are regenerated. The PFD emits a new bar in one pinned state, so its raster baseline, the composed-frame hashes, the screen-composition digest and the composition digest move; `release-manifest.json` is regenerated. The CHANGELOG entry states the migration: stamping version 8 without writing the longer Dynamics group has the whole frame refused, not just that group.

ATT-01 evidence re-records in its own commit. Both gates VALID; `fmt`, `clippy -D warnings`, `test --all-targets`, and the release checks are green.


luofang34:
Wrong head branch — this PR was opened from the merge-style docs branch, not from fix/51-airspeed-trend, so it carries none of the trend work. Reopening against the right head.

Issue #52, opened 2026-08-18
True airspeed exists nowhere: not in AirData, not on the glass


luofang34:
`AirData` carries indicated airspeed and the applied altimeter setting, nothing else (`crates/indicate-instrument-state/src/aircraft.rs:29-36`). There is no true-airspeed field and no TAS readout on any panel.

TAS cannot honestly be derived on the display side: it needs density (temperature, pressure) the state does not carry, and the standing rule is that a value the source did not supply renders `Missing`, not a substitute (ADR-0017, cited at `aircraft.rs:27-28`). So this is a wire field: `tas_mps: Option<f32>` on the Air group, NaN-absent like `baro_sel_hpa` (`abi/v7/declared.rs` conventions) — an ABI move, pinned in `release-manifest.json` and named in the CHANGELOG entry that ships it.

## Drawing

A readout box at the head of the speed tape mirroring the groundspeed box at its foot (`readout_box` at (0, 335, 90, 25), `sets/indicate-instrument-panels/src/pfd/tapes.rs:73-87`): `TAS {:.0}kt` at (0, 0, 90, 25). The unit is `kt` — in the glyph vocabulary; `kts` is not (`PANEL_VOCABULARY`, panel-contract.md).

Color: `primary` white. TAS is an air-data value; the GS box wears `gps` magenta because groundspeed is kinematic-derived, and keeping that distinction legible is the reason the two boxes are worth having side by side. Going through `readout_box` buys the status behavior for free — amber/red border under degradation, and the numeral claims the Air group through the attributed-text path.

Layout check before pinning: (0, 0) is currently bare tape backdrop, but the tape's hidden-status fallback text and the IAS source label sit lower on the same column — confirm no collision at any status combination, including the tape-hidden state.

## Guardrails, same PRs

Codec round-trip including the absent field; a draw test that an absent TAS with a live IAS renders an absent/dashed box rather than degrading the tape (a source may supply one Air field and not the other); corpus entries with TAS present and absent; the group-region addition so the new numeral's claim passes admission; baseline re-pin with its reason.

luofang34:
Landed on `main` in #64 (merge commit 46b6e44), which carried the whole state-ABI v8 allocation batch. Closed here rather than automatically: the pull request body listed the issues after a single `Closes` keyword, which only closes the first.

PR #72, opened 2026-08-18
Show where the airspeed is going, not just where it is (#51)

### LINKED ISSUE #51: The speed tape shows no airspeed trend, and no trend quantity exists in state

luofang34:
Closes #51. Targets `abi/v8-batch` — the second slice of the coordinated ABI revision (#58). Replaces #70, which was opened from the wrong branch and carried none of this work.

The speed tape showed the airspeed and nothing about its rate, so an approaching band edge was only visible once the needle reached it. The theme has named a token for this cue since it was written — `gps` is documented as "GPS guidance, trends, and rate cues" — and nothing consumed it.

## The rate is measured, not differenced

`ias_trend` is a wire field on the Dynamics group, NaN-absent, declared by Trust valid bit 9. A display that differenced its own frames would be measuring its own repaint clock, which is what DYN-01 exists to keep out of the presentation layer — the same reason turn rate is a declared sample with a declared basis rather than a body rate.

**This reverses the issue's own recommendation and is yours to ratify.** Issue #51 weighed three options and picked feeder derivation mirroring `turn.rs`. I took the wire field instead: the allocation table already reserved `ias_trend` on Dynamics, a derived rate would need the same dt-window and same-stream guards `turn.rs` carries without the source ever having agreed to supply the quantity, and a feeder can still fill the field later without moving the wire again.

The bar runs from the pointer line to where the airspeed will be six seconds ahead at the current rate, clamped to the tape's ends because past them it would point at a speed the tape is not showing. An absent rate draws nothing: a zero-length bar would say the airspeed is steady, which is a different statement from not knowing. A live rate beside a dashed readout draws nothing either. It joins the one declutter priority table.

## A pre-existing gap this uncovered, and the half of it that stays open

Every pinned state that resolved a flying attitude resolved an **unusual** one, so the turn cue and the trend bar were decluttered out of every pinned case — the turn cue has had no pinned coverage since it was written. I only found it because the composition digest did not move when I expected it to.

`level-accelerating` is a new PFD extreme state at a level attitude with a live trend, turn and slip. It draws both cues.

It does **not** close the speed-bands half. The bands need a configured `v_speeds`, and every pinned path renders the empty configuration by design, so that is a harness-capability question rather than a fixture one. It stays open and should be its own issue.

## What is and is not pinned

The composition digest and the screen-composition digest move. The **raster baselines do not** — each is pinned over the typical state, whose attitude is unusual, so the bar is decluttered out of them.

That leaves the bar's bytes pinned by this crate's own tests rather than across backends. The corpus entry issue #51 asked for needs the real-panel corpus helper that lives on `main`, which this batch branch does not carry yet; it belongs in the change that brings the batch up to `main`, and I would rather say that than add a synthetic fragment that pins an opcode already covered.

## Guardrails

- The bar's length is the rate times the look-ahead, and its direction follows the sign.
- Its tip stops at each end of the tape.
- An absent rate draws no bar; neither does a live rate beside an unreadable airspeed.
- Trust bit 9 gates the trend in both directions.

## Pins

The Dynamics payload grows 16 to 20 bytes, so the ABI goldens are regenerated. `release-manifest.json` is regenerated. The CHANGELOG states the migration: stamping version 8 without writing the longer Dynamics group has the whole frame refused, not just that group.

ATT-01 evidence re-records in its own commit — for a moved count, not a moved source: the bound sources are untouched, so the gate never reddened, and it is the freshness check that sees the stale summaries.

Local gates: fmt, clippy `-D warnings`, `test --all-targets`, both evidence gates VALID, check-structure, check-release-markers, check-release-manifest.


luofang34:
**REQUEST CHANGES** — the trend bar itself is well built and genuinely guarded, but `scripts/check-structure.sh` is red, the Dynamics group's non-finite rule was not extended to the new field the way the sibling Air append extended Air's, and the CHANGELOG contradicts itself about what this release carries.

Reviewed `origin/abi/v8-batch...fix/51-airspeed-trend` (head `e750825`) in a clean worktree.

---

## Blocking

### 1. `scripts/check-structure.sh` fails — two size limits blown

```
$ bash scripts/check-structure.sh ; echo $?
FORBIDDEN: ./sets/indicate-instrument-panels/src/pfd/tapes.rs has 510 lines (limit 500)
FORBIDDEN: ./sets/indicate-instrument-panels/src/pfd/tapes.rs:28 function body has 81 lines (limit 80)
check-structure: FAILED
1
```

`tapes.rs` goes 469 → 510 lines, and `speed_tape` (tapes.rs:28) crosses 80 lines because the `declutter` parameter and the `if !declutter { trend_bar(...) }` block landed on a function that was already at the ceiling. This is in CONTRIBUTING.md's local gate battery, so CI is red as pushed. Splitting the trend bar into `pfd/trend.rs` (or moving the speed tape's ladder loop into a helper) fixes both at once.

### 2. `dynamics_fault` does not gate the new field's finiteness — a non-finite `ias_trend_mps2` leaves turn and slip painting `Valid`

`crates/indicate-instrument-state/src/validate.rs:235`

```rust
if turn_bad || !opt_finite(dynamics.lateral_mps2) {
    return Some(GroupFault::NonFinite);
}
```

`ias_trend_mps2` is not in that condition. NaN is absent-by-convention so it is harmless, but `±inf` decodes to `Some(inf)` through `opt()` and reaches the state.

Reproduction (temporary test in `resolve/dynamics_tests.rs`, run and reverted):

```
ias_trend_mps2 = INFINITY -> integrity.dynamics = None
                             turn = Valid, slip = Valid, trend = Failed
lateral_mps2  = INFINITY -> integrity.dynamics = Some(NonFinite)
                             turn = Failed
```

The same batch's Air append set the precedent in this same file: commit `4c6712e` ("Carry true airspeed from the source to the glass (#52)") added `opt_finite(air.tas_mps)` at validate.rs:157 **and** a case in `validate/tests.rs`. Here neither landed. The trend signal itself fails safe via `resolve::finite`, so this is not a NaN-to-scene escape — it is that a corrupt Dynamics payload no longer condemns its group, and the turn cue and slip ball keep painting on it. Add `opt_finite(dynamics.ias_trend_mps2)` to the condition and extend `non_finite_dynamics_fail_with_typed_reason` to cover the new field (this PR left that test untouched, so there is no guardrail for the field either way).

### 3. CHANGELOG contradicts itself inside one section

`CHANGELOG.md:56` — "This release carries two of them: true airspeed on the Air group, and the airspeed trend on the Dynamics group."
`CHANGELOG.md:73` — "**The Dynamics group grows from 16 bytes to 20.**"
`CHANGELOG.md:80` — "**The batch allocates four group ids and three field appends. This release implements the Air append only.**"

The third statement is now false and was not touched. `scripts/check-release-markers.sh` passes because it only checks the five contract values, so nothing catches this. A consumer reading the section cannot tell whether the Dynamics payload moved.

While that bullet is open: the new one at CHANGELOG.md:75 reads "A producer that stamps version 8 and keeps writing 16 bytes **has the whole frame refused**, not that group." The behaviour is right — `decode_state` returns `Err(GroupTruncated { .. })` and abandons the frame — but the passive-causative construction is not the active voice the project's docs are held to. "The decoder refuses the whole frame, not that group" says the same thing.

---

## Non-blocking

### 4. The ATT-01 priority table gained a row with no behavioral assertion

`sets/indicate-instrument-panels/src/pfd.rs:93` adds "the airspeed trend bar" to the declutter priority table. The table's own named guardrail, `declutter_follows_the_one_priority_table` (`pfd/attitude_tests.rs:117`), asserts `!turn_rate_cue_in(&decluttered)` and `!bands_in(&decluttered)` but says nothing about the bar.

I deleted the `if !declutter` gate (always draw) and re-ran the workspace: **only** the three `descriptors::digest_tests` reddened. Those are opaque byte pins that happen to move because `fully-fed` resolves an unusual attitude with `ias_trend` declared; nothing states the intended behavior. One line beside the two existing assertions closes it.

### 4b. The new flag reached the codec and the resolver but not the group's other two validity consumers

Same root cause as finding 2 — the append did not ripple to every place that enumerates the Dynamics group's validity bits. Three consumers were missed, not one:

- `crates/indicate-instrument-state/src/group_id.rs:229` — `withhold_group(GroupId::Dynamics)` clears `valid.turn` and `valid.slip` and leaves `valid.ias_trend` set, against that function's own doc at group_id.rs:193 ("the validity flags covering the group are cleared"). Invisible today only because the withheld group's data is `None`, so the trend resolves `Missing` regardless — but `withholding_is_idempotent_and_leaves_other_groups_fed` only exercises Air, so nothing would notice if that stopped being true.
- `crates/indicate-instrument-state/src/resolve/group_status.rs:123` — the Dynamics group status folds `state.valid.turn && state.valid.slip`. Kinematics two arms above it folds all three of its flags. A source declaring only the trend valid reports the group `Failed` while the bar paints; one declaring turn and slip but not the trend reports it `Valid`.

### 4c. `unusual-inverted`'s trend rate is dead data

`descriptors/extreme_states.rs:31` adds `ias_trend_mps2: Some(-4.0)` to `pfd_unusual_inverted`, but that state inherits `states::typical()`'s `valid`, where `ias_trend` is false — so the rate never resolves and the state cannot distinguish declutter from an undeclared flag. Harmless, but it reads as coverage it is not. Either set the bit (and let the state prove the declutter row from finding 4) or drop the field.

### 5. No Dynamics-specific wire tests; the CHANGELOG's Dynamics claim is untested

`crates/indicate-instrument-state/src/abi/v8/tests.rs` gained exactly one line (a struct-literal field). The short-payload, append-after-`age_ms`, and longer-payload-counted tests are all written against Air (0x03): `a_known_group_below_its_minimum_length_fails_that_group` (tests.rs:113), `air_decodes_tas_appended_after_the_trailing_age` (tests.rs:146), `an_appended_payload_tail_is_accepted_and_counted` (tests.rs:165).

The mechanism is group-agnostic and the values are pinned indirectly — I reverted `min_len(GroupId::Dynamics)` to 16 and three tests reddened (`fixtures_round_trip_bit_exactly`, `full_frame_is_pinned`, `flight_controller_frame_is_pinned`), and dropping `| (u16::from(v.ias_trend) << 9)` from `encode_trust` reddened the same three. So the layout is held. But CHANGELOG.md:75 makes a group-specific claim — "A producer that stamps version 8 and keeps writing 16 bytes has the whole frame refused" — that no test states for 0x0B. A three-line `GroupTruncated { id: GroupId::Dynamics }` case would make the claim checkable.

### 6. `crates/indicate-instrument-state/src/abi/v8.rs:48` is now stale

> "Landed so far: the Air group carries `tas_mps` after its trailing `age_ms`, so its minimum payload is 16 bytes."

The Dynamics append lands in this PR and is not named. Separately, "Landed so far" is the temporal, other-file-dependent phrasing CLAUDE.md's comment rules ban — this doc has to be edited by every remaining batch member or it lies. Consider pointing at the `group_id` registry table instead of restating it.

### 7. `if height <= 0.0` does not stop a NaN bar

`sets/indicate-instrument-panels/src/pfd/tapes.rs:137`. `f32::clamp` returns NaN for a NaN input, `tip < CENTER_Y` is then false, and `height = NaN` passes `<= 0.0`. Measured, with the signal forced:

```
NaN trend -> Rect { x: 90.0, y: 180.0, w: 4.0, h: NaN }
```

Unreachable through `resolve()` because `resolve::finite` (resolve.rs:176) fails non-finite signals, and `vsi()` has the same shape today — so this is defence-in-depth, not a live defect. But the guard was added specifically to reject degenerate geometry, and `if !(height > 0.0)` costs one character and actually does it.

### 8. Zero and absent draw the same thing, so the stated distinction is not one the display makes

`dynamics.rs:82` and the `trend_bar` doc both justify the missing-stays-missing rule with "a zero-length bar would claim the airspeed is steady, which is a different statement from not knowing." A *valid zero* rate also draws nothing:

```
zero trend  -> []          (Sig::with_status(0.0, Valid))
tiny trend  -> [(90.0, 179.99995, 4.0, 4.6e-5)]
```

Disappearing at zero is standard trend-vector behaviour and I am not asking you to change it — but the comment argues for a distinction the pixels do not carry, and nothing pins the zero case. Either reword to "a bar is drawn only while the rate is non-zero, and an absent rate is not a zero rate" or add the zero case to `an_absent_trend_draws_no_bar`.

### 9. Trust bit 9 is written out of bit order in both codecs

`abi/v8/declared.rs:87` puts `ias_trend: flags & 0x0200` before `velocity_vertical: flags & 0x0100`, and `declared.rs:136` puts `<< 9` before `<< 8`. Every other term in both chains ascends, and the module doc (declared.rs:16) enumerates them ascending. Cosmetic, but an out-of-order chain is where the next append's off-by-one hides.

### 10. The clamp's lower bound is a bare `335.0`

`tapes.rs:132`: `.clamp(SPEED_TAPE_TOP, 335.0)`. The top correctly follows the constant the sibling change introduced — that is what commit `e7e5b26` is for — but `335.0` is a bare literal here and at tapes.rs:41, :49, :99, :207, :428. Naming `SPEED_TAPE_BOTTOM` is the symmetric half of the same fix, and it is the kind of thing that goes wrong exactly once.

### 11. Past-tense narrative in the new extreme state's doc

`descriptors/extreme_states.rs:41` — "Every pinned state that resolved a flying attitude resolved an unusual one, so the turn cue and the trend bar **were** decluttered out of all of them and no pinned case drew either."

CLAUDE.md asks for invariants in present or hypothetical tense, and this sentence is both past-tense refactor narrative and a claim about the contents of other files that #75 and #76 can falsify. Present tense fixes it: "The shared corpus resolves an unusual attitude in every state that has a flying attitude, so the turn cue and the trend bar are decluttered out of all of them." The paragraph below it, about the speed bands having no pinned coverage `v_speeds` can reach, is exactly the right kind of note — keep it.

---

## What I tried to break and could not

- **Pinned coverage of the new paint is real.** I replaced `trend_bar`'s body with an early `return Ok(())`: `the_trend_bar_reaches_the_speed_the_rate_predicts`, `the_trend_bar_stops_at_the_tape_ends`, and all three `digest_tests` went red. `an_absent_trend_draws_no_bar` correctly stayed green. Rendering the pinned state directly confirms which case carries it — `level-accelerating` resolves `unusual=false`, `ias=103.0 Valid`, `trend=2.33 kt/s Valid`, and emits `Rect { x: 90.0, y: 79.23, w: 4.0, h: 100.77 }`. It is the *only* pinned state that draws the bar. The others either fail the trend signal outright — `typical`, `ias-without-tas`, `unusual-inverted`, `readout-extremes` and `director-engaged` all inherit `states::typical()`'s `valid`, where `ias_trend` is false, and `source-unusable` fails on quality — or, in `fully-fed`'s case, carry a resolvable trend into the declutter tier and lose it there. So the coverage rests on the composition digest, not on the raster baselines (which are taken over `typical` and correctly did not move). That is a reasonable place for it, and the new extreme state is exactly the right answer to the "every canonical state is at an unusual attitude" problem.
- **Geometry is pinned, not asserted-into-existence.** `TREND_LOOK_AHEAD_S` 6.0 → 10.0 and widening the clamp to `(0.0, 360.0)` each reddened the two velocity tests plus the digest.
- **The "no bar beside a dashed airspeed" rule is load-bearing.** Deleting `|| !data.ias_kt.status.shows_value()` reddened `an_absent_trend_draws_no_bar` on its own.
- **Wire layout is pinned.** `min_len(Dynamics)` 20 → 16 and dropping bit 9 from `encode_trust` each reddened `fixtures_round_trip_bit_exactly` and both golden-hex fixture tests.
- **Freshness/quality/staleness split behaves.** A stale IAS with a valid trend still draws (`Stale.shows_value()` is true, and the tape is still showing a number, which is what the doc says); a stale trend draws the same bar with no status differentiation — consistent with the turn cue and the VSI needle, which are also flat magenta.

No test stayed green when the thing it claims to cover was removed, except the declutter row in finding 4.

## Pin consistency

Internally consistent — every re-pinned value is what this tree computes:

```
$ cargo run -q -p instrument-bench
scene digest: 5a982aac...0035b2 (matches pin)
screen-composition digest: 74bb892b...07f1b4 (matches pin)
admission: 298 cases pass, 226 counted warnings
rasterized 15 panel x state x canonical frame renders

$ bash scripts/check-release-manifest.sh
check-release-manifest: OK (release-manifest.json is what the generator produces from this tree)
```

298 = ((5 canonical + 4 extreme) × 9 + 7 × 8 + 6 × 2) × 2, and 226 = 7 × 30 + 16 with `nothing-fed` contributing none — the rewritten comment at `admission/tests.rs:26` states the arithmetic correctly. The PFD raster baseline `411ef00f…` correctly did **not** move, because `typical` decluttered the bar away. Both evidence-gate invocations return `VALID`, and the recorded ATT-01 counts (86 / 182 / 98) match what the suites now report. As you note, all three of #72/#75/#76 move the same globals, so whichever lands second recomputes.

## Full battery as run

| Gate | Result |
|---|---|
| `cargo fmt --all --check` | pass |
| `cargo clippy --locked --all-targets --workspace -- -D warnings` | pass |
| `cargo test --locked --workspace --all-targets --no-fail-fast` | pass (21 binaries, 0 failed) |
| `cargo doc` with `-D missing_docs -D rustdoc::broken_intra_doc_links` | pass |
| `cargo build --locked --release --workspace` | pass |
| `cargo check --target thumbv7em-none-eabihf` (12 crates) | pass |
| `scripts/check-structure.sh` | **FAILED** (finding 1) |
| `scripts/check-release-manifest.sh` | pass (+ 8-case selftest) |
| `scripts/check-release-markers.sh` | pass |
| `scripts/check-instrument-requirements.sh` | pass (73 requirements) |
| `scripts/check-certification-claims.sh` | pass |
| `scripts/check-standards-registry.sh` | pass |
| `scripts/trace-report.sh`, `scripts/detect-target.sh` | pass |
| `evidence-gate --resolve-selectors` / `--require-resolvable` | `verdict: VALID` |

luofang34:
All three blocking findings are fixed. Five commits, `f20b810..e41e92d`.

## Finding 1 — `check-structure.sh` was red

Split by tape, not by line count: `tapes.rs` + `tapes/airspeed.rs` + `tapes/altitude.rs` (`foo.rs` + `foo/`, no `mod.rs`, no generic module name).

The seam is the tape because the two tapes share no geometry beyond the strip they run down — each has its own scale, its own bands or bugs, its own boxes. What genuinely is shared stays in `tapes.rs`: the vertical extent both span (`CENTER_Y`, `TAPE_BOTTOM`) and the pointed readout each hangs at the pointer line (`PointedBox`, `IAS_READOUT`, `ALT_READOUT`, `pointed_readout`, `fitted_text_size`). The VSI travels with the altitude tape it reads against.

The function limit needed its own fix — splitting the file does nothing for an 81-line body. `speed_tape` drew both the TAS box and the groundspeed box, and only one of them had a name; the other is now `gs_box`, mirroring `tas_box`. The body is 66 lines by the script's own counter.

`bash scripts/check-structure.sh` → `check-structure: OK`.

## Finding 2 — `dynamics_fault` never gated the new field

Fixed in `validate.rs`: the finiteness check is now `turn_bad || !opt_finite(dynamics.lateral_mps2) || !opt_finite(dynamics.ias_trend_mps2)`, beside the other two rather than in the trend's own resolver — a group fault is the whole group's.

Two sibling gates had the same hole and are closed in the same commit:

- `group_status` folded `valid.turn && valid.slip` and not `valid.ias_trend`, so the group-level surface could report better than the signals under it — the one thing that surface documents it cannot do.
- `withhold_group` cleared `valid.turn` and `valid.slip` and not `valid.ias_trend`, so a state with Dynamics withheld still declared the trend valid.

**Mutation proof.** `non_finite_dynamics_fail_with_typed_reason` now walks every member of the group (turn rate, lateral force, airspeed trend) against NaN, +inf, and -inf, and asserts all three cues stop showing a value — the cross-signal effect you named, not just the integrity field.

Mutation (a): deleted `|| !opt_finite(dynamics.ias_trend_mps2)` from `dynamics_fault`.

```
thread '...non_finite_dynamics_fail_with_typed_reason' panicked at
crates/indicate-instrument-state/src/resolve/dynamics_tests.rs:160:9:
assertion `left == right` failed: airspeed trend = NaN
  left: None
 right: Some(NonFinite)
```

Mutation (b): same deletion, with the loop narrowed to your exact case, `f32::INFINITY`:

```
assertion `left == right` failed: airspeed trend = inf
  left: None
 right: Some(NonFinite)
```

Both restored (`\cp -f` from a saved copy, verified by re-grepping the line); test green.

The shape of the test is the guardrail as much as the assertion: a test that only ever moved the lateral force is what let the trend through, so it now loops over the members and a member added without being added here leaves the loop visibly short of the struct.

## Finding 3 — `CHANGELOG.md` contradicted itself

Deleted the false clause rather than renumbering it. The bullet now reads:

> - **The batch allocates four group ids and three field appends.** The ids are 0x12 BearingPointers (stamped, #53), 0x13 AirframeConfig (stamped, #57), …

Read back: parses, no fragment.

One residual I did not take, flagged for your call: the sentence three lines above the table — "This release carries two of them: true airspeed on the Air group, and the airspeed trend on the Dynamics group" — is the same class of running tally and will need editing on the next append that lands. It is currently true, and the per-group bullets already say which appends are in, so it is a candidate for deletion too; I left it rather than widen the finding unasked.

## Non-blocking items — took all six

- **Declutter test.** `declutter_follows_the_one_priority_table` now asserts the trend bar in both directions. Both states carry a live trend rate first, so the bar's absence under declutter is the tier removing it and not a signal that never arrived — a bar missing because the signal is Missing proves nothing. Mutation proof: removed the `!declutter` guard around `trend_bar` → red at `assertion failed: !trend_bar_in(&decluttered)`.
- **NaN and the `height <= 0.0` guard.** NaN fails every ordering comparison, so the guard said "not non-positive" and let a rect out with an extent no backend can paint. Now `!height.is_finite() || height <= 0.0`. The resolver keeps NaN out today; the guard makes that a property of the panel rather than of its caller.
- **Bit 9 before bit 8 in both trust codecs.** Reordered. The value is unaffected — the flags are OR-ed — but the sequence is the only thing that makes the codec readable against the bit list documented above it.
- **`withhold_group` / `group_status`.** Both fold the new flag; see Finding 2.
- **Bare `335.0`.** It is the foot both tapes end at and both hang their readout boxes from, and it appeared as a bare literal in a clamp, a min, and two box anchors. Named once as `TAPE_BOTTOM` in the shared parent module.
- **Past-tense narrative in the extreme state's doc comment.** Rewritten in present tense, same fact.

## Evidence

Extending `attitude_tests.rs` moves the blob `RESULT-PFD` binds, so ATT-01 is re-recorded in its own commit against `c42b8f4`: `config-digest`, `run-id`, `CFG-WORKTREE` commit and tree, and all three `output-digest`s (the header embeds the commit). Test counts did not move in any of the three suites — the change extends existing tests rather than adding any. The presentation and raster `source-digest`s are unchanged, and that was verified by recomputing `git hash-object` on both locators, not assumed.

`evidence-gate --resolve-selectors --require-resolvable` → `verdict: VALID` (and VALID for each invocation separately).

## CI, as run locally

- `cargo fmt --all --check` — clean.
- `cargo clippy --locked --all-targets --workspace -- -D warnings` — clean.
- `cargo test --locked --workspace --all-targets --no-fail-fast` — 21 suites, 0 failed. Panels 86, state 182, raster 98, conformance 24, evidence 22+9+1+50.
- `RUSTDOCFLAGS="-D missing_docs -D rustdoc::broken_intra_doc_links" cargo doc --locked --workspace --no-deps` — clean.
- `cargo build --locked --release --workspace` — clean.
- Every `scripts/check-*.sh` — OK: structure, certification-claims, instrument-requirements (73), release-manifest, release-markers (0.5.0; ABI 8, scene 1, corpus 4), standards-registry.
- `cargo check --target thumbv7em-none-eabihf` over the kernel set — clean.
- `cargo run -q -p instrument-bench` — scene digest and screen-composition digest both match their pins; admission 298 cases / 226 warnings, unchanged.

No pinned value moved, so nothing was re-pinned.

luofang34:
**REQUEST CHANGES (narrow)** — all three blocking findings are closed and I verified each by breaking it again. What remains is two one-assertion tests: the `group_status` and `withhold_group` gates in `c3a440d` were changed without guardrails, and I confirmed that by reverting each and watching the whole workspace stay green. Separately, `abi/v8-batch` has moved under this branch, so every global pin needs recomputing before merge.

Re-reviewed `f20b810..e41e92d` (head `e41e92d`) in a clean worktree, gut-testing each claim rather than reading the diff for it.

---

## Per finding

| # | Round-1 finding | Status |
|---|---|---|
| 1 | `check-structure.sh` red | **Closed**, verified |
| 2 | `dynamics_fault` did not gate `ias_trend` finiteness | **Closed**, verified with a guardrail that reddens |
| 3 | CHANGELOG self-contradiction | **Closed**, verified |
| 4 | Declutter priority table had no assertion | **Closed**, verified |
| 4b | `group_status` / `withhold_group` missed the new flag | Behaviour closed, **untested** — see below |
| 4c | Dead `ias_trend_mps2` in `unusual-inverted` | Not addressed |
| 5 | No Dynamics-specific wire tests | Not addressed |
| 6 | `v8.rs` "Landed so far" stale | Not addressed |
| 7 | NaN passed the `height <= 0.0` guard | **Closed** |
| 8 | Zero-vs-absent comment claims a distinction the pixels do not | Not addressed |
| 9 | Trust bit 9 written before bit 8 | **Closed** |
| 10 | Bare `335.0` clamp literal | **Closed** |
| 11 | Past-tense narrative in the new extreme state's doc | **Closed** |

### 1 — closed

```
$ bash scripts/check-structure.sh ; echo $?
check-structure: OK
0
```

`tapes.rs` 510 → 131, `tapes/airspeed.rs` 217, `tapes/altitude.rs` 208. `speed_tape`'s body is 60 lines. `gs_box` is the right extraction — it mirrors `tas_box` and it is what took the function under the wall without moving a pixel.

### 2 — closed, and the guardrail is real

I deleted `|| !opt_finite(dynamics.ias_trend_mps2)` from `validate.rs:235` and narrowed the new test's loop to `[f32::INFINITY]` so it could only fail on my original case:

```
panicked at crates/indicate-instrument-state/src/resolve/dynamics_tests.rs:160:9:
assertion `left == right` failed: airspeed trend = inf
  left: None
 right: Some(NonFinite)
```

That is exactly the reproduction from the round-1 review, now red. The rewritten `non_finite_dynamics_fail_with_typed_reason` is better than what I asked for: it walks all three members against NaN, `+inf` and `-inf` and asserts the group fault *and* all three cues stop showing a value, so the next member appended to this group cannot repeat the omission. The doc comment's note that NaN is the wire's absent convention and `±inf` is the case that actually decodes to `Some(non-finite)` is correct and is the reason this test earns its length.

### 3 — closed

`CHANGELOG.md:56` states the scope once ("This release carries two of them"), and the allocation bullet at line 80 no longer claims scope of its own. Nothing else in the section states a count. `check-release-markers.sh: OK`.

### 4 — closed

Removing the `if !declutter` gate from `tapes/airspeed.rs:82` now reddens the named guardrail, not just the digest:

```
test pfd::attitude_tests::declutter_follows_the_one_priority_table ... FAILED
```

Giving both `oriented()` states a live trend rate first is the right shape — without it, `!trend_bar_in(&decluttered)` would pass on a signal that never arrived. The added `assert!(trend_bar_in(&normal), …)` is what makes the negative assertion mean something.

---

## Blocking: 4b is closed in behaviour and untested

Both sibling gates in `c3a440d` are correct. Neither is held down. I reverted each on its own and ran `cargo test --locked --workspace --all-targets --no-fail-fast`:

- Reverting `resolve/group_status.rs:123` to `state.valid.turn && state.valid.slip` — **entire workspace green.**
- Reverting `group_id.rs:233`, dropping `out.valid.ias_trend = false` — **entire workspace green.**

This is the same shape as round-1 blocker 2, and CONTRIBUTING.md states the rule directly: a fix without a guardrail is temporary. `withholding_a_stamped_group_resolves_missing` cannot see either change, because the withheld group's data is `None` and the trend resolves `Missing` from that alone.

Two assertions close it. For `withhold_group`, in `group_id/tests.rs` beside the existing idempotence test:

```rust
assert!(!withhold_group(&fixtures::full(), GroupId::Dynamics).valid.ias_trend);
```

For the group fold, the reproduction is this probe, which I ran against the current tree:

```
trendless source: dynamics group = Failed, turn = Valid, slip = Valid, trend = Missing
declaring the bit: dynamics group = Valid, trend = Missing
```

A source that supplies turn and slip, declares both, and has no trend sensor at all now reports the Dynamics group `Failed` while both cues it does feed read `Valid`. I want to be clear that I do **not** think this is the wrong direction: `lateral_mps2` is already `Option` and `valid.slip` was already folded, so a slip-less source reported `Failed` before this PR too, and `Kinematics` folds all three of its flags with a `ValidFlags` doc that explicitly anticipates partial declaration. The fold follows the group's own convention. But `GroupStatuses` is `pub use`d, the behaviour moved for every producer that omits the trend, and nothing states it — which is precisely why it needs the test rather than a comment.

---

## Base has moved under this branch

`origin/abi/v8-batch` is now `bcc00a9` — #76 (BearingPointers 0x12) merged after this branch last took its base, and it is not an ancestor of `e41e92d`. Every global value this PR re-pins has moved on the base:

| Pin | This branch | `abi/v8-batch` |
|---|---|---|
| Composition digest | `5a982aac…` | `b15a5d86…` |
| Screen-composition digest | `74bb892b…` | `59d8086a…` |
| Admission cases | 298 | 312 |
| Admission warnings | 226 | 196 |

The branch is internally consistent on its own tree — `instrument-bench` reports both digests and `298 cases / 226 warnings` as matching their pins, and `check-release-manifest.sh` is OK — so nothing here is wrong, but it cannot merge without taking the base and recomputing all four values plus `release-manifest.json` and the CHANGELOG table. Note the warnings count is not additive: the base is still at 196 because #76 added no overflowing readout, while this branch's +30 comes from the fourth PFD extreme state, so the merged figure is a fresh measurement, not a sum.

---

## The file split: right seam, no redo coming

Judging this separately because it was not the original author's work. **It is the right seam and I do not expect it to move again.**

- The parent keeps exactly what both tapes share and nothing else: `CENTER_Y`, `TAPE_BOTTOM`, `PointedBox` with both readout geometries, `pointed_readout`, `fitted_text_size`. `PX_PER_KT`/`SPEED_TAPE_TOP` went to airspeed, `PX_PER_FT` to altitude — correctly, since neither tape's scale is shared. `fitted_label_size` is used only by `tas_box`, and it stayed with it rather than being hoisted "just in case".
- `foo.rs` + `foo/` matches the convention; there is no `mod.rs`.
- The re-exports (`pub use airspeed::speed_tape; pub use altitude::{altitude_tape, vsi};`) mean zero call-site churn in `pfd.rs`, which is why the digests did not move.
- `gs_box` is kinematic data in the airspeed module, which looks odd until you notice the seam is geometric, not by data source — the box hangs off the speed tape's foot at `TAPE_BOTTOM` and moves when the tape moves. Same for `vsi` sitting in `altitude.rs`.
- The only future pressure I can see is the VSI outgrowing `altitude.rs` if it gains a trend arrow or a selected-VS bug. That is a third module added beside these two, not a re-cut of this seam. `airspeed.rs` at 217 and `altitude.rs` at 208 both have room.

Extracting `TAPE_BOTTOM` also closed finding 10 for free, and the split is what made the NaN fix (finding 7) legible:

```rust
// A not-a-number rate fails every ordering comparison, so a bare
// length test passes it straight through to a rect no backend can
// paint. Finiteness is the first question, length the second.
if !height.is_finite() || height <= 0.0 {
```

That comment says why, not what, and it is the right why.

---

## Still open, non-blocking

- **4c.** `descriptors/extreme_states.rs:31` still gives `pfd_unusual_inverted` an `ias_trend_mps2: Some(-4.0)` it never declares, so the rate is dead data.
- **5.** `abi/v8/tests.rs` is untouched this round. `CHANGELOG.md:75` still makes a Dynamics-specific claim — a producer writing 16 bytes has the frame refused — that only the Air group has a test for.
- **6.** `abi/v8.rs:48` still reads "Landed so far: the Air group carries `tas_mps`…". Now that the CHANGELOG says two appends landed, this module doc is the last place in the tree claiming one. Same defect as blocker 3, one file over.
- **8.** The zero-vs-absent rationale in `tapes/airspeed.rs:100` and `dynamics.rs:82` is unchanged: a valid zero rate draws nothing, same as an absent one, so the comment still argues a distinction the display does not make.
- **STE.** `CHANGELOG.md:75` "has the whole frame refused" is still passive-causative; "the decoder refuses the whole frame" says it actively.

---

## Battery, this revision

| Gate | Result |
|---|---|
| `cargo fmt --all --check` | pass |
| `cargo clippy --locked --all-targets --workspace -- -D warnings` | pass |
| `cargo test --locked --workspace --all-targets --no-fail-fast` | pass (21 binaries, 0 failed) |
| `cargo doc` with `-D missing_docs -D rustdoc::broken_intra_doc_links` | pass |
| `cargo build --locked --release --workspace` | pass |
| `cargo check --target thumbv7em-none-eabihf` (12 crates) | pass |
| `scripts/check-structure.sh` | **OK** (was FAILED) |
| `scripts/check-release-manifest.sh` | pass (+ 8-case selftest) |
| `scripts/check-release-markers.sh` | pass |
| `scripts/check-instrument-requirements.sh` | pass (73 requirements) |
| `scripts/check-certification-claims.sh` | pass |
| `scripts/check-standards-registry.sh` | pass |
| `scripts/detect-target.sh` | pass |
| `evidence-gate --resolve-selectors` / `--require-resolvable` | `verdict: VALID` |
| `cargo run -p instrument-bench` | both digests match pins; 298 / 226 |

**Coverage survived the refactor.** Replacing `trend_bar`'s body with an early return still reddens the two velocity tests, the new declutter assertion, and all three `digest_tests` — so `level-accelerating` is still the pinned state carrying the bar after the split.

**Evidence re-record is internally consistent.** `config-digest`/`run-id` `c42b8f4e…` is the commit before the evidence commit; `tree a074cc69…` matches `git rev-parse c42b8f4e…^{tree}`; the PFD `source-digest git-blob:e0b131aa…` matches `git hash-object` of the new `attitude_tests.rs`. The recorded counts 86 / 182 / 98 are exactly what the three suites report on this tree.

luofang34:
The narrow blocker is closed in `72f011f`, and the branch is now merged up to the current batch head (`96f7a51`) with every pin recomputed.

**The two sibling gates now have guardrails.** You were right that both could be reverted with the suite green, and right that the fold semantics themselves are defensible — so the fix is two assertions, not a behaviour change:

- `withholding_dynamics_clears_every_bit_the_group_covers` asserts the fixture declares all three bits before withholding, then asserts all three are cleared. Verified it reddens when `out.valid.ias_trend = false` is dropped.
- `a_cleared_dynamics_bit_takes_the_group_status_with_it` clears each bit in turn and asserts the group status stops being Valid. Verified it reddens when the fold is narrowed back to `turn && slip`.

The commit message states the exported-surface consequence you flagged — a source that declares turn and slip but not the trend now reports the group Failed — and says why that is the same treatment `valid.slip` already gets over an equally optional `lateral_mps2`. That was the part with no statement anywhere, which was your actual objection.

**On the base merge you flagged:** done. #76, #75 and #74 have all landed on `abi/v8-batch` since your review, and this branch now merges `0ff7582`. Re-pinned by running the tests, not by resolving conflicts with a strategy flag: composition digest `b9494f30`, screen-composition digest `2a787ce5`, admission cases 330 (the batch's 312 plus this branch's fourth PFD extreme state, drawn quiet and saturated against nine withholding cases), the v8 golden frames, `release-manifest.json`, and the CHANGELOG table.

Worth recording, since it is exactly the failure mode you have caught twice: taking the batch head's `descriptors.rs` wholesale during conflict resolution silently dropped this branch's own `level-accelerating` extreme-state registration. The admission count came out at 312 instead of 330, which is what caught it — the ratchet earned its keep. It is restored, and the trend bar's pinned coverage with it.

**On the split judgment** — thank you for answering it directly. That was the thing I could not assess from outside, and "the only future pressure is a third module beside these, not a re-cut" is the answer I wanted.

Full local CI green: fmt, clippy `-D warnings`, `test --all-targets --no-fail-fast`, doc gate, release build, every `check-*.sh`, and the evidence gate `VALID` against re-recorded artifacts.

PR #73, opened 2026-08-18
Track the documentation language rules the repository is written to

luofang34:
`CLAUDE.md` tells every contributor that project-authored documentation uses ASD-STE100 Simplified Technical English and points at `AGENTS.md` for the rules. Neither file was tracked.

The consequence is not theoretical. Every review in this batch has held new prose to those rules, and the review of the airspeed-trend work noted it could not find the authority in the repository or in any worktree, and judged against the standard generally instead. A contributor cloning this repository could not read the rule they were being held to.

Both files are tracked now, and `CONTRIBUTING.md` names them in its own section, where a contributor looks before pushing.

Split out of #68, where a `git add -A` had swept them into an unrelated change to the structure gate. A policy document should not ride in with, or be reverted alongside, a `find` prune.

No gate ships with this and none should. Sentence length and word choice are review's judgement; an approximation of them would either wave through bad prose or block good prose while claiming an authority it does not have. The rules say plainly that they bind new and changed text rather than legacy prose, so there is nothing here for CI to ratchet.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,20 @@
+# Indicate Repository Instructions
+
+## Documentation language
+
+All project-authored documentation MUST use ASD-STE100 Simplified Technical
+English.
+
+- Write in English.
+- Use an approved word when it has the required meaning.
+- Use one word for one meaning.
+- Use short, direct sentences.
+- Use active voice when you can identify the actor.
+- Give one instruction or one main statement in each sentence.
+- Define a necessary technical name before you use it.
+- Use the same technical name in all related documents and diagrams.
+- Keep code identifiers, protocol fields, quotations, and product names exact.
+- Apply these rules to all new text and to all text that you change.
+
+Do not copy noncompliant legacy text into a new document. Record a broad legacy
+rewrite as separate work.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,6 +5,10 @@
 All project-authored documentation MUST use ASD-STE100 Simplified Technical
 English.
 
+The rules below are the binding subset. Apply them without consulting the
+standard. The ASD-STE100 dictionary is advisory here: a reviewer cites a rule
+from this list, not a word from a dictionary neither party has.
+
 - Write in English.
 - Use an approved word when it has the required meaning.
 - Use one word for one meaning.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -11,11 +11,13 @@ against ASD-STE100.
 
 - Write in English.
 - Use one word for one meaning.
-- Use short, direct sentences.
+- Keep a sentence under 20 words in a procedure and under 25 words in a
+  description. The standard this list is modelled on carries those
+  counts, and a rule with a number is one a reviewer can adjudicate.
 - Use active voice when you can identify the actor.
 - Give one instruction or one main statement in each sentence.
 - Define a necessary technical name before you use it.
-- Use the same technical name in all related documents and diagrams.
+- Use the same technical name across the documents a change touches.
 - Keep code identifiers, protocol fields, quotations, and product names exact.
 - Apply these rules to all new text and to all text that you change.
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2,15 +2,14 @@
 
 ## Documentation language
 
-All project-authored documentation MUST use ASD-STE100 Simplified Technical
-English.
-
-The rules below are the binding subset. Apply them without consulting the
-standard. The ASD-STE100 dictionary is advisory here: a reviewer cites a rule
-from this list, not a word from a dictionary neither party has.
+Project-authored documentation follows the rule list below. The list is
+modelled on ASD-STE100 Simplified Technical English, and it is not that
+standard: the repository does not distribute the specification or its
+dictionary, so a reviewer cites a rule from this list and nothing else.
+A document that follows every rule here makes no conformance claim
+against ASD-STE100.
 
 - Write in English.
-- Use an approved word when it has the required meaning.
 - Use one word for one meaning.
 - Use short, direct sentences.
 - Use active voice when you can identify the actor.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,4 @@
+# Indicate Repository Instructions
+
+Write all project-authored documentation in ASD-STE100 Simplified Technical
+English. Apply the full documentation language rules in `AGENTS.md`.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,4 +1,5 @@
 # Indicate Repository Instructions
 
-Write all project-authored documentation in ASD-STE100 Simplified Technical
-English. Apply the full documentation language rules in `AGENTS.md`.
+Write all project-authored documentation to the language rules in
+`AGENTS.md`. Those rules are modelled on ASD-STE100 Simplified Technical
+English; following them is not a conformance claim against it.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -127,9 +127,9 @@ the tag is what a human owes it.
 ## Documentation language
 
 Project-authored documentation uses ASD-STE100 Simplified Technical
-English. `AGENTS.md` at the repository root holds the rules, and they
-apply to new text and to text you change, not to legacy prose you happen
-to be near. No gate checks this; review does.
+English. `AGENTS.md` at the repository root holds the rules. They apply
+to new text and to text you change. They do not apply to legacy prose
+you happen to be near. No gate checks the prose itself. Review does.
 
 ## Discipline that is easy to miss
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -124,6 +124,13 @@ entry is open, so a CI check for the tag would fail every release on the
 one run that matters. The changelog entry is what CI can hold honest;
 the tag is what a human owes it.
 
+## Documentation language
+
+Project-authored documentation uses ASD-STE100 Simplified Technical
+English. `AGENTS.md` at the repository root holds the rules, and they
+apply to new text and to text you change, not to legacy prose you happen
+to be near. No gate checks this; review does.
+
 ## Discipline that is easy to miss
 
 - REN-03 frame hashes and the scene digest are pinned invariants, not
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -126,10 +126,13 @@ the tag is what a human owes it.
 
 ## Documentation language
 
-Project-authored documentation uses ASD-STE100 Simplified Technical
-English. `AGENTS.md` at the repository root holds the rules. They apply
-to new text and to text you change. They do not apply to legacy prose
-you happen to be near. No gate checks the prose itself. Review does.
+Project-authored documentation follows the rule list in `AGENTS.md` at
+the repository root. The list is modelled on ASD-STE100 Simplified
+Technical English, and following it is not a conformance claim against
+that standard: this repository does not distribute the specification or
+its dictionary. The rules apply to new text and to text you change.
+They do not apply to legacy prose you happen to be near. No gate checks
+the prose itself. Review does, citing a rule from the list.
 
 ## Discipline that is easy to miss
 
--- a/scripts/check-structure.sh
+++ b/scripts/check-structure.sh
@@ -184,19 +184,17 @@ check_safety_constant_count() {
     fi
 }
 
-# A document that cites a rule nobody can read is the condition this
-# check exists to end: `CLAUDE.md` named ASD-STE100 and pointed at
-# `AGENTS.md` for years while neither file was in the repository, so
-# reviews held contributors to a standard a clone did not contain. The
-# check is not over prose — sentence quality is review's judgement. It is
-# over the pointer, which is mechanical, and which is what actually
-# broke.
+# A document that cites a rule nobody can read is not a rule: a reviewer
+# who holds a contributor to a standard the clone does not contain is
+# citing something the contributor cannot obey. The check is not over
+# prose — sentence quality is review's judgement. It is over the
+# pointer, which is mechanical.
 check_root_document_pointers() {
     local doc target
     for doc in ./CLAUDE.md ./CONTRIBUTING.md; do
         [ -f "$doc" ] || continue
         # Every backticked root-level markdown file the document names.
-        for target in $(sed -n 's/.*`\([A-Za-z0-9_-]*\.md\)`.*/\1/p' "$doc" | sort -u); do
+        for target in $(grep -oE '`[A-Za-z0-9_-]*\.md`' "$doc" | tr -d '`' | sort -u); do
             if [ ! -s "./$target" ]; then
                 echo "FORBIDDEN: $doc points at \`$target\`, which is missing or empty; a cited rule a clone does not contain is not a rule" >&2
                 status=1
--- a/scripts/check-structure.sh
+++ b/scripts/check-structure.sh
@@ -184,6 +184,27 @@ check_safety_constant_count() {
     fi
 }
 
+# A document that cites a rule nobody can read is the condition this
+# check exists to end: `CLAUDE.md` named ASD-STE100 and pointed at
+# `AGENTS.md` for years while neither file was in the repository, so
+# reviews held contributors to a standard a clone did not contain. The
+# check is not over prose — sentence quality is review's judgement. It is
+# over the pointer, which is mechanical, and which is what actually
+# broke.
+check_root_document_pointers() {
+    local doc target
+    for doc in ./CLAUDE.md ./CONTRIBUTING.md; do
+        [ -f "$doc" ] || continue
+        # Every backticked root-level markdown file the document names.
+        for target in $(sed -n 's/.*`\([A-Za-z0-9_-]*\.md\)`.*/\1/p' "$doc" | sort -u); do
+            if [ ! -s "./$target" ]; then
+                echo "FORBIDDEN: $doc points at \`$target\`, which is missing or empty; a cited rule a clone does not contain is not a rule" >&2
+                status=1
+            fi
+        done
+    done
+}
+
 # The family is named after this repository, not after a consumer of it.
 # This checks NAMES ONLY — crate directories and package names. Values
 # that are hashed or pinned downstream keep whatever string they were
@@ -369,6 +390,7 @@ check_file_length
 check_function_length
 check_safety_palette_aliases
 check_safety_constant_count
+check_root_document_pointers
 check_crate_naming
 check_tier_law
 check_crate_map
--- a/scripts/test-structure.sh
+++ b/scripts/test-structure.sh
@@ -19,12 +19,21 @@ probe_dir="sets/indicate-tier-probe"
 probe_doc="docs/instruments/zz-structure-probe.md"
 lock_backup="$(mktemp)"
 cp Cargo.lock "$lock_backup"
+# Every tracked file this script edits is restored by the trap, not by
+# the case that edited it. A case that cleans up after itself leaves the
+# repository dirty the moment it aborts, and `CONTRIBUTING.md` carrying
+# an invented citation makes `check-structure.sh` fail against a
+# document the contributor never wrote.
+citation_backup="$(mktemp)"
+cp CONTRIBUTING.md "$citation_backup"
 
 cleanup() {
     rm -rf "$probe_dir"
     rm -f "$probe_doc"
     cp "$lock_backup" Cargo.lock
     rm -f "$lock_backup"
+    cp "$citation_backup" CONTRIBUTING.md
+    rm -f "$citation_backup"
 }
 trap cleanup EXIT
 
@@ -148,8 +157,6 @@ expect_term_refusal "Swift SceneKit backend" "Swift SceneKit backend"
 # citation per line, so a document that named a missing file before a
 # present one passed — which is the spelling a contributor would most
 # easily write by accident.
-citation_backup="$(mktemp)"
-cp CONTRIBUTING.md "$citation_backup"
 for line in 'It cites `GHOST.md` and then `AGENTS.md`.' \
     'It cites `AGENTS.md` and then `GHOST.md`.'; do
     printf '%s\n' "$line" >> CONTRIBUTING.md
@@ -162,7 +169,6 @@ for line in 'It cites `GHOST.md` and then `AGENTS.md`.' \
     fi
     cp "$citation_backup" CONTRIBUTING.md
 done
-rm -f "$citation_backup"
 
 if [ "$failed" -ne 0 ]; then
     echo "structure-selftest: FAILED ($failed of $((passed + failed)) cases)" >&2
--- a/scripts/test-structure.sh
+++ b/scripts/test-structure.sh
@@ -143,6 +143,27 @@ expect_term_refusal "InstrumentSceneKit" "InstrumentSceneKit"
 expect_term_refusal "IndicateAppleDisplay" "IndicateAppleDisplay"
 expect_term_refusal "Swift SceneKit backend" "Swift SceneKit backend"
 
+# A citation the clone cannot resolve must be refused wherever it sits
+# on the line. The first version of this check read only the last
+# citation per line, so a document that named a missing file before a
+# present one passed — which is the spelling a contributor would most
+# easily write by accident.
+citation_backup="$(mktemp)"
+cp CONTRIBUTING.md "$citation_backup"
+for line in 'It cites `GHOST.md` and then `AGENTS.md`.' \
+    'It cites `AGENTS.md` and then `GHOST.md`.'; do
+    printf '%s\n' "$line" >> CONTRIBUTING.md
+    if INDICATE_STRUCTURE_SELFTEST_CHILD=1 bash scripts/check-structure.sh >/dev/null 2>&1; then
+        echo "REGRESSION: a citation of a missing document was accepted: $line" >&2
+        failed=$((failed + 1))
+    else
+        echo "ok: a citation of a missing document refused"
+        passed=$((passed + 1))
+    fi
+    cp "$citation_backup" CONTRIBUTING.md
+done
+rm -f "$citation_backup"
+
 if [ "$failed" -ne 0 ]; then
     echo "structure-selftest: FAILED ($failed of $((passed + failed)) cases)" >&2
     exit 1


luofang34:
**REQUEST CHANGES** — the pointer check does refuse the case it was built for (I deleted `AGENTS.md` and it went red, twice), but it fails open on a second citation in the same line, it has no selftest in a file that carries a 12-case one, and the comment justifying it states a repository history that did not happen.

Reviewed at `23d4782`, worktree detached on the PR head. Full battery below.

---

## Blocking

### 1. `scripts/check-structure.sh:187-193` asserts a history this repository does not have.

> `CLAUDE.md` named ASD-STE100 and pointed at `AGENTS.md` **for years** while neither file was in the repository

Three checks against the tree, all disagreeing:

```console
$ git log --reverse --format='%ad %s' --date=short | head -1
2026-07-05 Initialize repository                       # six weeks ago

$ git log --oneline --all -- CLAUDE.md AGENTS.md
23d4782 Check that a cited rule is in the repository   # this PR
87f9172 Track the documentation language rules …      # this PR

$ grep -rn 'STE100' . --exclude-dir=target --exclude-dir=.git
CONTRIBUTING.md:129 / AGENTS.md:5,9 / CLAUDE.md:3 / scripts/check-structure.sh:188
```

`CLAUDE.md` has never been tracked here, `ASD-STE100` has never appeared here before this branch, and the repository is six weeks old. Whatever this describes — an untracked local file, another repository, a different project — it did not happen in the history a clone can read, which is the only history a comment in this tree can refer to.

This is not a style point in this repository. Every script in `scripts/` closes by stating precisely what it did and did not prove (`check-standards-registry: external freshness … is NOT proven by CI`; `check-recorded-counts: freshness only`); `standards-registry.toml` records a `verified_on` date for each claim; the evidence gate exists so that a record cannot say more than it can support. A guard whose stated reason for existing is unverifiable narrative undercuts that. It is also the banned form on its own terms — refactor history in a docstring, with a temporal marker ("for years") whose truth depends on nothing in the file.

The check stands up fine without it. Something like: *the repository documents a writing standard and names the file holding it; a named file that a clone does not contain is not a rule. This checks the pointer, not the prose — sentence quality is review's judgement.* Put the origin story in the commit message, where it has a date.

### 2. `scripts/check-structure.sh:199` — only the last citation on a line is checked.

`sed -n 's/.*`\([A-Za-z0-9_-]*\.md\)`.*/\1/p'` opens with a greedy `.*`, so it captures the final match on each line and drops every earlier one. Order alone decides whether a dead pointer is seen:

```console
$ printf 'See `AGENTS.md` and `GHOST.md` for details.\n' >> CLAUDE.md
$ bash scripts/check-structure.sh
FORBIDDEN: ./CLAUDE.md points at `GHOST.md`, which is missing or empty; …
check-structure: FAILED

$ printf 'See `GHOST.md` and `AGENTS.md` for details.\n' >> CLAUDE.md   # swapped
$ bash scripts/check-structure.sh
check-structure: OK
```

Same two citations, same missing file, opposite verdicts. Nothing about a document says one root-level `.md` reference per line, and a reviewer rewording `AGENTS.md at the repository root holds the rules` into a sentence that also names `CONTRIBUTING.md` silently removes coverage.

The correct idiom is already in this file, 168 lines further down at `:367`:

```sh
for target in $(grep -oE '`[A-Za-z0-9_-]+\.md`' "$doc" | tr -d '`' | sort -u); do
```

```console
$ grep -oE '`[A-Za-z0-9_-]+\.md`' probe.md | tr -d '`' | sort -u
AGENTS.md
GHOST.md
```

Given the history this repository has with text scanners that pass by accident — the tier-law selftest exists because an earlier text scanner accepted six spellings `cargo` built happily — a completeness check that misses by word order should not land.

### 3. No selftest case, in the one file that runs a selftest on every invocation.

`scripts/test-structure.sh` is untouched, and `check-structure.sh` still reports `structure-selftest: OK (12 cases)` — twelve tier-law and backend-term cases, none for the new check. The file's own comment at `:399-400` is the argument against this PR:

> The tier law is the one check here that a manifest can spell its way around, so it carries a selftest proving it refuses each spelling.

Finding 2 is a spelling that gets around the new check. It is now the second check here that can be spelled around, and it is the one without a selftest. Three cases cover it and cost nothing — the harness in `test-structure.sh` already plants and restores files with a trap: missing target refused, empty target refused, second-citation-on-one-line refused. Two of those I ran by hand and they pass today; the third is finding 2. CONTRIBUTING.md:59-61 requires the guardrail in the same PR.

### 4. `AGENTS.md:5-10` claims conformance to a standard it tells the reader not to consult, and pins no revision.

```
5:  All project-authored documentation MUST use ASD-STE100 Simplified Technical English.
8:  The rules below are the binding subset. Apply them without consulting the
9:  standard. The ASD-STE100 dictionary is advisory here: a reviewer cites a rule
10: from this list, not a word from a dictionary neither party has.
```

Those cannot both hold. `:5` states conformance to ASD-STE100; `:8-10` states that the actual authority is the ten bullets below and that the standard's own controlled vocabulary — which is most of what ASD-STE100 *is* — does not bind here. `:13` "Use an approved word when it has the required meaning" is then unenforceable by construction: approved by what, if not the dictionary `:10` says neither party has?

The unpinned identity compounds it. ASD-STE100 is issued (Issue 6, 7, 8…) and `AGENTS.md` names no issue, so "MUST use ASD-STE100" does not identify a document. This repository has an explicit position on exactly that failure — `docs/instruments/standards-registry.toml` fixes `selected_revision`, `publisher_url`, `supersedes`, and `verified_on` for every external reference it cites, precisely so a citation identifies something. That registry is scoped to vision-guidance references, so I am not claiming ASD-STE100 belongs in it; I am claiming the standard of citation the repository holds itself to is one this new citation does not meet.

The honest form costs one word and loses nothing: *these are the repository's writing rules, derived from ASD-STE100*. The bullet list is genuinely useful; the conformance claim is what cannot be supported. CONTRIBUTING.md:129 inherits the same wording.

---

## Non-blocking

- **`AGENTS.md:5` and `AGENTS.md:21` disagree on scope.** `:5` says *all* project-authored documentation MUST use it; `:21` says apply to new text and changed text; CONTRIBUTING.md:131-132 adds "not to legacy prose you happen to be near". The narrow reading is the intended one and the only workable one — pick it and state it once, at `:5`.

- **`AGENTS.md:14` "Use one word for one meaning" is already false in the tree, and cannot be fixed by new text alone.** `docs/instruments/evidence-plan.md` alone carries requirements baseline (`:31`), design baseline (`:41`), code baseline (`:52`), test baseline (`:60`), review baseline (`:77`), configuration baseline (`:85`), and baseline commit (`:129`); `scripts/structure-function-baseline.tsv` is a function-length ratchet; commit `1ed2f55` is "Add architecture baseline". Unlike the other nine bullets, this one is a property of the whole corpus — a new document obeying it in isolation just adds a tenth meaning. Either scope it ("do not introduce a new meaning for a term already used in this repository") or accept it is aspirational and say so.

- **`AGENTS.md` is not scanned by the check it ships with**, though it is the file most likely to cite others:
  ```console
  $ printf 'See `GHOST.md` for the rest.\n' >> AGENTS.md
  $ bash scripts/check-structure.sh
  check-structure: OK
  ```
  `:196` lists `./CLAUDE.md ./CONTRIBUTING.md`. Add `./AGENTS.md` and `./README.md`.

- **`:197` `[ -f "$doc" ] || continue` means the check enforces "no dead citation", not "the rules exist".** With both root files removed and `CONTRIBUTING.md` at `origin/main`'s text — the exact state the comment says this check exists to end — it passes:
  ```console
  $ rm AGENTS.md CLAUDE.md && git show origin/main:CONTRIBUTING.md > CONTRIBUTING.md
  $ bash scripts/check-structure.sh
  check-structure: OK
  ```
  Today the chain holds only because `CONTRIBUTING.md:130` happens to backtick `AGENTS.md`, and nothing pins that sentence. If the intent is that the rules file must exist, assert `AGENTS.md` directly — one `[ -s ./AGENTS.md ]`.

- **Path-qualified citations are not covered.** `[A-Za-z0-9_-]*\.md` cannot match a `/`, so `` `docs/instruments/evidence-plan.md` `` (CONTRIBUTING.md:53) — a document CONTRIBUTING sends contributors to for a four-step procedure — is unchecked. No false alarm today, since it exists, but it is the same breakage the check is named for.

- **The file header at `:2-18` lists what the script enforces and was not updated** with the new check.

- **The documents are not written to their own rules.** `check-structure.sh:187-190` is one 45-word sentence with three subordinate clauses, against `:15` "Use short, direct sentences" and `:17` "Give one instruction or one main statement in each sentence". CONTRIBUTING.md:132 "Review does." is an elliptical fragment carrying its verb from the previous sentence. `AGENTS.md:10` "a dictionary neither party has" is a rhetorical construction with no defined actor, against `:16`. A rules document that a reviewer could cite against itself on its first page will not be cited against anything else.

- **shellcheck** now reports SC2013 and SC2016 on `:199`. Info level and not a CI gate here — `:367` carries SC2016 too — but the `grep`/`while read` form in finding 2 clears both.

---

- **The primary case works, loudly and twice.** Deleting `AGENTS.md`, and separately truncating it to zero bytes, both produce two `FORBIDDEN` lines (one per citing document) and `check-structure: FAILED`, exit 1. I confirmed `status=1` inside the new function reaches the exit at `:408-411` rather than being swallowed — it does.
- **The single-citation path is correct** for a name that does not exist, at any position in the line, in either scanned file.
- **No false positives on the current tree**, including the code fences in CONTRIBUTING.md and the path-qualified `docs/…` citations.
- **`AGENTS.md` / `CLAUDE.md` do not contradict any existing convention** I could find in `CONTRIBUTING.md`, `README.md`, or the docs — the problems above are internal to the new text, not collisions with the old.

## Battery, run on `23d4782` with a clean worktree

| command | result |
| --- | --- |
| `cargo fmt --all --check` | exit 0 |
| `cargo clippy --locked --all-targets --workspace -- -D warnings` | exit 0, no warnings |
| `cargo test --locked --workspace --all-targets --no-fail-fast` | 21 suites, all `ok`, 0 failed |
| `RUSTDOCFLAGS="-D missing_docs -D rustdoc::broken_intra_doc_links" cargo doc --locked --workspace --no-deps` | exit 0 |
| `cargo build --locked --release --workspace` | exit 0 |
| `check-structure.sh` | OK (12 selftest cases — none for the new check) |
| `test-structure.sh` | OK (12 cases) |
| `check-release-manifest.sh` | OK (8 selftest cases) |
| `check-release-markers.sh` | OK (0.5.0; ABI 7, scene 1, corpus 5) |
| `check-instrument-requirements.sh` | OK (6 selftest cases, 73 requirements) |
| `check-standards-registry.sh` | OK (9 selftest cases, 3 references) |
| `check-recorded-counts.sh --selftest` / `check-recorded-counts.sh` | OK (4 cases) / OK (3 records) |

All mutations were reverted; `git status` clean.

luofang34:
Took all three blocking findings in `db48053`.

**The fabricated rationale.** You are right, and this is the one that bothers me most: the comment asserted that `CLAUDE.md` named the standard and pointed at a missing `AGENTS.md` "for years", in a repository six weeks old where neither file was ever tracked. In a repo whose whole practice is scripts declaring exactly what they prove and what they do not, inventing a history to justify a check is the wrong failure to ship. The reason is now the one that holds with no history at all: a reviewer who cites a standard the clone does not contain is citing something a contributor cannot obey.

**The greedy `sed`.** Confirmed exactly as you described — I reproduced both orderings. Now a `grep -oE` over every citation, matching the idiom already at `:367`. Verified ```It cites `GHOST.md` and then `AGENTS.md`.``` is refused, which is the case that previously passed.

**No selftest.** Added, covering both positions, since position is the thing that broke. `structure-selftest` is at 14 cases and both new ones go red against the old greedy pattern.

**The conformance claim.** Also right, and the contradiction was plain once you named it: the document claimed ASD-STE100 conformance, told readers not to consult the standard, and required "an approved word" from a dictionary the repository does not ship. All three are gone. The list now says it is *modelled on* the standard, that following it is not a conformance claim against it, and that a reviewer cites a rule from this list and nothing else. `use an approved word` is deleted rather than reworded — an unenforceable rule in a list of enforceable ones teaches contributors that the list is decorative.

Your remaining non-blocking items are noted and I have not dropped them. Two I want to answer rather than defer:

- **"one word for one meaning" is already violated seven ways by "baseline".** That is real and it is the most useful thing in the list to enforce, so it deserves its own change with a decided vocabulary — not a rushed rename inside a PR about the rule list.
- **`AGENTS.md` is not itself scanned, and the pre-PR state passes.** Correct on both counts. The check is over the pointer, not the prose, and the pointer is only broken once something cites it — which is the honest scope, now stated as such rather than implied.

Full battery green: fmt, clippy `-D warnings`, the workspace suite `--no-fail-fast`, doc gate, release build, and every `check-*.sh` with its own selftest count.

luofang34:
**REQUEST CHANGES** — three of the four findings are closed and verified by probe. The fourth is closed in two documents of three: `CONTRIBUTING.md` still asserts flat ASD-STE100 conformance, and that section is this PR's own addition, not legacy text.

Verified at `db48053`, detached on `origin/docs/track-the-documentation-language-rules` (both PR branches are checked out in other worktrees, so I reviewed the remote ref directly).

## Prior findings

**1. Fabricated history in the `check-structure.sh` comment — CLOSED.** Diffed `23d4782` against `db48053`. The claim that `CLAUDE.md` "named ASD-STE100 and pointed at `AGENTS.md` for years while neither file was in the repository" is gone. What replaces it needs no history and stays true whatever the repository did: a reviewer who holds a contributor to a standard the clone does not contain is citing something the contributor cannot obey. Good — the reason now survives a history rewrite.

**2. Greedy `sed` → `grep -oE` over every citation — CLOSED.** The old `sed -n 's/.*`\(…\)`.*/\1/p'` had a greedy `.*` that reached the *last* citation on a line. The replacement scans all of them, and I confirmed by construction rather than by reading the regex:

| appended to the document | exit | flagged |
|---|---|---|
| ``It cites `GHOST.md` and then `AGENTS.md`.`` | 1 | `GHOST.md` |
| ``It cites `AGENTS.md` and then `GHOST.md`.`` | 1 | `GHOST.md` |
| ``Three on one line: `AGENTS.md`, `GHOST.md`, `CLAUDE.md`.`` | 1 | `GHOST.md` |
| same, appended to `CLAUDE.md` instead | 1 | `GHOST.md` |

Both orderings refused, and the middle-of-three case — which the greedy `sed` would also have missed — is refused too.

**Deletion / emptying probes:** deleting `AGENTS.md` fails the check with a FORBIDDEN line from *both* `CLAUDE.md` and `CONTRIBUTING.md`; truncating it to zero bytes fails identically. `[ ! -s ]` covers both, as intended.

**3. Selftest covers both citation positions — CLOSED.** Present at `test-structure.sh:153-154`, and `test-structure.sh` reports `OK (14 cases)`, up from 12.

**4. ASD-STE100 conformance claim — PARTIALLY CLOSED (2 of 3 documents).** This is the blocking point.

`AGENTS.md` and `CLAUDE.md` are exactly right. `AGENTS.md` says the list is *modelled on* the standard "and it is not that standard," gives the reason (the repository distributes neither the specification nor its dictionary), scopes review to "a rule from this list and nothing else," and closes with "makes no conformance claim against ASD-STE100." `CLAUDE.md` mirrors it. `use an approved word` is deleted, not softened — `grep -rn "approved word"` returns nothing.

`CONTRIBUTING.md:129` still reads:

> Project-authored documentation **uses ASD-STE100 Simplified Technical English**.

That is the flat conformance claim the other two documents were rewritten to retract, and the two statements cannot both be true. This is not legacy prose the PR happens to sit near: `git log origin/main..HEAD -- CONTRIBUTING.md` shows the section was introduced by `87f9172` and revised by `23d4782`, and the fix commit `db48053` touched `AGENTS.md`, `CLAUDE.md`, `check-structure.sh`, and `test-structure.sh` — but not `CONTRIBUTING.md`. The document most likely to be read first by a human contributor is the one left asserting conformance to a specification the clone does not contain, which is the precise condition `check_root_document_pointers` was added to end. One sentence, aligned with the other two documents' framing, closes it.

## Does the list claim only what it can enforce?

Mostly, and the framing is honest: `CONTRIBUTING.md` says plainly "No gate checks the prose itself. Review does," so these are review criteria, not gates. Judgement-dependence is therefore fine — but a rule still has to be *citable*, meaning two reviewers reading the same sentence should reach the same verdict.

**Still unenforceable: "Use short, direct sentences."** It names a property with no threshold, so a reviewer can cite it but cannot adjudicate it, and a contributor cannot tell in advance whether a sentence complies. ASD-STE100's own version of this rule carries word counts — 20 for procedural sentences, 25 for descriptive — and those numbers are the enforceable part. This list kept the adjective and dropped the number. Since the repository has decided not to distribute the specification, the number has to be restated here or the rule cannot be applied. A stated bound would also make it the one rule in the list that could later become a script.

Second, weaker: **"Use the same technical name in all related documents and diagrams"** — "all related" is unbounded, so a reviewer cannot establish the set over which the rule holds. Naming the scope (this repository's tracked documentation) would fix it.

The rest are fine as review criteria. "Use one word for one meaning" and "Define a necessary technical name before you use it" are judgement calls a reviewer can defend from the text in front of them, which is the standard the list should be held to.

## New findings

**A. Should fix — the new selftest case sits outside the load-bearing trap.** `test-structure.sh` states its own invariant at the top: "Each case mutates the worktree and restores it, so the trap is load-bearing: an aborted run must not leave a planted crate behind." The citation case backs `CONTRIBUTING.md` up to its own `mktemp` and restores it with a bare `cp` at the end of the loop body; `cleanup()` (lines 23-28) covers `probe_dir`, `probe_doc`, and `Cargo.lock`, and knows nothing about `CONTRIBUTING.md` or that temp file.

I injected an abort after the append and ran it:

```
SIMULATED ABORT (reviewer probe)
aborted with exit=9
$ git status --porcelain
 M CONTRIBUTING.md
$ tail -1 CONTRIBUTING.md
It cites `GHOST.md` and then `AGENTS.md`.
```

The consequence is worse than a dirty file. The leftover line makes `check-structure.sh` fail from then on:

```
FORBIDDEN: ./CONTRIBUTING.md points at `GHOST.md`, which is missing or empty
```

A contributor who interrupts the selftest gets a repository that fails its own structure check citing a document they never wrote, and the mktemp backup leaks too. Registering the restore in the existing `cleanup()` is the whole fix, and it makes the new case honour the rule the file already states.

**B. Nit — `[A-Za-z0-9_-]*` should be `+`.** The star allows an empty basename, so a document containing a bare backticked `` `.md` `` (writing about the extension itself) is flagged: `FORBIDDEN: ./CONTRIBUTING.md points at `.md`, which is missing or empty`. Confirmed by probe. Harmless today, a confusing false positive the day someone writes it.

**C. Nit — `AGENTS.md` is exempt from the pointer discipline it introduces.** The loop covers `./CLAUDE.md` and `./CONTRIBUTING.md` only. I appended a citation of a missing `GHOST.md` to `AGENTS.md` and the check passed. The rules file is the one document whose citations a reviewer is most likely to follow; adding it to the loop costs one word.

**D. Nit — the new comment reintroduces refactor history.** `test-structure.sh:146-150` explains the case with "The first version of this check read only the last citation per line." That is true rather than fabricated, so it is much milder than finding 1 — but `db48053` removed narrative history from `check-structure.sh`'s comment and added narrative history here in the same commit. The invariant states without it: *a citation must be refused wherever it sits on the line; a parser that reads only the last citation per line accepts a missing file named before a present one.* Same warning, nothing to rot.

**Observations, not findings.** Nested citations such as `` `docs/instruments/evidence-plan.md` `` are out of scope by construction — the character class excludes `/`, so they are skipped silently; that matches the comment ("root-level markdown file") and I mention it only so the boundary is on the record. And an `AGENTS.md` containing a single newline passes `-s`, since one byte is not empty; "missing or empty" is what the check claims, so this is accurate rather than wrong.

## Battery

All green on this branch, run in this worktree:

```
cargo test --locked --workspace --all-targets --no-fail-fast   exit 0
check-structure.sh                OK          test-structure.sh               OK (14 cases)
check-release-manifest.sh         OK          test-release-manifest.sh        OK (8 cases)
check-instrument-requirements.sh  OK (73)     test-instrument-requirements.sh OK (6 cases)
check-certification-claims.sh     OK          check-standards-registry.sh     OK
test-standards-registry.sh        OK (9 cases)  check-release-markers.sh      OK
check-recorded-counts.sh          OK          trace-report.sh                 OK
detect-target.sh                  OK
```

`check-baseline-survives-merge.sh` and `test-baseline-survives-merge.sh` are absent on this branch — they belong to #69, so that is expected, not a failure. Every mutation was reverted with `git restore`; `git status --porcelain` is empty.

luofang34:
Blocker fixed in `1a8e441`, along with both of the should-fix items.

**`CONTRIBUTING.md:129`.** You are right, and the way it happened is worth naming: I retracted the claim in the two documents the *review* had pointed at, and never checked whether the same branch had put it anywhere else. It had — in the section this PR itself added, in the document a human contributor opens before either of the other two. So the branch shipped the claim it elsewhere disclaims. All three documents now say the same thing: the list is modelled on the standard, following it is not a conformance claim, and a reviewer cites a rule from the list. I grepped for every remaining form of the claim and there are none.

**The unenforceable rule.** "Use short, direct sentences" is now "under 20 words in a procedure, under 25 in a description" — the counts the standard this list is modelled on actually carries. Your framing settled it: the value of the list is that a reviewer can point at a rule and be right, and an adjective with no threshold cannot do that. "The same technical name in all related documents" now reaches the documents a change touches, which is a scope a contributor can satisfy rather than an unbounded one.

**The selftest mutating a tracked file outside the trap.** Confirmed by your method — I injected an abort mid-case and watched it leave the `GHOST.md` line in `CONTRIBUTING.md`, after which the structure gate fails citing a document the contributor never wrote. Every tracked file the script edits is now restored by the trap, beside `Cargo.lock` which was already doing it correctly. Verified with the abort probe: zero `GHOST` lines survive.

That failure has a sibling I hit myself today — a plain `cp` restore prompting interactively and silently leaving a gutted file in place — so the general lesson is landing from two directions at once: cleanup belongs somewhere that runs when the happy path does not.

Structure selftest 14 cases, structure gate OK, workspace suite green.

luofang34:
**Approve.** The blocker is closed, both should-fix items are closed, and the abort and positional probes both come back clean. Three non-blocking notes below, one of which is that the rule list's own prose breaks the rule it introduces.

Reviewed at `1a8e441` against `origin/main` (`830e9a6`), detached, in a scratch worktree.

## The blocker: the conformance claim in `CONTRIBUTING.md`

**Closed.** I grepped every form of the claim — `ASD-STE`, `STE100`, `Simplified Technical English`, `conformance`, `conform`, `compliant`, `noncompliant` — across the whole tree. Five hits land in the three documents at issue, and every one is a retraction:

- `AGENTS.md:6` — "modelled on ASD-STE100 Simplified Technical English, and it is not that standard"
- `AGENTS.md:9` — "makes no conformance claim against ASD-STE100"
- `CLAUDE.md:5` — "following them is not a conformance claim against it"
- `CONTRIBUTING.md:130` — "The list is modelled on ASD-STE100 Simplified Technical English"
- `CONTRIBUTING.md:131` — "following it is not a conformance claim against that standard"

No surviving sentence asserts the repository *uses* the standard. The remaining tree-wide hits are the `indicate-instrument-conformance` crate and the scene-conformance corpus, which are a different word doing a different job.

## Should-fix items

**Word counts — closed.** `AGENTS.md:14` now reads "Keep a sentence under 20 words in a procedure and under 25 words in a description." That is a threshold a reviewer can hold up against a sentence and be right or wrong about, which is what the old adjective was not.

**Scope of the technical-name rule — closed.** `AGENTS.md:20` now reads "across the documents a change touches" — a set a contributor can enumerate before pushing, where "all related documents and diagrams" was not.

**The selftest mutating a tracked file outside the trap — closed, and I confirmed the guardrail is load-bearing.**

Abort probe: injected `exit 7` immediately after `printf '%s\n' "$line" >> CONTRIBUTING.md` in the citation loop of `scripts/test-structure.sh`, so the run dies with the invented citation on disk and the case's own restore never reached.

- run exits 7;
- `git diff -- CONTRIBUTING.md` is empty — the trap restored it;
- no temp file leaked;
- `git status --porcelain` shows only my injected edit, and is empty after I put the file back.

One methodology note, because it bit me: BSD `mktemp` ignores `TMPDIR` for a no-template invocation and writes to the per-user Darwin temp directory, so a `TMPDIR`-based leak check proves nothing on macOS, and a snapshot-diff of the shared directory picks up whatever else is running on the box — my first attempt "found" a leaked `Cargo.lock` copy that turned out to belong to another process. Re-run with a `mktemp` shim forcing every temp into an isolated directory: **zero leftovers, on both the clean run and the aborted one.**

Counterfactual, to show the trap lines are doing work rather than duplicating the in-case restore: with the two `cleanup()` lines removed and the same `exit 7` injected, the run leaves

```
+It cites `GHOST.md` and then `AGENTS.md`.
```

appended to the tracked `CONTRIBUTING.md`, and `check-structure.sh` then reports `FORBIDDEN: ./CONTRIBUTING.md points at 'GHOST.md'` against a document the contributor never wrote. Exactly the failure described, and the fix removes it.

## Positional probe on the citation check

All four spellings refused, each naming the file it could not resolve:

| citation appended to `CONTRIBUTING.md` | exit | finding |
| --- | --- | --- |
| `` `GHOST.md` `` then `` `AGENTS.md` `` (missing first) | 1 | names `GHOST.md` |
| `` `AGENTS.md` `` then `` `GHOST.md` `` (missing last) | 1 | names `GHOST.md` |
| `` `AGENTS.md` ``, `` `GHOST.md` ``, `` `CHANGELOG.md` `` (missing in the middle of three) | 1 | names `GHOST.md` |
| `` `GHOST.md` ``, `` `PHANTOM.md` ``, `` `AGENTS.md` `` (two missing) | 1 | names **both** |

The greedy-`sed` defect is gone: position on the line no longer decides anything, and two missing citations on one line produce two findings rather than one. The unmutated tree exits 0.

## Does the list now claim exactly what it can enforce?

Close to it. Taking "enforceable" as the PR itself defines it — a reviewer cites a rule and a specific sentence, and the author can check the finding without a taste argument, since `CONTRIBUTING.md:134` states plainly that no gate reads the prose:

- *Write in English* — adjudicable, near-vacuous.
- *Keep a sentence under 20/25 words* — adjudicable; you can count. The soft edge is that nothing says which passages are "a procedure" and which "a description", so an argument bottoms out at the looser bound of 25. That is a bounded disagreement, not an open one.
- *Use active voice when you can identify the actor* — adjudicable by demonstration: the escape clause is the author's, but a reviewer rebuts it by naming the actor.
- *One instruction or one main statement per sentence* — adjudicable; you point at the second main clause.
- *Define a necessary technical name before you use it* — adjudicable on ordering. "Necessary" is the author's call, but that only ever narrows the rule.
- *Same technical name across the documents a change touches* — adjudicable, and now over a stated set.
- *Keep code identifiers, protocol fields, quotations, and product names exact* — adjudicable against the referent.

**The one rule I would still call unenforceable as written is `AGENTS.md:13`, "Use one word for one meaning."** It is the only rule whose corpus is unstated. To cite it a reviewer must exhibit two usages, and the list never says where the second usage may come from. If it may come from legacy prose, the rule contradicts `AGENTS.md:22`, which binds only new and changed text; if it may not, the rule silently means "within this change" and should say so. The neighbouring rule got precisely this narrowing in the last commit — "across the documents a change touches" — and this one did not. Same treatment would close it.

## Non-blocking findings

**1. The new text breaks the new rule.** `AGENTS.md:22` says the rules apply to all new text, and two new sentences exceed 25 words in a passage that is plainly a description:

- `AGENTS.md:5-8` — "The list is modelled on ASD-STE100 … and nothing else." — **37 words**, and three main clauses joined by "and" and "so", so it is also the clearest available violation of `AGENTS.md:18`.
- `CONTRIBUTING.md:130-133` — "The list is modelled on ASD-STE100 … the specification or its dictionary." — **30 words**.

Both contain a colon, and both fall under 25 on each side of it, so there is a reading that rescues them — but the rule as written bounds a sentence, not a clause. Everything else in all three documents is inside the bound; I counted. This is not a correctness problem, and the surrounding prose predates the number by one commit. It is worth a pass anyway, because the argument for putting a number in the rule is that a reviewer can point at a rule and be right, and the first paragraph a reviewer meets is one they would have to cite.

**2. `check_root_document_pointers` scans two documents, not the root.** The loop covers `./CLAUDE.md` and `./CONTRIBUTING.md`. `README.md` cites `` `CONTRIBUTING.md` `` and `` `CHANGELOG.md` `` and is not scanned, and `AGENTS.md` — the cited rule list itself — is not scanned either. Both are clean today, so nothing is broken; but the check's stated reason is that a citation a clone cannot resolve is not a rule, and that reason does not stop at those two files. Adding them to the loop is a one-word change and needs no new case.

**3. Path-form citations are out of scope, correctly.** `` `docs/instruments/GHOST.md` `` is not matched and does not refuse — I probed it. The pattern is root-level by construction and the comment says so, so this is a deliberate boundary rather than a hole. Flagging it only so the next reader does not rediscover it as a bug.

## Gates

All from the detached PR head; `git status --porcelain` empty before and after each.

- `bash scripts/check-structure.sh` — OK (runs the selftest: **14 cases**)
- `bash scripts/test-structure.sh` — OK (14 cases)
- `cargo test --locked --workspace --all-targets --no-fail-fast` — exit 0
- `check-certification-claims` OK · `check-instrument-requirements` OK (6 selftest cases, 73 requirements) · `test-instrument-requirements` OK · `check-release-manifest` OK (8 cases) · `test-release-manifest` OK · `check-release-markers` OK · `check-standards-registry` OK (9 cases) · `test-standards-registry` OK · `check-recorded-counts --selftest` OK (4 cases) · `check-recorded-counts` OK (3 records) · `trace-report` OK · `detect-target` OK
- `cargo run -p instrument-bench` — both digests match pin, 280 admission cases pass
- `evidence-gate --resolve-selectors` and `--require-resolvable` — VALID, 30 nodes / 73 edges
- `shellcheck scripts/check-structure.sh scripts/test-structure.sh` — clean at `-S warning`. Four `SC2016` infos, all on single-quoted grep patterns; one predates this branch at `check-structure.sh:365`, so the new lines match the file's existing style, and there is no shellcheck step in CI.

PR #74, opened 2026-08-18
Say what a dot is worth (#54)

### LINKED ISSUE #54: The CDI scale is fixed and unannunciated

luofang34:
Closes #54. Targets `abi/v8-batch` — the third slice of the coordinated ABI revision (#58).

The CDI's lateral scale was a constant: two dots meant fifty metres, everywhere, because the one feeder that fills the group flies a terminal profile. Nothing on the glass said so. When guidance later flies phases that want different scales, the same needle position silently means different distances — the "same picture, different meaning" failure the reference annunciations elsewhere on this panel exist to prevent.

## The scale is declared, not derived

`scale` is an enumeration on the Nav group, in the spare byte the payload already carried, so **the payload length does not move**. It is an enumeration rather than a metres-per-dot number because a display that derived a label from a distance would be inventing meaning the source never stated.

**An unknown scale fails the whole nav group.** A deflection in dots means nothing until the scale says what a dot is worth, and a needle drawn at a guessed scale is worse than no needle. This is the same fail-closed direction as an unknown source or an unknown TO/FROM.

The feeder's conversion becomes a table keyed by the scale — the one place metres become dots — and the panel annunciates the scale under the CDI's own gate, so the label and the needle appear and disappear together.

## Vocabulary

`ENR`, `TERM`, `APR`. The glyph pack has no `+` and no `/`, so the richer per-approach names of the LNAV+V / L/VNAV family cannot be spelled yet. That is a glyph-pack change and it belongs with the change that needs it, as issue #54 says.

## What the fail-closed direction found

Every fixture that had been relying on a defaulted scale now declares one — the heading-reference sweep, the resolve fixtures, the validate fixtures and the canonical states. That is the mechanism doing its job: a scale nobody declared is not a scale.

## Guardrails

- Each scale annunciates its own label.
- An unknown scale takes the needle with it rather than drawing at a default, and invents no label.

## Pins

The Nav payload changes value but not length, so the ABI goldens are regenerated. The HSI emits a new label, so its raster baseline, the side-by-side composed hash, the composition digest and the screen-composition digest move; `release-manifest.json` is regenerated. ATT-01 re-records in its own commit.

Local gates: fmt, clippy `-D warnings`, `test --all-targets`, both evidence gates VALID, check-release-markers, check-release-manifest.


luofang34:
**REQUEST CHANGES** — the wire field does not land where the batch's own layout contract says it does (and that placement is what makes an undeclared scale read as `Enroute` rather than fail closed), and the HSI's only full-scale-CDI stress fixture went dark without a single test noticing.

Reviewed `origin/abi/v8-batch...origin/fix/54-cdi-scale-mode` (base `176624c`, head `43d4b9f`). Every gate is green — that is the problem in finding 2, not a reassurance.

---

## Blocking

### 1. The Nav field does not land where the batch contract says, and the placement it chose is not fail-closed

`crates/indicate-instrument-state/src/group_id.rs:53` states the growth policy and `:61` states this exact allocation:

> Field appends to existing groups follow the stamped lane growth policy: a new field appends AFTER the trailing `age_ms` […]
> | Nav (0x04) | `scale_mode u8` (Enroute/Terminal/Approach, 255 fail-closed Unknown) and `facility_type u8`, both after `age_ms` |

The implementation puts `scale` in the reserved padding byte at payload offset 3 (`crates/indicate-instrument-state/src/abi/v8/stamped.rs:208` encode, `:166` decode), leaving `min_len(Nav) == 42` untouched (`crates/indicate-instrument-state/src/abi/v8.rs:112`). The PR does not amend the contract row. Worse, `stamped.rs:3`–`:5` — three lines above the nav layout row it *did* edit — still reads "a batch field append lands AFTER that trailing age (see `crate::group_id` for the v8 allocation contract)". The tree now contradicts itself in two files, and `CHANGELOG.md:74` documents the third position ("takes the spare byte at offset 3").

This is not bookkeeping. The append-after-`age_ms` design buys fail-closed absence for free: a 42-byte payload from a writer that does not know about `scale` is shorter than `min_len`, gets `AbiError::GroupTruncated`, and the nav group goes Missing. The padding-byte design makes "undeclared" indistinguishable from `Enroute` — the *loosest* row in the new table, 100 m/dot, four times looser than the terminal scale every fixture in the tree declares. `CHANGELOG.md:120` states the consequence outright:

> A writer that leaves the spare byte at zero declares enroute, which is a scale, so this one fails loudly only for writers that put something else there.

A writer whose zero byte silently becomes a declaration of the loosest scale, annunciated "ENR" on the glass in the source colour, is exactly "a deviation read against the wrong scale is misleading information" — the hazard the PR exists to close.

Reproduction, using the suite's own fail-safe test:

`crates/indicate-instrument-state/src/abi/v8/tests.rs:244` `unknown_wire_enum_values_decode_fail_safe_not_benign` builds `let mut nav = [0u8; 42]` (`:249`), sets only bytes 0 and 1, and asserts source and fromto decode `Unknown`. Byte 3 stays 0, so the same frame now decodes `NavScale::Enroute` — and the test's own docstring says unknown values "must decode to Unknown […] it never masquerades as no-source". The scale masquerades, in the one test written to forbid it.

There is also no purpose-written guardrail for the new codec. Grep confirms `scale`/`NavScale` appears nowhere in `crates/indicate-instrument-state/src/abi/v8/tests.rs`. I gutted the decoder two ways to measure what covers it:

- `scale: NavScale::from_u8(get_u8(p, 3))` → `scale: NavScale::Terminal`: exactly one failure, `abi::v8::tests::unknown_variants_encode_as_255_and_round_trip_explicitly` — a pre-existing test written for source/fromto that happens to assert whole-struct equality and happens to use `NavData::default()` (scale `Unknown`).
- `NavScale::from_u8`'s `_ => Self::Unknown` → `_ => Self::Terminal`: same single incidental failure.

So the fail-closed sentinel is held up by one accident. Nothing asserts byte 3 == 255 for `Unknown`, nothing asserts an out-of-range byte resolves `Unknown`, and nothing asserts what a 42-byte payload from a scale-unaware writer does. "Every PR lands the fix **and** the guardrail" (CONTRIBUTING.md:51) is not met for the wire half.

Either move the field to where the contract says and let a short payload fail closed, or amend the `group_id.rs` row and the `stamped.rs` header to describe the padding-byte allocation with the justification for accepting `0 == Enroute` — but not both positions in one tree. Whichever way it goes, the codec needs its own tests.

Two notes on scope, since it cuts the other way: v8 is unreleased (`CHANGELOG.md`, "Each allocation lands as its own change onto this version before it is released"), so there is no fielded v8 writer to break — that is a real argument for the padding byte, it just is not the argument the tree currently makes. And `facility_type` is correctly **absent**: this PR implements only what #54 needs and drags none of #55's work in. One issue per PR is respected; only the contract table over-promises.

### 2. `hsi_reciprocal_course` — the HSI's full-scale CDI stress fixture — went dark, and the digests were re-pinned over the blank picture

`sets/indicate-instrument-panels/src/descriptors/extreme_states.rs:86` is documented as "Course exactly reciprocal to the flown track, **full-scale deviation**, a zero-distance waypoint, and a heading on the 360/0 wrap." It rebuilds `state.nav` from scratch and closes with `..NavData::default()` at `:99`. `NavScale`'s `#[default]` is `Unknown` (`crates/indicate-instrument-state/src/aircraft.rs:120`), so the new `nav_fault` scale check (`validate.rs:134`) fails the whole group.

The PR updated six other `NavData` construction sites for the new field and missed this one.

Reproduction (temporary probe over `HSI_DESCRIPTOR.extreme_states`, resolved through `resolve` + `draw_hsi`, reverted after):

on `origin/abi/v8-batch`:
```
extreme reciprocal-course nav_status=Valid shows=true
  texts=[… "DIST NM", "0.0", "CRS", "180°", "029°", "V", "SIM"]
```
on `origin/fix/54-cdi-scale-mode`:
```
extreme reciprocal-course scale=Unknown nav_status=Failed shows=false
  texts=[… "DIST NM", "--.-", "CRS", "---°", "029°", "SIM"]
```

The full-scale deviation bar, the course arrow, the FROM triangle, the vertical-deviation diamond, the distance and the course readout are all gone. The fixture that exists to stress CDI geometry now stresses nothing, and `BUILTIN_SCENE_DIGEST` (`sets/indicate-instrument-panels/src/descriptors.rs:377`) and the composed-frame hash were re-pinned to that blanked picture.

That the digests enshrine it is provable: delete the scale clause from `nav_fault` and `descriptors::digest_tests::the_builtin_scene_digest_is_pinned`, `composing_the_builtin_set_reproduces_the_pinned_digest` and `regrouping_panels_without_reordering_them_is_digest_neutral` all go red — the pinned digest depends on that group being Failed.

Nothing flagged it. `cargo run -p instrument-bench` reports "admission: 280 cases pass, 196 counted warnings" on both the base and this branch: the admission harness did not notice an extreme state losing its entire guidance layer.

Fix is one line at `extreme_states.rs:99` (declare a scale, presumably `Approach` given the full-scale deflection this fixture models), then re-pin. Please also re-check that the re-pinned baselines represent the picture you intend rather than the picture the tree happened to produce.

### 3. The drawing code has no defense against an unknown scale — the whole guard is delegated to `validate.rs`

`sets/indicate-instrument-panels/src/hsi/cdi.rs:35`:
```rust
let label = nav.data.scale.label();
if label.is_empty() {
    return Ok(());
}
```
`draw_cdi` has already drawn the needle by then (`hsi.rs:87`), so an empty label means dots with no scale — the exact picture the PR's own docstring at `cdi.rs:30` says cannot happen ("An unknown scale never reaches here: the nav group fails first, and the needle goes with it").

It happens today, in this repo's own test suite. `hsi/tests.rs:119`, `:325` and `:405` construct `NavResolved` directly with `status: Valid/Degraded` and `..NavData::default()`, i.e. `scale: Unknown`, and `draw_hsi` paints the full CDI. Probe over all three statuses that pass `shows_value()`:

```
status=Valid    polys=4 texts=[… "DIST NM", "40.3", "CRS", "020°", …]   (no scale label)
status=Degraded polys=4 texts=[… "CRS", "020°", …, "NAV", …]            (no scale label)
status=Stale    polys=4 texts=[… "CRS", "020°", …, "NAV", …]            (no scale label)
```

Needle, distance and course all paint; nothing names the scale. `hsi.rs:43` already documents the house rule for precisely this case — "Belt and suspenders: the basis names the selection, the signal's own status still gates the paint — a basis that disagrees with its signal draws nothing rather than a plausible rose." The CDI deserves the same second gate: when the scale has no label, suppress the needle rather than returning `Ok(())`, and let the panel's own tests prove it. As landed, the safety property lives in exactly one place and the comment asserting it is falsifiable from within the crate.

---

## Non-blocking

- **`cdi.rs:19` — forward reference to a label that does not exist.** "Where the scale label sits: under the receiver label, so the two facts about the needle — which receiver drives it, and what a dot is worth — read together." The HSI draws no receiver label; the only `draw_source_label` call is for `HDG` at `hsi.rs:140`. The receiver label is #55's deliverable. This is an annotation whose truth depends on a future event, which the comment rules ban outright.

- **`cdi.rs:22` — the label overlaps the CRS box border.** `SCALE_LABEL_POS = (60.0, CY + 128.0)` = (60, 318) at size 14, `Anchor::CENTER`. `VAlign::Middle` places the ink top at `y - size/2` (`crates/indicate-instrument-raster/src/text.rs:46`), so ink spans y ∈ [311, 325]; `course_box` fills `rect(2.0, 322.0, 112.0, 36.0)` with a 1.5 stroke (`boxes.rs:77`), outer edge 321.25. "TERM" ink width is `4·14·6/7 − 14/7 = 46` px centred at x=60, so x ∈ [37, 83] — inside the box horizontally. The Guidance layer paints after Tapes, so the glyph bottoms land on the box's top border. Nudge it up ~6 px or move it clear.

- **`nav_display.rs:69`–`:79` — two unrelated comments fused by the edit.** The pre-existing paragraph explains why a non-tracking snapshot clears TO/FROM and uses a finite zero; the new paragraph about the declared scale was appended to it with no blank line, and `const SCALE` now sits between that paragraph and the `let tracking` line it describes. Reads as one comment for `SCALE`, half of which is about something else. Separate them and put each above its own statement.

- **`nav_display.rs:33` overstates.** "An undeclared or unknown mode selects none" is true of `NavScale::Unknown` but not of the wire: an undeclared scale on the wire is byte 0, which selects `Enroute`. Same sentence, different answer at the two layers — see blocking finding 1.

- **Two thirds of the new scale table is dead and untested.** `LATERAL_M_PER_DOT_ENROUTE` (100.0) and `LATERAL_M_PER_DOT_APPROACH` (10.0) have no caller and no assertion anywhere in the workspace; grep for `LATERAL_M_PER_DOT` finds only `nav_display.rs` and the one test using `_TERMINAL`. `lateral_m_per_dot`'s sole caller passes a compile-time `const SCALE: NavScale = NavScale::Terminal` (`nav_display.rs:78`), so the `?` at `:79` can never fire and the `None` arm is unreachable. A transposed digit in either constant would ship silently. Note also that `fly_to_signs_and_units_convert_exactly` divides by the same constant it is checking, so it pins the sign convention, not the magnitude — pre-existing, but now the only coverage the table has.

- **`ENR` never appears in a pinned artifact.** Probe over the corpus and the HSI extremes: `typical` / `ias-without-tas` / `track-up` are Terminal, `fully-fed` is Approach, `nothing-fed` has no nav, and `reciprocal-course` is the broken Unknown above. TERM and APR are genuinely in the raster baseline and the scene digest — gutting `draw_scale_label` reddens `hsi_frame_hash_is_reproducible_and_pinned`, `side_by_side_composed_frame_is_reproducible_and_pinned` and all three digest tests, so those baselines are not vacuous. But `Enroute` — the value a zeroing writer produces, per finding 1 — is exercised only by the new unit test. Consider giving one canonical or extreme state an enroute scale.

- **`CHANGELOG.md:118`–`:122` reads poorly for a writer-facing instruction.** "A writer that leaves the spare byte at zero declares enroute, which is a scale, so this one fails loudly only for writers that put something else there" is three clauses and a double negative for what is a one-line warning. Something closer to "Byte 3 of the Nav payload is now the scale. A writer that leaves it zero declares enroute. Declare the scale the guidance is actually flown to." carries the same content in the register the rest of the section uses.

- **`NavData.scale` was appended after `from_ident`** (`aircraft.rs:99`), so the struct's field order no longer tracks the wire order it otherwise mirrors. Cosmetic, but it is also what let `extreme_states.rs:99` slip through — a field near `source`/`fromto` is harder to forget in a `..default()` literal than one after the idents.

---

## Gates run (full battery, from a clean worktree at `43d4b9f`)

| Gate | Result |
|---|---|
| `cargo fmt --all --check` | clean |
| `cargo clippy --locked --all-targets --workspace -- -D warnings` | clean |
| `cargo test --locked --workspace --all-targets --no-fail-fast` | 21 binaries, 0 failures (panels 85, state 181, raster 98) |
| `RUSTDOCFLAGS="-D missing_docs -D rustdoc::broken_intra_doc_links" cargo doc --locked --workspace --no-deps` | clean |
| `cargo build --locked --release --workspace` | clean |
| `cargo check … --target thumbv7em-none-eabihf` | clean |
| `cargo run -q -p instrument-bench` | scene digest `b4b9030f…` matches pin; screen-composition `00a8f979…` matches pin; admission 280 pass / 196 warnings; 15 renders |
| `check-structure.sh` | OK (12 self-test cases) |
| `check-release-manifest.sh` | OK (8 self-test cases; manifest is what the generator produces) |
| `check-release-markers.sh` | OK (0.5.0; ABI 8, scene 1, corpus 4) |
| `check-instrument-requirements.sh` | OK (73 requirements) |
| `check-standards-registry.sh` | OK |
| `trace-report.sh` / `detect-target.sh` | OK |
| evidence gate, `--resolve-selectors` and `--require-resolvable` | VALID, 30 nodes / 73 edges / 0 exceptions, both invocations |

**Pin consistency:** internally consistent. Every re-pinned value — `BUILTIN_SCENE_DIGEST`, the HSI raster baseline, `SIDE_BY_SIDE_HASH`, `BENCH_COMPOSITION_DIGEST`, both v8 golden hex fixtures, `release-manifest.json`, the CHANGELOG composition-digest row, and the ATT-01 evidence artifacts and graph — is what this tree computes. `CFG-WORKTREE` records commit/tree `b9890a0`/`b1aa839a` while head is `43d4b9f`; that is the usual self-referential evidence recording (the last commit only rewrites the artifacts) and both gate invocations accept it. The admission count did not move (280/196 on base and here). As you note, all of this needs recomputing the moment #72, #75 or #76 merges — and the scene digest in particular must be recomputed *after* fixing finding 2, since its current value encodes the blanked reciprocal-course frame.

Beyond the gutting runs above: the encoder/decoder agree on offset 3 and the golden fixtures cover the round trip in both directions (`assert_pinned` checks encode→hex and hex→decode→state); reverting `put_u8(p, 3, nav.scale.to_u8())` to `put_u8(p, 3, 0)` reddens both fixture tests. Gutting `draw_scale_label` reddens six tests across two crates, so the annunciation itself is not fail-open. `an_unknown_scale_fails_the_group_rather_than_guessing` genuinely fails when the scale clause is removed from `nav_fault` — that guardrail is real. Longer-payload forward compatibility is intact and unaffected: `decode_state` still counts a >42-byte nav group in `extended_groups`, and short payloads still return `GroupTruncated` — but neither behaviour changed here, because the payload length did not move, which is finding 1.

luofang34:
All three blocking findings fixed in `52ea313` (plus an evidence re-record in `69aaa69`). Thank you — the second one was the serious one and I would not have caught it.

**1. Wire placement.** You were right that the padding byte was the wrong home and that the contract already said so. `scale` now appends at offset 42; `min_len(Nav)` goes 42 → 43. An undeclared scale is now a short payload and the frame is refused, which is the fail-closed absence you pointed out was available for free. Two tests hold it: `a_nav_payload_without_the_scale_byte_is_refused` (removes the tail byte, fixes up the length, expects `Err`), and `unknown_wire_enum_values_decode_fail_safe_not_benign` — which you correctly identified as having gone quietly wrong — now carries a 43-byte payload with `0x7F` in the scale slot and asserts `NavScale::Unknown` beside the source and from/to assertions it already made.

**2. `hsi_reciprocal_course` went dark.** Confirmed exactly as you described: `..NavData::default()` picks up `NavScale::Unknown`, `nav_fault` fails the group, and the digests were re-pinned over a frame with no needle, no `DIST`, and no `CRS`. The fixture now declares `NavScale::Approach` — apt for a full-scale-deflection stress with vertical deviation — and the digests move back to a frame that stresses something. This is the finding I would most have wanted caught before merge: the pins would have looked healthy forever while the fixture protected nothing.

**3. No drawing-side defense.** The needle now carries its own gate on a nameable scale, in the same belt-and-suspenders shape `hsi.rs` already uses for the rose basis, with the reason stated where the gate is. `a_valid_group_with_an_unnameable_scale_still_draws_no_needle` builds the case you described — status says show, scale is `Unknown` — and asserts the needle's marks disappear. Verified it reddens when the gate is removed.

Also took the CHANGELOG contradiction: "This release implements the Air append only" is gone rather than reworded, since the bullet listing the batch's allocations is what states scope, and a second sentence tracking what has landed is a thing that goes stale every time something lands. The registry's append row now states the byte growth (42 → 43 → 44) so the `facility_type` follow-on has an unambiguous slot.

Left for their own changes: the `ENR` label appearing in no pinned artifact and the label-versus-CRS-box overlap at y 322 are real, and both are about where the label sits rather than about the wire — I would rather move it once, deliberately, than nudge it inside a wire-format fix. The dead `LATERAL_M_PER_DOT_*` constants come out with the `facility_type` work that gives them a caller.

Full local CI green: fmt, clippy `-D warnings`, `test --all-targets --no-fail-fast`, doc gate, release build, every `check-*.sh`, and the evidence gate `VALID` against the re-recorded artifacts (86/182/98).

luofang34:
**REQUEST CHANGES** — all three blocking findings are genuinely closed and independently verified, but the doc sweep missed one paragraph, and it is the one downstream writers act on: `CHANGELOG.md:127` still instructs them to use the design that was rejected, and states the opposite of the real failure mode.

Re-reviewed `43d4b9f..69aaa69` (`52ea313` fixes, `69aaa69` evidence). Full battery re-run from a clean worktree; every gate green.

---

## Finding 1 — wire placement and the fail-closed sentinel: **CLOSED**

`min_len(Nav)` is 43 (`abi/v8.rs:112`), `encode_nav` restores the reserved zero at offset 3 and writes `put_u8(p, 42, nav.scale.to_u8())`, `decode_nav` reads offset 42, and `group_id.rs:61` now reads "both after the group's existing tail (42 to 43 to 44 bytes)" — which also gives `facility_type` an unambiguous slot at 43. The tree no longer contradicts itself.

I read it out of the regenerated fixtures rather than trusting the diff. In `state-abi-v8.full.hex` the nav group is now `04 2b00` (tag, length 43), its payload opens `01 01 02 00` — source Gps, fromto To, course-ref 2, **reserved byte back to zero** — and the group closes `… 04 4b4d5259 00000000 | 01`, the scale byte appended after `from_ident`. Both ends of the codec are visible in the committed bytes.

The guardrails are now purpose-written rather than incidental. Two gutting runs:

- `min_len(Nav)` 43 → 42: `a_nav_payload_without_the_scale_byte_is_refused` FAILED (with `full_frame_is_pinned`, `data_gateway_frame_is_pinned` and `fixtures_round_trip_bit_exactly`).
- `NavScale::from_u8`'s `_ => Self::Unknown` → `_ => Self::Enroute`: `unknown_wire_enum_values_decode_fail_safe_not_benign` FAILED **directly** this time, not just the incidental whole-struct round-trip. That is the specific regression I raised, and it is now caught by the test whose docstring forbids it.

The strengthened test carries `7` in the scale slot (not `0x7F` as described in the handoff — behaviourally identical, both outside the known set).

Forward compatibility for a longer payload is unchanged and still holds: `decode_state`'s `len > min_len` branch counts the tail, and `an_appended_payload_tail_is_accepted_and_counted` exercises that shared loop on the Air group, so #55's 44-byte nav payload will be accepted and counted by this decoder.

## Finding 2 — `hsi_reciprocal_course` went dark: **CLOSED**

Independently re-probed rather than taken on the claim, same probe as before (resolve → `draw_hsi`, counting text runs and filled marks):

```
extreme reciprocal-course scale=Approach nav_status=Valid shows=true marks=15
  texts=[… "DIST NM", "0.0", "CRS", "180°", "029°", "APR", "V", "SIM"]
extreme track-up          scale=Terminal nav_status=Valid shows=true marks=15
canonical typical         scale=Terminal nav_status=Valid shows=true marks=15
canonical fully-fed       scale=Approach nav_status=Valid shows=true marks=15
canonical source-unusable scale=Terminal nav_status=Failed shows=false marks=7
canonical nothing-fed     scale=Unknown  nav_status=Missing shows=false marks=6
```

`DIST 0.0` and `CRS 180°` are back — matching the fixture's pre-PR behaviour on `origin/abi/v8-batch` — and 15 filled marks puts it level with every other live-guidance state, against 6–7 for the states that correctly draw no needle. `APR` annunciates. The zero-distance, full-scale-deflection stress fixture stresses again. The comment at `extreme_states.rs:99` explains why the scale is declared rather than defaulted, which is the right place for it.

## Finding 3 — no drawing-side defense: **CLOSED**

`hsi.rs:86` adds the gate with the belt-and-suspenders reasoning mirroring the rose-basis precedent twelve lines up, which is exactly the house pattern I pointed at.

Removing `&& !data.nav.data.scale.label().is_empty()` and running the whole workspace reddens **exactly one** test — `a_valid_group_with_an_unnameable_scale_still_draws_no_needle` — and nothing else. That is the right shape: the gate is a no-op for every state whose scale is nameable, so a test that fails only when the gate is gone is measuring the gate and not the paint. `filled_marks` comparing a named against an unnamed scale is a good differential; it does not care where the needle lands, only that its ink leaves with the scale.

---

## New blocking

### `CHANGELOG.md:127`–`:130` still documents the rejected design, and inverts the failure mode

Under **Notes for anyone re-pinning** — the section a downstream writer reads to migrate:

> - **A state writer must also declare the nav scale.** A writer that leaves the spare byte at zero declares enroute, which is a scale, so this one fails loudly only for writers that put something else there. Declare the scale the guidance is actually flown to.

There is no longer a spare byte carrying the scale, and the consequence is the opposite of what this says. A writer that omits the field emits a 42-byte Nav payload, which is below `min_len(Nav) == 43`, so `decode_state` returns `GroupTruncated` and **the whole frame is refused** — the behaviour the new `a_nav_payload_without_the_scale_byte_is_refused` asserts, and the behaviour the immediately preceding bullet spells out for the Air group: "The decoder rejects the whole frame, not that group, so every panel blanks — the failure looks like total signal loss rather than one short group."

So the release notes tell a writer their under-length frame will be quietly accepted as enroute, when in fact every panel in the set goes dark. That is the worst direction for this error to point, in the one document whose whole purpose is that "a changelog that has to be checked against the source is the archaeology it was written to remove" — and `check-release-markers.sh` checks the five values, not prose, so nothing else will catch it. The JS state writer pinned against these fixtures is precisely the reader affected.

The body text was updated correctly (`:73`–`:88` now describes the append and says why the spare byte was rejected); this bullet was left behind. It wants to mirror the Air bullet: emit the 43-byte payload, declare the scale the guidance is flown to, and a 42-byte payload takes the whole frame down.

---

## Non-blocking

**Closed by the fix:** `nav_display.rs:33` ("An undeclared or unknown mode selects none") is now accurate at both layers — with the append, the wire can no longer produce an undeclared scale at all, so the function doc and the codec finally agree.

**Still open from the first pass, all unchanged:**

- `hsi/cdi.rs:19` — "under the receiver label" still forward-references #55's not-yet-existing label. The HSI draws no receiver label today.
- `hsi/cdi.rs:22` — `SCALE_LABEL_POS` ink spans y ∈ [311, 325] (`VAlign::Middle`, `raster/src/text.rs:46`) against `course_box`'s rect top at 322; the Guidance layer paints after Tapes, so the glyph bottoms sit on the box border. Now visible in the reciprocal-course and fully-fed frames too, since both annunciate `APR`.
- `nav_display.rs:69`–`:79` — the new scale paragraph is still fused to the pre-existing TO/FROM paragraph with no blank line, and `const SCALE` still sits between that text and the `let tracking` line it describes.
- `LATERAL_M_PER_DOT_ENROUTE` (100.0) and `LATERAL_M_PER_DOT_APPROACH` (10.0) still have no caller and no assertion; `lateral_m_per_dot`'s `None` arm is still unreachable from its one caller's compile-time `Terminal`.
- `ENR` still appears in no pinned artifact. Re-confirmed by the probe above: Terminal in `typical` / `ias-without-tas` / `track-up`, Approach in `fully-fed` / `reciprocal-course`, nothing enroute. Two of three rows are now in the baselines, which is better than before.

**New, minor:**

- `abi/v8.rs:48` — "Landed so far: the Air group carries `tas_mps` … so its minimum payload is 16 bytes." This is the same stale claim the CHANGELOG correctly deleted ("This release implements the Air append only"), surviving in the module doc and now omitting the Nav append this PR lands. Same sweep, one file missed.
- `CHANGELOG.md:104` — "Each append goes after the trailing `age_ms`" is loose for Nav, whose append goes after the idents that themselves follow `age_ms`. `group_id.rs:61` states it precisely now; worth matching the wording.
- `stamped.rs:3` — "Every stamped group's minimum layout ends with its `age_ms`" was already loose for Nav before this PR (the idents) and is looser now (the scale). Not introduced here; flagging while the file is open.

---

## Gates (clean worktree at `69aaa69`)

| Gate | Result |
|---|---|
| `cargo fmt --all --check` | clean |
| `cargo clippy --locked --all-targets --workspace -- -D warnings` | clean |
| `cargo test --locked --workspace --all-targets --no-fail-fast` | 0 failures; panels **86**, state **182**, raster **98** |
| `RUSTDOCFLAGS="-D missing_docs -D rustdoc::broken_intra_doc_links" cargo doc` | clean |
| `cargo build --locked --release --workspace` | clean |
| `cargo check … --target thumbv7em-none-eabihf` | clean |
| `cargo run -q -p instrument-bench` | scene `7c9d50db…` matches pin; screen-composition `d328803d…` matches pin; admission 280 pass / 196 warnings |
| `check-structure.sh` | OK (12 self-test cases) |
| `check-release-manifest.sh` | OK — "release-manifest.json is what the generator produces from this tree" |
| `check-release-markers.sh` | OK (0.5.0; ABI 8, scene 1, corpus 4) |
| `check-instrument-requirements.sh` / `check-certification-claims.sh` / `check-standards-registry.sh` / `trace-report.sh` / `detect-target.sh` | OK |
| evidence gate, `--resolve-selectors` and `--require-resolvable` | **VALID**, 30 nodes / 73 edges / 0 exceptions, both invocations |

**Digests are computed, not hand-typed.** The bench recomputes both digests from the registry and reports "matches pin"; `check-release-manifest.sh` regenerates the manifest and diffs it. The HSI raster baseline and `SIDE_BY_SIDE_HASH` are correctly *unchanged* — they render `typical`, whose Terminal scale did not move — while the scene and screen-composition digests moved because the corpus-wide composition includes the extremes.

**Evidence re-record is consistent.** Counts 86 / 182 / 98 match my own run exactly. `config-digest` and `run-id` are `52ea31337f17ba7851a7968da281d209bf523d92` across all three artifacts and all three `RESULT-*` nodes. Each `output-digest` matches the artifact I hashed myself:

```
6d1dba694d34d2…  raster.run.txt
9bf6a326bdccb6…  presentation.run.txt
6b1b22367fb306…  panel.run.txt
```

`CFG-WORKTREE` records commit `52ea3133…` / tree `c8ea02c1989d7109d14b8ba7844892d326a8159c`, which is `git rev-parse 52ea313^{tree}` — the tree the results were produced against. HEAD's tree differs only because `69aaa69` rewrites the artifacts themselves; both gate invocations accept it, same self-referential pattern as the previous record.

**Pin consistency:** internally consistent, and still needs recomputing after any of #72 / #75 / #76 merges.

Beyond the three gutting runs above: I re-parsed both regenerated golden fixtures by hand to confirm the length field, the restored reserved zero and the appended scale byte independently of the diff; confirmed the short-payload test's length fix-up (`short[payload - 2] = 42`) leaves every following group correctly framed, so the `is_err()` it asserts can only come from the `min_len` check; and confirmed the drawing-side gate is inert for all six canonical and extreme states, so it adds a guard without moving any pinned paint. Every mutation was reverted with `git checkout` and the worktree is clean at `69aaa69` — `git status` empty, `git diff HEAD` empty.

PR #75, opened 2026-08-18
Show what the airframe is set to (#57)

### LINKED ISSUE #57: No flap or trim indication exists; the airframe-config vocabulary is only a reserved id

luofang34:
Closes #57. Targets `abi/v8-batch` — the fourth slice, and the first to give a reserved id a variant.

Flap position and trim appeared nowhere. The only trace on any surface was a monitor-text fixture string, which is free advisory text the monitor contract is explicit is not a flight instrument.

## The group

`AirframeConfig` at 0x13, stamped, 24 bytes: sensed flap ratio, selected flap detent, and elevator, aileron and rudder trim ratios. Every field is optional, because a vehicle without the sensor must show `Missing` and not a substitute (ADR-0017). A ratio outside the range its axis is defined over **faults the group rather than being clamped** — a clamped pointer would sit at a limit the airframe never reached.

**Sensed and selected are never conflated.** The selected detent draws as its own mark beside the sensed pointer, so a flap in transit reads as two marks apart, and a flap that never reaches its detent reads as two marks that stay apart. Averaging them would erase exactly the fact worth seeing.

## The panel, and where it ships

A conventional-instrument surface with no primary-flight credit, so it ships in its own set, `config`, rather than joining `BUILTIN_PANELS`. A shell composes it when the airframe has the sensors, the builtin composition digest does not move for a panel nobody composed, and the panel set — one of the five release values — is unchanged. It declares no raster baseline, which the panel contract permits for a set the rasterizer does not yet cover.

## The sparse id has arrived

This is the first allocation to give a reserved id a variant, so assigned ids stop being contiguous at 0x0E. `GroupId::index` and the wire tag now genuinely disagree — 0x13 is slot 13, not 18 — which turns the test that could previously only *describe* that difference into one that proves it. That test was flagged in review of the v8 bump as unable to fail; it can now.

## Guardrails

- Each scale answers for its own sensor: a vehicle with a flap sensor and no trim sensor shows one reading and one set of dashes, not a trim pointer at neutral.
- A missing group dashes both scales.
- A selected detent does not move the sensed pointer.
- The set passes the same admission harness every other set does — nothing else would have exercised it.

## Pins

The ABI goldens are regenerated for the new group. `BUILTIN_SCENE_DIGEST`, the raster baselines and the composed-frame hashes are untouched, because the builtin set gained no panel and its panels emit the same bytes. ATT-01 re-records in its own commit.

Local gates: fmt, clippy `-D warnings`, `test --all-targets`, both evidence gates VALID, check-release-markers.


luofang34:
**REQUEST CHANGES** — the wire layout, the panel behaviour and the pins are all correct as written, but `check-structure.sh` is red on this branch and four of the behaviours this PR advertises can be deleted without a single test going red.

Reviewed at `76f76e4` against `origin/abi/v8-batch`. Every claim below was reproduced; mutations were reverted and the tree is clean.

---

## Blocking

### B1. `scripts/check-structure.sh` fails — CI gate red

```
$ bash scripts/check-structure.sh
FORBIDDEN: ./sets/indicate-instrument-panels/src/descriptors.rs has 501 lines (limit 500)
FORBIDDEN: ./crates/indicate-instrument-state/src/resolve/group_status.rs:66 function body has 86 lines (limit 80)
check-structure: FAILED
```

Both are introduced here — `git checkout origin/abi/v8-batch && bash scripts/check-structure.sh` prints `check-structure: OK`. It is a hard gate (`.github/workflows/ci.yml:49`).

`descriptors.rs` is one line over. Moving `CONFIG_DESCRIPTOR` / `CONFIG_SET` / `CONFIG_PANELS` / `draw_config_panel` into their own module fixes it and also fixes the module doc, which still reads "Built-in panel descriptors" while the file now holds a set that is deliberately not built in. `group_status` needs a split, not a `structure-function-baseline.tsv` entry (that file has one line in it today).

### B2. `airframe_fault` is fail-open — the whole function can be deleted

`crates/indicate-instrument-state/src/validate.rs:137-156`. Replace the body with `None`:

```rust
fn airframe_fault(_config: &crate::aircraft::AirframeConfig) -> Option<GroupFault> {
    None
}
```

`cargo test --locked --workspace --all-targets --no-fail-fast` → **all green, 0 failed**. Nothing exercises it: `grep -rn airframe crates sets tools --include="*.rs"` finds no test that constructs an out-of-range or non-finite ratio.

The CHANGELOG sells this behaviour explicitly ("A ratio outside the range its axis is defined over faults the group rather than being clamped — a clamped pointer would sit at a limit the airframe never reached"). The `in_unit` (0..1) versus `in_signed` (-1..1) split is the only thing that distinguishes a flap axis from a trim axis, and it is unpinned in both directions — swapping them would also stay green.

`abi/v8/tests.rs:390 director_faults_fail_the_group_per_branch` is the shape this wants: one `good` sample plus a table of per-branch bad ones.

### B3. The panel's `shows_value()` honesty gate is fail-open

`sets/indicate-instrument-panels/src/config.rs:44-54`. Drop `shown` and pass the values through:

```rust
flap_scale(scene, config.value.flap_ratio, config.value.flap_selected_ratio)?;
trim_scale(scene, config.value.elevator_trim_ratio)?;
```

Whole suite green. The behaviour it removes is not cosmetic — with the gate gone:

- `age_ms: Some(5000.0)` (past the 3000 ms fail threshold) draws `["FLAP", "50", "TRIM", "-20", "CFG"]` — live numerals *under* the red X.
- `flap_ratio: Some(1.8)` draws `"180"` with the pointer at `y = 70 + 220*1.8 = 466` in a 360-tall frame.

Why the three new tests miss it: `a_missing_group_dashes_both_scales` and `each_scale_answers_for_its_own_sensor` both distinguish *field* absence, never *status*. And the admission matrix cannot see it either — `withhold_group` resets `airframe` to `Default::default()`, so `resolve` hands the panel an all-`None` value in the withheld case; `Missing` and `Failed` are indistinguishable to every test in the tree. Nothing ever draws this panel with `Degraded` / `Stale` / `Failed` *and data present*.

Verified guardrail (fails with the gate removed, passes with it):

```rust
#[test]
fn a_failed_group_dashes_rather_than_drawing_under_the_x() {
    let t = texts(&panel(
        Some(AirframeConfig { flap_ratio: Some(0.5), elevator_trim_ratio: Some(-0.2),
                              ..AirframeConfig::default() }),
        Some(5000.0),
    ));
    assert_eq!(t.iter().filter(|s| *s == "---").count(), 2, "{t:?}");
}
```

### B4. The source-trust fold is fail-open, and `CONFIG_DESCRIPTOR` under-declares `required_groups`

`crates/indicate-instrument-state/src/resolve/group_status.rs:125-134` adds a comment asserting "Configuration folds source trust like the rest: a position reported by a source the state does not trust is not a position." Replace the fold with the trust-free wind form:

```rust
GroupId::AirframeConfig => {
    group_freshness(policy, state.airframe.data.is_some(), state.airframe.age_ms)
        .worst(fault_status(integrity.airframe))
}
```

Whole suite green. The cleanest demonstration uses a canonical state, no new fixture:

```rust
#[test]
fn an_unusable_source_dashes_both_scales() {
    let data = resolve(&indicate_instrument_descriptor::states::source_unusable(),
                       &FreshnessPolicy::default());
    let t = texts(&data);
    assert_eq!(t.iter().filter(|s| *s == "---").count(), 2, "{t:?}");
}
```

Real code: passes. With the fold removed: `["FLAP", "25", "TRIM", "-15"]`. `source_unusable`'s own docstring (`states.rs:216-217`) says "Panels must fail visibly, never render the numbers" — the admission matrix *draws* that state for the config panel (it is `typical()` with `quality = Unusable`, and `typical()` now carries airframe data thanks to this PR) but asserts nothing that separates dashes from numbers.

Separately and relatedly: `descriptors.rs:380` declares `required_groups: GroupSet::of(&[GroupId::AirframeConfig])`. `PFD_DESCRIPTOR` (`:79-87`), `HSI_DESCRIPTOR` (`:214-221`) and `TEMPLATE_DESCRIPTOR` all also declare `GroupId::Trust`, and the template's comment states the rule: "Declare the groups the readout DERIVES from … `Trust` decides whether it may be shown at all". `MONITOR_DESCRIPTOR` omits it correctly, because `MonitorText` genuinely does not fold trust. `AirframeConfig` does. Adding `GroupId::Trust` and moving `assert_eq!(report.cases, 20)` to `30` passes — I ran it.

But be aware, and this is the more interesting half: **adding `Trust` to `required_groups` does not by itself close the hole.** I tried the combination (Trust declared + trust fold removed) and admission still passed, because `check_provenance` only refuses a run claiming the *withheld* group, and these runs claim `AirframeConfig`. So the declaration is a correctness/convention fix; the guardrail has to be the unit test above.

### B5. The golden fixture cannot tell `flap_ratio` from `flap_selected_ratio`

`crates/indicate-instrument-state/src/abi/v8/fixtures.rs:157-165` gives both fields `Some(0.5)`. Swap offsets 0 and 4 in **both** `decode_airframe` and `encode_airframe` (`abi/v8/stamped.rs:394-424`) — an internally consistent codec that violates the layout table in `group_id.rs:49` and `stamped.rs:23` — and the whole suite is green, with `state-abi-v8.full.hex` byte-identical.

Elevator/aileron are safe (`-0.2` vs `0.05`, so a swap moves the hex and `full_frame_is_pinned` catches it). Only this one pair is blind, and it is the pair the group exists to keep apart ("sensed and selected are never conflated"). A producer written against the doc table and this consumer would silently disagree about which mark is the pilot's selection.

Fix: give the two distinct values in `fixtures::full()` (e.g. `flap_ratio: Some(0.25)`, `flap_selected_ratio: Some(0.5)` — which also makes the fixture show the in-transit case the panel is designed around) and re-pin the hex with `cargo xtask gen-state-fixture`.

---

## Non-blocking

**Docs this change falsifies.**

- `group_id.rs:158-163` — `index()`'s doc still says "Assigned ids are contiguous today, so `tag - 1` would give the same answers; they stop being contiguous at the first variant above 0x0D." 0x13 *is* that variant, as of this PR. The CHANGELOG bullet you added says the opposite of this docstring.
- `group_id.rs:21` — registry row `| 0x01–0x0D | the variants below |`. 0x13 is now also a variant.
- `abi/v8.rs:48-49` — "Landed so far: the Air group carries `tas_mps` …". You updated the equivalent CHANGELOG sentence but not this one.
- `group_id/tests.rs:31-33` — "0x0E–0x11 are planned and 0x12–0x15 are allocated with their layouts fixed, but neither has a variant. Until one lands …". 0x13 has one; line 34 had to be edited around it and the comment was left behind. This is exactly the stale-prose-after-a-mechanical-sweep case.
- `CHANGELOG.md:88-91` — "**The batch allocates four group ids and three field appends. This release implements the Air append only.**" Contradicted by the bullet you added ten lines above it.

**Comment rules.**

- `group_id/tests.rs:48-52` and `:58` — "The sparse allocation has arrived", "this now proves it rather than describing it", "the match and the arithmetic now disagree". Temporal phrasing and refactor history; the *why* survives without them ("`index()` is a match because tag arithmetic would answer 18 for a 14-slot table").
- `descriptors.rs:389-391` — "the region has to hold the dashes as well as the readings". The region family only counts runs carrying a claim (`admission/regions.rs:80-85`), and the dashes deliberately carry none, so this states a constraint nothing imposes.
- `descriptors.rs:1` — "Built-in panel descriptors" no longer describes the file's contents (see B1).

**Coverage.**

- No dedicated 0x13 wire tests: no short-payload rejection (`a_known_group_below_its_minimum_length_fails_that_group` covers Air only), no per-field NaN-absent assertion. `min_len = 24` *is* pinned indirectly — I set it to 20 and `fixtures_round_trip_bit_exactly` reddens through `extended_groups` — so this is coverage, not a defect.
- `group_id/tests.rs:85-95 withholding_a_stamped_group_resolves_missing` doesn't list `AirframeConfig`. One enum literal; the crate currently relies on a test in a different crate for this.

**Smaller.**

- `airframe_fault` returns `GroupFault::NonFinite` for a finite out-of-range ratio, and that variant is documented as "A mandatory value is NaN or infinite". `director_fault` sets the same precedent, so this is consistency rather than a new wrong — but the batch may be the moment to add `OutOfRange`.
- Match-arm placement: `AirframeConfig` sits between `Dynamics` and `MonitorText` in `min_len` / `decode_group` / `encode_group`, and between `Trust` and `Altitude` in `withhold_group`, while `GroupId::ALL` puts it last. Every other arm follows wire order.
- `PanelData` is now 25 fields; #76 makes it 26. Limit is 30.
- New display function with no row in `docs/instruments/intended-functions.md` and no `AIR-*` id, where the comparable monitor panel cites AIR-IN-014 in both its module doc and its descriptor. No gate enforces this — raising it as a question, not a rule.
- STE: the two new CHANGELOG bullets are each one sentence carrying two clauses across an em-dash; "now genuinely disagree" adds an intensifier; and "the test that could only describe that difference proves it" names no test and encodes what the test used to be.

---

## What I ran

All green: `cargo fmt --all --check`; `cargo clippy --locked --all-targets --workspace -- -D warnings`; `cargo test --locked --workspace --all-targets --no-fail-fast` (0 failed); `RUSTDOCFLAGS="-D missing_docs -D rustdoc::broken_intra_doc_links" cargo doc --locked --workspace --no-deps`; `cargo build --locked --release --workspace`; the twelve-crate `cargo check --target thumbv7em-none-eabihf`; `check-release-manifest`, `check-release-markers` (0.5.0; ABI 8, scene 1, corpus 4, panels pfd/hsi/monitor), `check-instrument-requirements` (73), `check-certification-claims`, `check-standards-registry`, `trace-report`, `detect-target`; both `evidence-gate` invocations → `verdict: VALID`.

Red: `check-structure` (B1).

**Pins verified against what the tree computes.** `cargo run -q -p instrument-bench` reports scene digest `add8e694…` *matches pin* and screen-composition digest `197f70d0…` *matches pin*, 280 cases / 196 counted warnings, 15 renders — so the composition digest, the screen-composition digest, the raster baselines and `release-manifest.json` correctly did **not** move, because `CONFIG_SET` sits outside `BUILTIN_PANELS`. The v8 golden hex is right: the header count goes `0d`→`0e` and the appended group is `13 18 00 | 0000003f 0000003f cdcc4cbe cdcc4c3d 0000c07f 0000a042` = 0.5 / 0.5 / -0.2 / 0.05 / NaN / 80.0, tag last in ascending order, 24 bytes matching `min_len`. Re-recorded ATT-01 counts are what the tree produces — I ran each crate on its own and got exactly 86 (panels), 181 (state), 98 (raster). Evidence baseline `commit 0bab5c5` / `tree 21225af1…` agrees with `git rev-parse 0bab5c5^{tree}`. The admission case count 20 = 5 canonical states × (1 fed + 1 withheld) × (quiet, alerted) is right for the declaration as written (30 once `Trust` is added per B4).

**What I tried that did not break** — these are real guardrails, not decoration:

- `withhold_group`'s `AirframeConfig` arm made a no-op → caught by `the_config_set_passes_admission`.
- `required_groups` emptied → caught (`RegionGroupNotRequired`). `required_groups` **and** `group_regions` both emptied → still caught (`ForeignClaim`), so the panel genuinely cannot escape declaring the group it draws from.
- The detent made to move the sensed pointer (`selected.unwrap_or(ratio)`) → caught by `a_selected_detent_does_not_move_the_sensed_pointer`.
- The flap dashes replaced by a pointer at 0 → caught twice (`a_missing_group_dashes_both_scales` and admission).
- `min_len(AirframeConfig)` 24 → 20 → caught by the round-trip and the golden frame.

The honesty claims in the description hold in the drawing code, not only in the tests: `flap_scale` returns early with red `"---"` before it can place a pointer (`config.rs:102-112`), the detent is its own cyan triangle on the opposite side of the scale and the sensed pointer's `y` is computed from `ratio` alone (`:113-134`), and the dashes carry no provenance claim, which is the correct choice — a `"---"` tagged `AirframeConfig` would be refused in the very case it exists for. `Degraded` and `Stale` show the value with a `CFG` flag, `Failed` blanks under a red X, `Missing` blanks silently — the right split. The problem is only that nothing holds it there.

---

## Sibling coordination (#76 BearingPointers 0x12, #72 ias_trend)

The pins in **this** PR are internally consistent, and nothing here re-pins a value that did not move. Nothing needs recomputing until a sibling lands.

When one does, expect these to conflict textually rather than semantically: `GroupId::COUNT` / `ALL` / `from_u8` / `index`, `min_len`, `decode_group`, `encode_group`, `withhold_group`, `PanelData`, `StateIntegrity`, `group_status`, and `state-abi-v8.full.hex` (all three PRs move it — #72 grows Dynamics, #76 adds 0x12). Whichever merges second regenerates the fixture and re-records the ATT-01 artifacts.

Two spots need a real edit rather than a merge resolution when #76 lands:

- `group_id/tests.rs:34` — `(0x0Eu8..=0x12).chain(0x14..=0x15)` must narrow to `(0x0Eu8..=0x11).chain(0x14..=0x15)`, and the comment above it (already wrong today, see above) needs rewriting rather than patching.
- `group_id/tests.rs:70` — `const NEXT_ALLOCATED: u8 = 0x12;` becomes false and has to move to `0x14`. The test is designed to fail loudly here, which is the right design; just flagging that it is a deliberate breakage and not a merge artefact.

`index_is_not_wire_tag_arithmetic` survives #76 unchanged, since 0x13 stays the highest tag.

luofang34:
All five blocking findings fixed in `44d1122` (evidence re-recorded in `371aa9b`). Every mutation you named now reddens the suite — I reran each one to confirm rather than assuming.

**1. `check-structure.sh` red.** `descriptors.rs` was 501 lines and `group_status` had an 86-line body. The measured criticality bands moved to their own module beside the descriptors — they are evidence the harness re-derives, not design, so they read better apart from the panels they describe — and the attitude and kinematics status arms became named functions, which is also where their "the group is no better than its worst member" reasoning belongs. `check-structure.sh` prints OK.

**2. `airframe_fault` fail-open.** You were right that nothing constructed an out-of-range ratio. `out_of_range_configuration_faults_the_group_per_field` now walks every field against both bounds plus an infinity, so a check that only ever fired on the flap ratio cannot pass. Verified: `if true { return None; }` at the top of `airframe_fault` reddens it.

**3. The `shows_value()` gate fail-open.** This was the sharpest observation in the review — that `withhold_group` zeroes the data, so `Missing` and `Failed` were indistinguishable to every existing test, and the dash tests only varied *field* absence. The two new cases carry a value **and** a status that refuses it, which is the only shape that proves the gate reads the status. `a_stale_group_draws_no_numerals_even_though_it_carries_values` asserts up front that the fixture is actually refused by the policy, so it cannot quietly become vacuous. Verified: replacing the gate with `true` reddens both.

**4. The source-trust fold fail-open.** `a_source_that_declares_itself_unusable_draws_no_numerals` builds the canonical unusable posture directly and asserts no numeral is drawn. Verified: replacing the fold with the trust-free wind form reddens it. I also added `GroupId::Trust` to `required_groups` (case count 20 → 30) — and I took your point that this alone does not close the hole, which is why the behavioural test is the fix and the declaration is the completeness.

**5. The golden hex could not tell sensed from selected.** Both were `0.5`. They are now `0.25` and `1.0`, with the reason on the fixture. Verified: swapping offsets 0 and 4 in both encoder and decoder now reddens `full_frame_is_pinned`.

Full local CI green: fmt, clippy `-D warnings`, `test --all-targets --no-fail-fast`, doc gate (which caught an intra-doc link broken by the module split), release build, every `check-*.sh`, and the evidence gate `VALID` against re-recorded artifacts (89/181/98).

One note on merge order: this PR, #76 (BearingPointers 0x12), #74 (nav scale), #72 (ias\_trend), and the autoflight PR still to come all branch from the same point and re-pin the same globals, and this one and #76 each add a `GroupId` variant. They have to land serially, with the pins recomputed by running the tests after each rebase — not resolved with a strategy flag.

luofang34:
**APPROVE WITH NITS** — all five blocking findings are closed, and I confirmed each by re-running my own mutation rather than reading the tests. Nothing blocking survives.

Verified at `371aa9b`. Every mutation below was applied, run, and reverted; `git status --short` and `git diff --stat` are empty at the end.

---

## Per finding

### B1 — `check-structure.sh` red · **CLOSED**

```
$ bash scripts/check-structure.sh
check-structure: OK
```

Both violations gone. `descriptors.rs` shed the bands into `descriptors/criticality_bands.rs` — which is the right shape (`foo.rs` + `foo/`, no `mod.rs`) and the right *reason*: those bands are measured evidence, not authored design, and the new module doc says so. `attitude_status` and `kinematics_status` are a real decomposition rather than a line-count trim; both carry a why-comment, and `group_status` is now a legible dispatch table.

Worth noting you split rather than reaching for `structure-function-baseline.tsv`. That file still has exactly one line in it.

### B2 — `airframe_fault` fail-open · **CLOSED**

Gutted the function (`-> None`), ran `cargo test --locked --workspace --all-targets --no-fail-fast`:

```
test config::tests::out_of_range_configuration_faults_the_group_per_field ... FAILED
```

I also tried the subtler mutation I care more about — widening the flap axis from `in_unit` to `in_signed`, so a flap ratio of −0.1 becomes legal while every other field's bound stays correct:

```
must fault: AirframeConfig { flap_ratio: Some(-0.1), flap_selected_ratio: Some(0.5), ... }
```

Caught. The `good` fixture carrying `elevator_trim_ratio: Some(-0.2)` closes the mirror direction: swapping trim to `in_unit` reddens the up-front `is_none()` assertion. So the unit-versus-signed distinction — the part that actually keeps a trim pointer on its scale — is pinned in both directions, not just the existence of a check.

### B3 — the `shows_value()` gate fail-open · **CLOSED**

Dropped `shown` and passed the values straight through. Two tests redden:

```
test config::tests::a_stale_group_draws_no_numerals_even_though_it_carries_values ... FAILED
test config::tests::a_source_that_declares_itself_unusable_draws_no_numerals ... FAILED
```

The vacuity guard is the part that makes this durable:

```rust
assert!(!stale.airframe.status.shows_value(),
        "the fixture must be refused by the policy, or this proves nothing");
```

That is exactly the failure mode the original tests had — a fixture that quietly stops exercising the branch — and it is now impossible for this one to rot into a tautology without saying so.

### B4 — the source-trust fold fail-open · **CLOSED**

Replaced the fold with the trust-free wind form:

```
test config::tests::a_source_that_declares_itself_unusable_draws_no_numerals ... FAILED
```

Correct call putting the weight on the behavioural test. `GroupId::Trust` in `required_groups` is the right declaration on its own merits — it matches PFD, HSI and the template, and it is what the group's status actually folds — but as we established it could never have caught this, and the commit does not pretend otherwise.

### B5 — golden hex blind to sensed-versus-selected · **CLOSED**

Swapped offsets 0 and 4 in both `encode_airframe` and `decode_airframe` — the internally-consistent codec that violates the layout table:

```
test abi::v8::fixture_tests::full_frame_is_pinned ... FAILED
```

`0.25` / `1.0` is a better pick than merely-distinct values: `0000803e` and `0000803f` differ in more than one byte, and the pair also happens to encode the in-transit case the panel's detent mark exists for, so the fixture now illustrates the feature as well as pinning it. The comment explaining why they must differ is the kind that earns its place.

---

## The three confirmations you asked for

**Intra-doc links.** `cargo doc` with `-D missing_docs -D rustdoc::broken_intra_doc_links` is clean — I forced it with `cargo clean -p indicate-instrument-panels --doc` first so the result is not a cached pass. The repair is visible: the moved doc uses `[`crate::BUILTIN_PANELS`]` (absolute, resolves through the crate-root re-export), while the two `[`BUILTIN_PANELS`]` links that stayed behind in `descriptors.rs` still resolve in their own module scope.

**Case count 30.** Load-bearing, not a number that happens to match. `case_matrix` builds `1 + |required_groups|` withholdings per state, × 5 canonical states × 2 alert variants × 1 canonical frame. I reverted `required_groups` to the single group and the assertion caught it exactly:

```
assertion `left == right` failed
  left: 20
 right: 30
```

**ATT-01 re-record.** Internally consistent on every field I can check:

| check | result |
|---|---|
| `CFG-WORKTREE tree` `68bc13cd…` | `git rev-parse 44d1122^{tree}` agrees |
| `config-digest` / `run-id`, all 3 artifacts + all 3 graph nodes | all `44d1122`, none left at `0bab5c5` |
| `output-digest` × 3 | `shasum -a 256` of each artifact matches |
| counts 89 / 181 / 98 | per-crate runs give exactly 89, 181, 98 |
| evidence gate, both invocations | `verdict: VALID`, 30 nodes / 73 edges / 0 exceptions |

The panels crate moved 86 → 89 for the three new tests; state and raster are unmoved, correctly, since nothing was added to them.

**Rest of the battery, all green:** `fmt --check`; `clippy --locked --all-targets --workspace -- -D warnings`; `test --locked --workspace --all-targets --no-fail-fast` (0 failed); `build --locked --release --workspace`; the twelve-crate `cargo check --target thumbv7em-none-eabihf`; `check-release-manifest`, `check-release-markers` (0.5.0; ABI 8, scene 1, corpus 4, panels pfd/hsi/monitor), `check-instrument-requirements`, `check-certification-claims`, `check-standards-registry`, `trace-report`, `detect-target`. `instrument-bench` still reports scene digest `add8e694…` *matches pin* and screen-composition digest `197f70d0…` *matches pin* — adding `Trust` to a non-builtin descriptor correctly moved nothing in `BUILTIN_PANELS`.

---

## Non-blocking

**One new fail-open, in the same family — mine to have missed in round one, not a regression.** Delete the `Stale | Degraded` arm of the annunciation match in `config.rs:62-64`, so a stale flap reading paints exactly like a fresh one:

```rust
SignalStatus::Stale | SignalStatus::Degraded
| SignalStatus::Missing | SignalStatus::Valid => {}
```

Whole suite green. The `CFG` flag carries no digits, so provenance does not reach it, and no test looks for it. This is the *shown-but-qualified* half of the same honesty rule the new tests pin the *refused* half of.

**Related, and the reason I noticed:** `a_stale_group_draws_no_numerals_even_though_it_carries_values` does not test a stale group. `age_ms: 5000.0` against the default policy (stale 750, fail 3000) resolves **`Failed`** — I printed it. A genuinely stale group is the opposite:

```
age 5000 -> Failed
age 1000 -> Stale
age 1000 texts -> ["FLAP", "50", "TRIM", "-20", "CFG"]
```

Which is correct behaviour — `shows_value()` is true for `Stale`, and showing the value under a flag is the design. But the name and the doc comment ("A group whose sample is too old", "`Missing` and `Stale` are different situations") assert the reverse for the one status they name, and a future reader could "fix" the panel to match the test name. Renaming to `a_failed_group_…` and adding the age-1000 case as its own positive test would close both this and the flag gap in one move.

**Round-one non-blocking items, all still open** — unchanged and untouched, which is fine given how they were classified; listing them so none is silently dropped:

- `group_id.rs:155` — "ids are contiguous today … they stop being contiguous at the first variant above 0x0D". 0x13 is that variant.
- `group_id.rs:19` — `| 0x01–0x0D | the variants below |`.
- `abi/v8.rs:48` — "Landed so far: the Air group carries `tas_mps` …", still omitting this group.
- `group_id/tests.rs:32` — "neither has a variant. Until one lands …", contradicted by line 34 having to skip 0x13.
- `group_id/tests.rs:48` — "The sparse allocation has arrived" / "now proves it rather than describing it".
- `CHANGELOG.md:89` — "This release implements the Air append only."
- `GroupFault::NonFinite` returned for a finite out-of-range ratio; `AirframeConfig` match arms out of wire order in `min_len` / `decode_group` / `encode_group` / `withhold_group`; `withholding_a_stamped_group_resolves_missing` still omits `AirframeConfig`; no `intended-functions.md` row for the new display function.

**Sibling coordination is unchanged from my last comment**, with one addition: `CONFIG_DESCRIPTOR` now declares `Trust`, so if #76 or #72 changes what the trust fold does, the config panel's case count moves with it and `assert_eq!(report.cases, 30)` will say so.

Good fixes. The three new tests are the kind that stay honest — the vacuity guard in particular.

luofang34:
Took both of the follow-up nits in `db2cf94` — they were the same finding twice over, and the second one is the more embarrassing of the two.

**The `Stale | Degraded` arm was fail-open.** You are right, and it belongs to the same family as the original five. `a_stale_group_keeps_its_readings_under_a_flag` now pins it; verified it reddens when the arm is deleted.

**The misnamed test.** Also right: age 5000 against a 3000 ms fail threshold is a *failed* group, so the test asserted the reverse of the behaviour for the status it named. Renamed to `a_failed_group_draws_no_numerals_even_though_it_carries_values`, and the genuine stale case (age 1000, between the two thresholds) is now its own test — which asserts `status == Stale` before asserting anything about paint, so it cannot drift into testing a different tier the way the first one did.

Worth saying explicitly since the two now read as opposites: a failed configuration draws nothing, a stale one keeps its readings under a flag. That is deliberate. A configuration that stopped updating is still the last configuration the airframe was in, and blanking it would be a bigger claim than the staleness supports — which is the same reasoning the monitor panel uses for its advisory channel.

The round-one documentation nits are noted and I have not dropped them; they are about where the labels sit rather than about behaviour, and I would rather move them once and deliberately.

Suite green: panels 90, state 181, raster 98; fmt, clippy `-D warnings`, structure gate, and the full workspace run all pass. The ATT-01 panel count moves 89 → 90 with the new test, so the evidence artifacts get a re-record before this merges.

PR #76, opened 2026-08-18
A bearing pointer answers a question the CDI cannot (#53)

### LINKED ISSUE #53: The HSI rose carries no bearing pointers

luofang34:
Closes #53.

The CDI says how far off a selected course the aircraft is. It cannot say where a second station lies, because a course selector is on one receiver at a time. Two bearing pointers follow their own receivers, so the rose carries the stations the selector is not on.

## The wire

Group `0x12 BearingPointers`, stamped, 20 bytes: two pointers, each a source enum, a heading reference, a validity byte, one pad byte, and an `f32` bearing in radians, then the trailing `age_ms`. This is the layout the registry table in `group_id.rs` allocated for the batch.

The reference travels with the bearing rather than being inherited from the heading group. A receiver that reports magnetic bearings beside an attitude source that reports true heading is a normal pairing, so each pointer converts into the rose's reference on its own. A pointer whose north the rose cannot resolve loses its angle and is not drawn: rotating an unconverted bearing onto the rose would put a needle at an angle nobody measured, which is what the conversion exists to prevent.

## The drawing

Two needles in distinct forms — one line against two, a plain head against a split one — because a pilot tells them apart by shape rather than by reading a label. Colour follows the source, so a receiver keeps one colour wherever it appears on the panel. The pointers carry their own gate rather than the CDI's: a needle draws whether or not a course is selected.

## The guardrails

- `hsi/bearing/tests.rs` pins what the pointers refuse (no source, declared invalid, unconvertible reference) and that the panel gains exactly the needles' marks when the receivers become usable.
- The HSI descriptor declares `BearingPointers` among the groups it consumes, so the admission withholding matrix drives the needles through the same honest-status checks as every other group the panel draws from.
- A `bearing-split-references` extreme state puts a convertible pointer beside an unconvertible one in one frame, so the harness and the raster baselines carry the mixed case.
- `withholding_a_stamped_group_resolves_missing` now walks `GroupId::ALL` and sorts the lanes with an exhaustive match instead of a hand-written list, so no `withhold_group` arm can ship untested. **The list had already drifted: `FlightDirector` was missing from it.**
- `group_id/tests.rs` now proves `index()` is a match rather than tag arithmetic from the registry itself, because the assigned tags are no longer contiguous.

## Re-pinned

Composition digest, HSI raster baseline, side-by-side composed frame, screen-composition digest, the v8 full golden frame, and the admission case count (280 → 312: the HSI gains a third extreme state and a ninth withholding case, each drawn quiet and saturated). The evidence gate stays VALID — no recorded source moved.

`resolve.rs` reached its 500-line limit, so `wind_signal` moved into the `resolve/*_signal.rs` module-per-group pattern it already names, and the new bearing resolution landed there rather than in the parent.

## Merge order

This PR, the `AirframeConfig` PR (#57), and the `ias_trend` PR (#51) all branch from the same `abi/v8-batch` point and all re-pin the same globals: the composition digest, the screen-composition digest, the admission case count, the v8 golden frames, `release-manifest.json`, and the CHANGELOG table. This PR and #57 each add a `GroupId` variant, so `COUNT` must reach 15 once both land.

**They cannot merge independently.** Whichever merges second is stale on every pin. Merge one, rebase the next onto the updated batch branch, re-run the suite to recompute the pins by running the tests (not by resolving conflicts with a strategy flag), full CI, then merge. The pins in this PR are valid only against the current batch head.

Local CI: `fmt --check`, `clippy --all-targets -D warnings`, `test --all-targets --no-fail-fast`, `doc -D missing_docs`, `rustdoc -D broken_intra_doc_links`, `build --release`, and every `scripts/check-*.sh` — all green.


luofang34:
**REQUEST CHANGES** — the feature is correct and the drawing half is genuinely guarded, but the two fail-closed properties this PR introduces (`bearings_fault`, the bearing source enum's wire coding) have no guardrail at all: I gutted each and the full workspace suite stayed green.

Reviewed at `f375765` against `origin/abi/v8-batch`, in a clean detached worktree.

---

## Blocking

### 1. `bearings_fault` is dead weight as far as CI is concerned

`crates/indicate-instrument-state/src/validate.rs:138`

Repro — insert at the top of `bearings_fault`:

```rust
fn bearings_fault(pointers: &crate::aircraft::BearingPointers) -> Option<GroupFault> {
    if true { return None; }
    ...
```

`cargo test --locked --workspace --all-targets --no-fail-fast` → **0 failures.** Nothing in the tree exercises the function's docstring ("A pointer whose source this build cannot name, or whose north it cannot resolve, fails the group"). It could be deleted and every gate stays green.

The behaviour is real — I probed it: an `Unknown` source on `first` gives `group=Failed rose0=Failed rose1=Failed heads=0`. It is just untested. Precedent for what the test should look like is in the same batch, same file: `abi/v8/tests.rs:390 director_faults_fail_the_group_per_branch`.

### 2. The bearing source byte's fail-closed coding is untested in both directions

`crates/indicate-instrument-state/src/abi/v8/stamped.rs:403` and `:420`

Repro A (decode only), `stamped.rs:403`:

```rust
-            _ => crate::aircraft::NavSource::Unknown,
+            _ => crate::aircraft::NavSource::None,
```

Full workspace suite → **0 failures.** An out-of-range source byte would silently decode as "no receiver selected" — the benign-variant laundering that the `abi/v8.rs` module doc forbids by name: *"a wire value outside the known set decodes to each type's `Unknown`, never to a benign variant (VAL-01)"*.

Repro B (encode), `stamped.rs:420`: `NavSource::Unknown => 255` → `=> 0`. Also green (I ran A alone and A+B together; both green).

The golden fixture only carries `Nav1`/`Nav2`, so it pins nothing here. `abi/v8/tests.rs:244 unknown_wire_enum_values_decode_fail_safe_not_benign` and `:269 unknown_variants_encode_as_255_and_round_trip_explicitly` cover the Nav group only — the bearings group needs the same two assertions (a `one_group_frame(0x12, …)` with source byte 7, and an encode of an `Unknown`-source pointer).

Under CONTRIBUTING's "every PR lands the fix **and** the guardrail that prevents its regression in the same PR", 1 and 2 are the blocking part. Everything below is non-blocking.

---

## What I tried to break and could not

- **Draw gate.** Removed `!bearing.status.shows_value()` from `hsi/bearing.rs:47` → `hsi::bearing::tests::a_bearing_that_cannot_convert_draws_no_needle` FAILED, plus `the_builtin_scene_digest_is_pinned`, `composing_the_builtin_set_reproduces_the_pinned_digest`, `regrouping_panels_without_reordering_them_is_digest_neutral`.
- **Reference-conversion claim, in the code path (not just the test).** In `resolve/bearings_signal.rs:31`, replaced `pointer.reference` with `rose` so conversion always succeeds → the same 4 tests FAILED. So the claim is bound by both the unit test and the pinned scene digest, not by assertion alone.
- **The extreme state does what its docstring says.** Probed all three HSI extremes: `reciprocal-course heads=3`, `track-up heads=3`, `bearing-split-references group=Valid rose0=Valid rose1=Failed heads=1` — one needle drawn and one removed in the same frame, as claimed.
- **`withhold_group`'s new arm** is bound: `withholding_a_stamped_group_resolves_missing` is now exhaustive over `GroupId::ALL` with a `match` that fails to compile on a new variant, and it asserts `groups.status(group) == Missing` per group.
- **`min_len(BearingPointers) = 20`** is pinned in both directions by the golden frame: a smaller value trips `assert_eq!(report.extended_groups, 0)` in `fixture_tests.rs`, a larger one trips `GroupTruncated`. Not a gap.
- **Wire bytes, decoded by hand** from `state-abi-v8.full.hex`: `12 1400 | 02 02 01 00 9a99993f | 03 02 01 00 33338340 | 0000a042` = len 20; Nav1/SimLocalTrue/valid/1.2; Nav2/SimLocalTrue/valid/4.1; age 80.0. Group-count byte 0x0d → 0x0e. Encoder, decoder, `min_len`, and the `stamped.rs:23` layout row all agree.
- **Degraded/Stale (asked for explicitly).** Measured: `group=Stale → rose0=Stale, heads=3`; `group=Degraded → rose0=Degraded, heads=3`. Both needles draw at full strength with no flag, no colour change, no annunciation — unlike the CDI, which gets a `NAV` flag at `hsi.rs:118` for `status != Valid`. Consistent with the heading bug, so probably intended, but nothing in the PR says so. Worth one sentence in `bearing.rs`'s module doc.

## Re-pinning — every moved value recomputed, none hand-typed

- `cargo run --locked -q -p instrument-bench`: `scene digest b15a5d86… (matches pin)`, `screen-composition digest 59d8086a… (matches pin)`, `admission: 312 cases pass, 196 counted warnings`.
- `bash scripts/check-release-manifest.sh`: OK — it regenerates `release-manifest.json` from the tree and diffs, so `compositionDigest`, `screenComposition.digest`, and the HSI `rasterBaselines` entry are generator output, not typed.
- HSI raster baseline `fe3ab150…` is re-rendered and compared in `crates/indicate-instrument-raster/src/raster/tests/frame_hashes.rs:113`.
- `SIDE_BY_SIDE_HASH` and `BUILTIN_SCENE_DIGEST` are asserted by tests I saw go red under gutting.
- `BENCH_COMPOSITION_DIGEST` is recomputed and refused on mismatch at `tools/xtask/src/manifest/document.rs:107`.

**Admission arithmetic (312) checks out against `case_matrix`**: PFD 8 required groups + 3 extremes → 8 states × 9 = 72; HSI 8 required (incl. the new `BearingPointers`) + 3 extremes → 8 × 9 = 72; monitor 1 required + 1 extreme → 6 × 2 = 12. (72+72+12) × 2 alert states × 1 canonical frame = **312**. The comment's wording matches the code.

## Gates — all green, run here

`fmt --all --check`; `clippy --locked --all-targets --workspace -- -D warnings`; `test --locked --workspace --all-targets --no-fail-fast` (0 failures); `RUSTDOCFLAGS="-D missing_docs -D rustdoc::broken_intra_doc_links" cargo doc`; `build --locked --release --workspace`; `check-structure` (**ran, did not skip**), `check-release-manifest`, `check-release-markers`, `check-instrument-requirements`, `check-certification-claims`, `check-standards-registry`, `trace-report`, `detect-target`; both `evidence-gate` invocations → `verdict: VALID`; and CONTRIBUTING's `cargo check --target thumbv7em-none-eabihf` over the twelve no_std crates.

---

## Non-blocking

1. **`abi/v8.rs:48` is now false.** *"Landed so far: the Air group carries `tas_mps` … "* — this PR lands 0x12 into the same version. The CHANGELOG was updated to say "true airspeed on the Air group, and the bearing-pointer group"; the module doc a consumer actually reads was not.

2. **The two registry tables in `group_id.rs` now disagree about 0x12.** Line 19 still reads `| 0x01–0x0D | the variants below |`, which no longer covers the variant set. Line 48's "New groups" row still describes 0x12 as an unlanded allocation.

3. **`group_id.rs:48` misstates the shipped source coding.** The row says `a source enum (None/Nav1/Nav2/Gps, 255 fail-closed Unknown)`. the `abi/v8.rs` module doc says enum bytes "count from zero in declaration order", so a third-party producer reading only this row writes `Nav1=1, Nav2=2, Gps=3`. The shipped codec reuses `NavSource`'s declaration order: `None=0, Gps=1, Nav1=2, Nav2=3`. Reusing `NavSource` is the right call — the row is what needs fixing. The same row also lists the fields as `source, bearing, reference, validity`; the shipped layout is `source, ref, valid, pad, bearing`.

4. **`decode_pointer`/`encode_pointer` copy the `NavSource` ↔ byte mapping out of `decode_nav`/`encode_nav`** (`stamped.rs:398-405` vs `:151-157`, `:414-421` vs `:189-195`). Two copies of a fail-closed wire mapping can drift. Extracting `nav_source_from_u8`/`nav_source_to_u8` would fix finding 2 for free — the existing Nav tests would then cover both call sites.

5. **`hsi/bearing.rs:22,24`** — `HEAD_R` and `TAIL_R` are both `ROSE_R - 22.0` but their doc comments describe different geometry ("inside the tick ring" / "on the far side of the rose"). Either they are the same distance and should say so, or one is wrong.

6. **`bearings_rose_rad: [Sig<f32>; 2]` is positionally coupled to `BearingPointers { first, second }`** across a crate boundary. An inversion is silent — `two_usable_pointers_draw_two_distinct_needles` would still see 3 heads. `colour_follows_the_source_not_the_needle` catches it only because the two sources differ; consider making that intent explicit.

7. **`GroupSet` (asked for explicitly): real latent problem, not this PR's.** `crates/indicate-instrument-descriptor/src/group_set.rs:20,28` do `1 << group as u32` on a `u32`. 0x12 (18) is fine; 0x13–0x15 (19–21) are fine. The first break is tag **0x20**: in `const` context that is a compile error, in a debug build a shift-overflow panic, and in release the shift masks to 5 bits so `1 << 0x20` sets bit 0 — silently aliasing **Attitude**. The registry reserves `0x16–0xDF`, so this is reachable by design. Follow-up issue against `group_set.rs` (widen to a `[u32; N]`/`u128`, or index by `GroupId::index()` and say so in `bits()`'s "wasm/FFI encoding" doc), not a change this PR should carry.

8. **Scope of the `wind_signal` move: justified, not creep.** `origin/abi/v8-batch:resolve.rs` is 486 lines; this PR adds two documented `PanelData` fields, the call site, and two module lines. Leaving `wind_signal` in place lands it at ~503, over the 500-line limit. The move is the size rule firing. `resolve.rs` is now 481. `draw_hsi` is 77 lines against the 80-line limit — worth knowing before the next HSI feature.

9. **Comment/doc-language sweep is clean on the Rust side.** No temporal phrasing, no PR/issue numbers, no line counts anywhere in the `.rs` diff; the deleted narrative in `group_id/tests.rs` was replaced with a real assertion (`GroupId::ALL[COUNT-1]`, `0x12-1 = 17 >= 14`) rather than left as a fragment. On the CHANGELOG: I could find no ASD-STE100 rule in the tree (no `AGENTS.md`/`CLAUDE.md` present, no STE reference anywhere), so I judged it against the register of the surrounding entries. The last sentence of the 0x12 bullet runs ~60 words with a colon-joined compound and a passive ("rather than being inherited from"); the neighbouring bullets run much shorter. Splitting it after "…travels with the bearing" would match them. Style only.

luofang34:
Fixed both blocking findings in 6393777.

**1. `bearings_fault` untested** — added `bearing_faults_fail_the_group_per_branch`, which walks each fault branch (unknown source, unknown reference, non-finite bearing) against each pointer slot, so a fault that only ever fired on the first pointer's source cannot pass. It also pins the complementary case you'd want: a pointer its own source calls unusable carries no claim about its bearing, so a NaN there does *not* fault the group. Verified against your mutation: `if true { return None; }` at the top of `bearings_fault` now reddens the suite.

**2. Source byte coding untested in both directions** — added `unknown_bearing_source_bytes_round_trip_as_the_fail_closed_sentinel` (reads the encoded bytes, not just the decoded struct) and `unassigned_bearing_source_bytes_decode_to_unknown` (corrupts one pointer's bytes and leaves the other, so a decoder that lost the distinction cannot pass by failing everything). Verified: changing decode's `_ => Unknown` to `_ => None` reddens both; changing encode's `Unknown => 255` to `=> 0` reddens the first.

Also took two of the non-blocking items, since both would mislead a third-party producer:

- `group_id.rs` said the source enum is coded `None/Nav1/Nav2/Gps`. That is not the shipped `NavSource` coding — someone reading only that row would put `Gps` where `Nav1` belongs. The row now states the byte values.
- The v8 module's "Landed so far" sentence is gone rather than corrected. It was a running commentary that went stale the moment a second group landed; the allocation table and `min_len` are where the answer lives, and the golden frames check those against each other.

`abi/v8/tests.rs` crossed its 500-line limit with the additions, so the bearing cases moved to their own test module beside it.

Left alone deliberately: the `GroupSet` u32 bitset. You're right that it breaks at tag 0x20 rather than 0x12, and it is a `1 << n` in a `const fn`, so it fails at compile time rather than silently — filing it as its own issue rather than widening this PR.

Full local CI green again: fmt, clippy `-D warnings`, `test --all-targets --no-fail-fast`, doc gate, release build, every `check-*.sh`, and the evidence gate.

luofang34:
**APPROVE WITH NITS** — both blocking findings are fixed, and I re-ran my own mutations rather than taking the claim: every one of them now reddens the suite.

Verified at `6393777`, clean worktree, each mutation applied and reverted individually.

## The two blocking findings, re-tested

| Mutation | Before (`f375765`) | Now (`6393777`) |
|---|---|---|
| `if true { return None; }` at the top of `bearings_fault` | green | **FAILED** `bearing_faults_fail_the_group_per_branch` |
| decode `_ => NavSource::Unknown` → `_ => NavSource::None` | green | **FAILED** `unknown_bearing_source_bytes_round_trip_as_the_fail_closed_sentinel` + `unassigned_bearing_source_bytes_decode_to_unknown` |
| encode `NavSource::Unknown => 255` → `=> 0` | green | **FAILED** `unknown_bearing_source_bytes_round_trip_as_the_fail_closed_sentinel` |

## Three more I ran, because "per branch" is a claim too

`bearing_faults_fail_the_group_per_branch` earns its name — I tried to slip past it three ways and it caught all three:

- Narrowed the loop to `for pointer in [&pointers.first]` (second slot unchecked) → FAILED.
- Deleted the `|| matches!(pointer.reference, HeadingReference::Unknown)` arm, leaving the source check → FAILED.
- Dropped the `pointer.valid &&` guard so an unusable pointer's NaN faults the group → FAILED on the complementary assertion, with its own message: `an unusable pointer's bearing is not a claim`.

That last one is the part I'd have expected to be missing. Pinning the *negative* case — that a pointer its own source calls unusable makes no claim about its bearing — is what stops the next person from "tightening" the check into a false fault. Good test.

`unassigned_bearing_source_bytes_decode_to_unknown` corrupting one pointer and asserting the other still reads `Nav1` also rules out a decoder that fails everything on any bad byte, which a single-pointer test would not have.

## Gates, re-run here at `6393777`

`fmt --all --check`; `clippy --locked --all-targets --workspace -- -D warnings`; `test --locked --workspace --all-targets --no-fail-fast` (0 failures, 21 suites); doc gate (exit 0); `build --locked --release --workspace`; `check-structure` (ran, did not skip), `check-release-manifest`, `check-release-markers`, `check-instrument-requirements`, `check-certification-claims`, `check-standards-registry`, `trace-report`, `detect-target`; both `evidence-gate` invocations → `verdict: VALID`; `cargo check --target thumbv7em-none-eabihf` over the twelve no_std crates.

`instrument-bench` still reports `scene digest b15a5d86… (matches pin)`, `screen-composition digest 59d8086a… (matches pin)`, `admission: 312 cases pass`. The module split moved no behaviour: `tests.rs` is 435 lines and `bearing_tests.rs` is 163, both under the limit, and `encode`/`locate_payload` widening to `pub(super)` is the minimum needed.

---

## One nit the fix commit introduced

**`crates/indicate-instrument-state/src/abi/v8.rs:48` — the new module-doc link points at a private module.** The doc gate emits:

```
warning: public documentation for `v8` links to private item `crate::group_id`
```

`lib.rs:34` is `mod group_id;` (private); the public surface is the `pub use` on `lib.rs:54`. rustdoc drops links to private items, so "The allocation table in [`crate::group_id`]" renders as plain text in the published docs — the stable anchor that replaced the stale prose does not actually anchor.

Not a gate failure: `rustdoc::private_intra_doc_links` is not in the workflow's deny list (`.github/workflows/ci.yml:43` denies `missing_docs` and `broken_intra_doc_links` only), and the gate exits 0. But it is new — I ran the doc gate at `f375765` and at `6393777`: zero such warnings before, one after. [`crate::GroupId`] resolves and is re-exported.

## Previously-raised nits still open (all non-blocking, unchanged)

- `group_id.rs:19` — `| 0x01–0x0D | the variants below |` still doesn't cover 0x12, which is now also a variant.
- `stamped.rs` still carries a second copy of the `NavSource` ↔ byte mapping alongside `decode_nav`/`encode_nav`. Lower risk now that both copies are tested, but a shared helper would still be better than two tested copies.
- `hsi/bearing.rs:22,24` — `HEAD_R` and `TAIL_R` are the same value with doc comments describing different geometry.
- `bearings_rose_rad: [Sig<f32>; 2]` positionally coupled to `BearingPointers { first, second }`.
- A `Stale`/`Degraded` bearing group draws both needles at full strength with no flag, unlike the CDI's `NAV`. Worth one sentence in `bearing.rs`'s module doc if it is deliberate.
- `GroupSet`'s `1 << group as u32` breaks at tag 0x20 — follow-up issue, not this PR.

Nice fix. The `group_id.rs` row now stating the actual byte coding, and deleting the stale "Landed so far" sentence rather than patching it, were both the right calls.

PR #78, opened 2026-08-18
Say what the automation is doing, and what it is about to do (#50)

### LINKED ISSUE #50: No surface annunciates AP/FD modes, and no AP target values exist in state

luofang34:
Closes #50.

**Merges last, and now can.** #76, #75, #74 and #72 have all landed; this branch is merged up to the settled batch head with every pin recomputed by running the tests. It is the sixth and final allocation of the v8 batch, after which #64 takes the batch to `main`.

## What it is for

A pilot flying against automation has two questions the rest of the panel set cannot answer: what is the automation holding now, and what is it about to hold. The active mode answers the first and the armed mode answers the second, so they are separate fields rather than one field the display guesses a tense for.

## The wire

- **0x14 ApModes**, stamped, 12 bytes: engagement, active and armed lateral mode, active and armed vertical mode, three reserved zero bytes, then the trailing `age_ms`. Stamped because a frozen `AP` is exactly the failure the freshness discipline exists to prevent.
- **0x15 ApTargets**, declared, 20 bytes, mirroring the selections layout field for field: target airspeed, the altitude target with the same reference-identity trio as the selected altitude, and target vertical speed.

Every mode enum carries `None` and `Unknown`, and they are not the same thing. `None` is the automation saying it holds nothing on that axis; `Unknown` is a wire value this build cannot read. One `Unknown` fails the whole group — annunciating the modes that *did* decode would say the unreadable axis is holding nothing, which nobody claimed.

An altitude target whose reference identity does not match the displayed datum is withheld rather than drawn. The rule that decides this is the one the pilot's selected altitude already used, extracted so both callers share it: two rules for what makes a reference identity complete is one rule too many.

## The surface

Its own panel rather than a band inside the PFD — the PFD already annunciates the flight director's mode, and a second mode vocabulary competing for the same strip would leave a reader deciding which annunciation answers which question. Joining `BUILTIN_PANELS` moves the composition and screen-composition digests once, and makes the release's panel set change (the CHANGELOG row and the "panel set changed" line both move, and `check-release-markers.sh` enforces it).

`Theme` gains `mode_active` **and a validation rule with it**: the active-mode color must keep the same distance from both guidance colors that those two must keep from each other. Green is what a conventional annunciator would reach for, and green already means radio-nav guidance here.

## Guardrails

Every one below was verified by gutting the code it covers and confirming the suite goes red — the four reviews on the sibling PRs made it clear that is the only way to know:

- Fail-closed decode for all three enums in both directions, including an unassigned byte and the 255 sentinel.
- Each mode byte at its own offset, with a golden fixture whose five mode bytes are **distinct** — so a codec that read the armed mode where the active one lives cannot round-trip. (This is the trap the airframe-config review found with `flap_ratio`/`flap_selected_ratio` both set to 0.5.)
- Faults per branch for both groups.
- The panel annunciates nothing under a stale group or an untrusted source, tested with states that carry modes **and** a status that refuses them — withholding a group zeroes its data, so withholding alone cannot tell a refused status from an absent value.
- The panel declares both groups it consumes, so the admission withholding matrix covers them.
- Target row names carry a `SEL` prefix so no row name is also a mode label, held by a test that walks every mode label the vocabulary can produce.

## Posture

No shipped feeder publishes either group, matching the flight director's precedent. `typical()` therefore stays unfed and the panel's pinned raster baseline is its unfed presentation — which is itself worth pinning, since it proves the panel draws nothing rather than last-known modes. `fully_fed()` feeds both groups, and the panel contributes two extreme states: modes and targets engaged on both axes, and an altitude target measured against another origin.

## Re-pinned

Composition digest, screen-composition digest, admission case count (280 → 372) and criticality bands, the v8 golden frames, `release-manifest.json`, the CHANGELOG table and panel-set row, the scene-layer protocol's panel-profiles table, and the new panel's raster baseline. ATT-01 evidence re-recorded (113/195/99); evidence gate `VALID`.

Docs: `AIR-IN-015` for the input, and its own intended-function row rather than the flight director's, since the function matrix withholds command-guidance credit without a separate intended function.

Local CI green: fmt, clippy `-D warnings`, `test --all-targets --no-fail-fast`, doc gate, release build, and every `check-*.sh`.


luofang34:
## Review of #78 — part 1 of 2: mutation testing

Posting in parts; the review is still running. **No blocking finding in this part.**

Method: each mutation applied to the tree at `c2e601b`, then
`cargo test --locked --workspace --all-targets --no-fail-fast` over the whole
workspace, then restored and the tree confirmed clean (`git status --porcelain`
empty) before the next. Baseline at `c2e601b` is green.

| # | mutation | result | first test to catch it |
|---|----------|--------|------------------------|
| a | `ap_modes_fault` body → `None` (`crates/indicate-instrument-state/src/validate.rs:302`) | **RED** | `abi::v8::autoflight_tests::autoflight_mode_faults_fail_the_group_per_branch`, `autoflight::tests::one_unknown_mode_takes_the_whole_annunciation_down` |
| b | `ap_targets_fault` body → `None` (`validate.rs:317`) | **RED** | `abi::v8::autoflight_tests::non_finite_autoflight_targets_fault_their_group` |
| c | `VerticalMode::from_u8` fallback `Unknown` → `None` (`crates/indicate-instrument-state/src/autopilot.rs:170`) | **RED** | `unassigned_autoflight_bytes_decode_to_unknown_not_to_a_benign_variant`, `unknown_autoflight_bytes_round_trip_as_the_fail_closed_sentinels` |
| d | `ApEngagement::to_u8` `Unknown => 255` → `=> 0` (`autopilot.rs:62`) | **RED** | `unknown_autoflight_bytes_round_trip_as_the_fail_closed_sentinels` |
| e | panel gate `data.ap_modes.status.shows_value()` → `true` (`sets/indicate-instrument-panels/src/autoflight.rs:55`) | **RED** (3 tests) | `a_stale_group_annunciates_nothing_even_though_it_carries_modes`, `a_source_that_declares_itself_unusable_annunciates_nothing`, `one_unknown_mode_takes_the_whole_annunciation_down` |
| f | `GroupId::ApModes` arm → trust-free `group_freshness(..).worst(fault_status(..))` (`crates/indicate-instrument-state/src/resolve/group_status.rs:177`) | **RED** | `autoflight::tests::a_source_that_declares_itself_unusable_annunciates_nothing` |
| g | `identity_compatible` → always `true` (`crates/indicate-instrument-state/src/resolve/altitude_signal.rs:89`) | **RED** (9 tests) | 5 in `resolve::altitude_tests` (the selection side), `autoflight::tests::an_altitude_target_against_another_datum_is_withheld`, 3 digest tests |
| h | swap `lateral_armed` (off 2) / `vertical_armed` (off 4) in **both** `decode_ap_modes` and `encode_ap_modes` (`crates/indicate-instrument-state/src/abi/v8/equipment.rs:121,123,187,189`) | **RED** | `abi::v8::fixture_tests::full_frame_is_pinned` |
| i | delete the `mode_active` guidance-distance loop (`crates/indicate-instrument-symbology/src/theme.rs:321`) | **RED** | `theme::tests::a_mode_color_wearing_a_guidance_hue_is_rejected` |

9 of 9 red. Unlike the four siblings in this batch, this PR has no fail-open
guardrail among the ones I probed. Two are worth calling out as done right:

- **(h)** is the trap the airframe-config review found. A *consistent* swap of
  both sides round-trips perfectly, so only a golden fixture with **distinct**
  mode bytes can catch it — and `state-abi-v8.full.hex` does. The five mode
  bytes in `fixtures.rs` are genuinely distinct, which is what makes the pin
  load-bearing rather than decorative.
- **(g)** killing `identity_compatible` reddens the *selection* tests
  (`equal_values_from_different_origins_are_incompatible`,
  `equal_values_from_different_geoid_models_are_incompatible`,
  `disputed_baro_setting_suppresses_the_baro_bug`,
  `bug_requires_matching_class_before_identity_is_even_considered`,
  `unknown_reference_class_fails_closed`) **and** the new target test. The
  extracted rule is covered from both callers, so a future edit to it cannot
  quietly change the altitude-bug behavior without a test noticing.

luofang34:
## Blocking finding (posting now, rest of the review to follow)

**`crates/indicate-instrument-state/src/abi/v8/stamped.rs:22-23` — the wire-layout table now carries two contradictory `dynamics` rows.**

```
22  //! | dynamics | basis u8; 0×3; turn f32; lateral f32; age f32; ias trend f32 | 20 |
23  //! | dynamics | basis u8; 0×3; turn f32; lateral f32; age f32 | 16 |
```

Line 23 is new in this PR and is wrong. `encode_dynamics` writes
`put_f32(p, 16, or_nan(dynamics.ias_trend_mps2))` and returns `Ok(Some(20))`
(`stamped.rs:357`), and `min_len(GroupId::Dynamics) => 20`
(`crates/indicate-instrument-state/src/abi/v8.rs:122`). Line 22 is the truth;
line 23 is the pre-airspeed-trend layout.

Reproduction:

```sh
git show origin/abi/v8-batch:crates/indicate-instrument-state/src/abi/v8/stamped.rs | sed -n '11,25p'
# one dynamics row: "... ias trend f32 | 20 |"
git diff origin/abi/v8-batch...HEAD -- crates/indicate-instrument-state/src/abi/v8/stamped.rs | head -20
# -  | bearings | ... | 20 |
# -  | airframe | ... | 24 |
# +  | dynamics | basis u8; 0×3; turn f32; lateral f32; age f32 | 16 |
```

The base had exactly one, correct, `dynamics` row. Moving `bearings` and
`airframe` out to `equipment.rs` was right and their rows correctly left with
them — but the deletion resurrected a stale `dynamics` row in their place
rather than simply removing two lines. The shape of it (16 bytes, no `ias
trend`) is exactly the layout from before #72 landed the trend field, so this
reads as a conflict resolved by taking an old hunk rather than by rebuilding
the table.

That matters beyond the typo: this PR's own description commits to recomputing
after each sibling lands "by running the tests, not by resolving conflicts with
a strategy flag", and no test or script reads this table, so it is precisely
the artifact that discipline exists to prevent. A feeder author reading the
stamped-lane contract now finds two mutually exclusive statements about the
same group and no way to tell which is current.

Fix: delete line 23. (Line 22 already says the right thing.)

Nothing else is blocking so far. Full part 2 — reference-identity refactor,
docs, `BUILTIN_PANELS`, the splits, the loosened guard, and the CI battery —
follows shortly.

luofang34:
## Review of #78 — part 2 of 2: REQUEST CHANGES (three documentation defects; the code is sound)

Every gate is green and every guardrail I probed is real — see part 1 for the
9-of-9 mutation table. What blocks is three places where a mechanical edit was
not read back, one of which I posted separately above. They are all one- or
two-line fixes.

---

## Blocking

### 1. `crates/indicate-instrument-state/src/abi/v8/stamped.rs:23` — two contradictory `dynamics` rows

Posted in full above. Line 23 (`… age f32 | 16 |`, new in this PR) contradicts
line 22 (`… ias trend f32 | 20 |`) and the code, which writes `ias_trend_mps2`
at offset 16 and returns 20 (`stamped.rs:357`, `v8.rs:122`). Delete line 23.

### 2. `CHANGELOG.md:56-60` — the release paragraph ends in a sentence fragment, and its list omits this PR's own groups

```
56  This release carries the allocations that have landed so far: true
57  airspeed on the Air group, the airspeed trend on the Dynamics group,
58  the deflection scale on the Nav group, the bearing-pointer group, and
59  the airframe-configuration group.
60  airspeed on the Air group, and the two autoflight groups.
```

Line 60 is the orphaned tail of an earlier, shorter version of the same
sentence, appended after the complete one instead of merged into it. It does
not parse as English, it repeats "airspeed on the Air group", and — the part
that actually matters to a consumer — the surviving list at 56-59 never
mentions the two autoflight groups, so **this release note does not say that
`ApModes` and `ApTargets` landed.** Everything else in the entry (the digest,
the panel-set row, the two 0x14/0x15 bullets) is correct; only the summary
sentence is wrong.

Reproduction:

```sh
git diff origin/abi/v8-batch...HEAD -- CHANGELOG.md | head -12
# +airspeed on the Air group, and the two autoflight groups.   <- added after a complete sentence
sed -n '56,60p' CHANGELOG.md
```

`check-release-markers.sh` compares the five table values against the tree and
passes; it does not read prose, so nothing but a reader catches this.

Suggested fix — one sentence, no orphan:

> This release carries the allocations that have landed so far: true airspeed
> on the Air group, the airspeed trend on the Dynamics group, the deflection
> scale on the Nav group, the bearing-pointer group, the airframe-configuration
> group, and the two autoflight groups.

### 3. `docs/instruments/intended-functions.md:27` — the "Output and modes" cell cites a requirement about a different function

The new row's Output column reads
``[`AIR-OUT-003`](requirements.md#air-out-003); conventional and degraded/reversionary modes``.

`AIR-OUT-003` (`requirements.md:202`) is *Conventional instrument surface*:

> The conventional-instrument function … shall output independently
> identifiable **attitude, airspeed, altitude, heading, turn/slip, and
> vertical-speed indications** without fabricating a missing source.

`draw_autoflight` outputs none of those six. It draws mode annunciations
(`AFCS`/`FD`/`AP`/`ROL`/`HDG`/`NAV`/`APR`/`PIT`/`ALT`/`ALTS`/`VS`/`IAS`/`GS`)
and three *target* rows the row itself calls targets, not indications. There is
also no horizon in the panel, so "conventional … modes" does not describe it
either. I checked all eleven `AIR-OUT-*` requirements
(`grep -n '^### AIR-OUT' docs/instruments/requirements.md`) and none covers an
autoflight annunciator.

The precedent for exactly this situation is the row directly below, added for
the other advisory non-flight surface, which cites no `AIR-OUT` at all:

> | Machine-monitoring readout | … | **Advisory text rows only; no flight-data mode participates** | … |

That is the honest form, and it is what this row should take — something like
"Advisory mode and target annunciation only; no flight-data mode participates".
This is the one claim in the row that is not true of the code; every other
claim in it checks out (see below).

Nothing automated catches this: `check-instrument-requirements.sh` and
`check-certification-claims.sh` both pass, because the anchor resolves — they
verify that the link *works*, not that the requirement *applies*.

---

## What I verified and found correct

**Priority 2 — the reference-identity refactor is faithful.** Compared
`identity_compatible` against `origin/abi/v8-batch`'s `selection_compatible`
arm by arm:

| base | extracted | same? |
|---|---|---|
| `altitude_sel_m.is_none() \|\| altitude_sel_class != displayed` → `false` | `!identity.present \|\| identity.class != displayed` → `false` | yes (`present = altitude_sel_m.is_some()`) |
| `LocalRelative` → `altitude_sel_origin == decl.origin` | `identity.origin == decl.origin` | yes |
| `GeometricMsl` → `altitude_sel_model == decl.geoid_model && != UNDECLARED` | `identity.model == decl.geoid_model && != UNDECLARED` | yes — the `UNDECLARED` exclusion survives |
| `BaroIndicated` → `!setting_mismatch` | identical | yes |
| `Pressure \| Agl` → `true` | identical | yes |
| `Unknown` → `false` | identical | yes |

`selection_compatible` is now a thin adapter that fills `AltitudeIdentity` from
`state.selections` and forwards. Nothing about the altitude bug changed. The
empirical confirmation is mutation (g): stubbing `identity_compatible` to `true`
reddens five *selection*-side tests as well as the new target test, so the
shared rule is held from both callers.

**Priority 3 — both doc rows landed.** `AIR-IN-015` exists at
`requirements.md:172-181` with a well-formed `<a id="air-in-015"></a>` anchor
matching the two `requirements.md#air-in-015` links, and it is cited from
`aircraft.rs:394,397` and `supplementary.rs:86`. Unlike the sibling in #79,
this one is not missing. Checking the intended-function row's remaining claims
against the code:

| claim | verdict |
|---|---|
| "reporting what the automation holds and what it is armed to hold" | true — `mode_rows` draws active above armed per axis |
| "with the altitude datum under AIR-IN-002" | true — `ap_targets_resolved` compares the target's identity against `state.altitude` |
| "Missing modes annunciate nothing rather than the last known ones" | true — `a_missing_group_annunciates_no_mode`; mutation (e) confirms the gate is load-bearing |
| "a stale or failed group wears its own flag under AIR-OUT-004" | true — `draw_flag` on Stale/Degraded, `draw_red_x` on Failed (`autoflight.rs:62-70`) |
| "a mode this display cannot name fails its group rather than annunciating a neighbour" | true — `one_unknown_mode_takes_the_whole_annunciation_down`; mutation (a) confirms |
| "AIR-OUT-003; conventional and degraded/reversionary modes" | **false — blocking finding 3** |

**Priority 4 — the panel joins the set cleanly.**

- Declares both groups it consumes: `required_groups: GroupSet::of(&[GroupId::ApModes, GroupId::ApTargets])` (`supplementary.rs:97`), so `case_matrix`'s `panel.required_groups.contains(group)` loop generates a withheld case for each.
- **Case count 372 is measured, not typed.** `report.cases` is incremented once per case in `check_case` (`admission.rs:259`); the assertion pins the computed total. The comment's arithmetic checks out: PFD 9×9 + HSI 8×9 + autoflight 7×3 + monitor 6×2 = 186, ×2 for the alert axis = 372. Note the PR body still says "280 → 322" — stale after the siblings landed; **the tree is right and the body is wrong**, so no action beyond updating the description.
- **Criticality band is measured.** `the_pinned_criticality_bands_are_the_measured_ones` (`admission/tests.rs:45`) compares `criticality_bands(&registry)` against `BUILTIN_CRITICALITY_BANDS.panels`; it is green. The band being byte-identical to the monitor's is expected — both are text panels whose only critical region is the shared alert stack.
- **`scene-layer-protocol.md:117`** carries the row, it matches `required_layers` exactly (`Tapes`, `Annunciation` required; `Background` optional, as for the monitor, even though the panel paints it), and `layer_profile_doc_tests` parses the table and fails on disagreement.
- **Vocabulary is covered twice over.** `every_mode_label_is_in_the_panel_vocabulary` walks all 256 byte values through all three enums' `from_u8().label()` — the strongest form. The strings that test does *not* cover (`SEL IAS`/`SEL ALT`/`SEL VS`, `AFCS`/`LATERAL`/`VERTICAL`/`TARGETS`, `kt`/`FT`/`FPM`, `---`, and `{:.0}` output including a leading `-`) are covered by admission, which checks every character of every text run in all 372 cases against `PANEL_VOCABULARY` (`admission.rs:239-248`). `k` and `t` are the only lowercase in the list and both are present.

**Priority 5 — the seams.**

- `abi/v8/stamped.rs` → `equipment.rs` is the right cut: the four groups that describe equipment rather than motion (director, bearings, airframe, ap modes) move together, and `absent`/`sized` become `pub(super)` rather than being duplicated. `stamped.rs` drops ~130 lines. Aside from the table row above, clean.
- `descriptors.rs` → `descriptors/supplementary.rs` moves monitor, autoflight and config out on a stated principle ("none of these carries primary-flight credit"), leaving PFD and HSI. `descriptors.rs` drops ~120 lines while gaining one. Good.
- `resolve/equipment_signal.rs` earns its existence: `ap_targets_resolved` needs the *resolved* altitude, and bundling forces one resolution order instead of resolving the datum twice.
- **The loosened guard still refuses.** I constructed the case: added `GHOST_DESCRIPTOR,` to `BUILTIN_PANELS` (a name no descriptor defines) and ran the script.

  ```
  DRIFT: entry 0.5.0 says 'Panel set' is 'pfd, hsi, autoflight, monitor'; the tree says
         'pfd, hsi, autoflight, <unresolved:GHOST_DESCRIPTOR>, monitor'
  UNRESOLVED: a BUILTIN_PANELS entry has no readable id: … <unresolved:GHOST_DESCRIPTOR> …
  check-release-markers: FAILED   (exit 1)
  ```

  The widening from one file to `find … -name '*.rs' | sort` does not weaken it:
  the awk accumulates the slice and resolves ids in `END`, so file order is
  irrelevant, and the `<unresolved:` sentinel plus the `case` at line 164 still
  fails the build. Restored, and the script returns to `OK (0.5.0; ABI 8, scene 1,
  corpus 4, panels: pfd, hsi, autoflight, monitor)`.

**Priority 6 — full CI, all green.** `fmt --check`, `clippy --locked
--all-targets --workspace -D warnings`, `test --locked --workspace --all-targets
--no-fail-fast`, the doc gate with `-D missing_docs -D
rustdoc::broken_intra_doc_links`, `build --locked --release --workspace`, all
six `check-*.sh`, all four `test-*.sh`, `trace-report.sh`, `detect-target.sh`,
`instrument-bench`, and both evidence-gate invocations. Evidence gate:
`nodes: 30  edges: 73  exceptions: 0  verdict: VALID`.

**Re-pinned values are computed, not hand-typed** — each has a gate that
recomputes it: `BUILTIN_SCENE_DIGEST` (`digest_tests`), criticality bands
(`the_pinned_criticality_bands_are_the_measured_ones`), case count
(`case_matrix`), raster baselines (`frame_hashes`), `state-abi-v8.full.hex`
(`full_frame_is_pinned`), `release-manifest.json`
(`check-release-manifest.sh` regenerates and diffs), CHANGELOG's five values
(`check-release-markers.sh`). I also re-ran the three recorded evidence suites
by hand: **113 / 195 / 99**, exactly the recorded counts. (The PR body's
"93/188/99" is stale, like the case count — the artifacts themselves are right.)

The golden fixture's five mode bytes are `2, 1, 4, 6, 3` — genuinely distinct,
which is the only reason mutation (h)'s consistent encode+decode swap is
detectable. That trap was closed deliberately.

---

## Non-blocking

- **`crates/indicate-instrument-state/src/group_id.rs:168`** — 123 characters
  in a file that wraps doc comments at ~67. The rewrite of the `index()`
  rationale lost its wrap; `cargo fmt` does not reflow doc comments, so it
  passed. Same root cause as blocking findings 1 and 2 — an edit not read back.
  The claim itself is correct (`tag - 1` for `ApTargets` is 20 against a
  17-entry table).
- **`crates/indicate-instrument-state/src/abi/v8/declared.rs:66-69`** — "The
  target layout mirrors the selections layout field for field … putting them at
  different offsets would invite reading one against the other." The altitude
  class and model are at offsets **5 and 6** in selections but **4 and 5** in ap
  targets; only the altitude `f32` (8) and origin `u32` (12) actually align. The
  weaker true statement — both carry the same reference-identity trio, and the
  20-byte shape matches — is worth keeping; "field for field" is not.
- **`crates/indicate-instrument-state/src/resolve/altitude_signal.rs:111`** —
  `selection_compatible` lost the doc comment it had on the base. It is private
  so `missing_docs` does not fire, but the surviving one-line adapter now has no
  explanation of what "the selection" means here.
- **`crates/indicate-instrument-state/src/resolve/equipment_signal.rs:4-8`** —
  the second paragraph justifies the module by the shared datum, which is true
  of `altitude` + `ap_targets` and not of `bearings` or `ap_modes`, both of which
  also live here. The first paragraph's "signals that come from equipment rather
  than from the aircraft's motion" is the principle that actually covers all
  four; the datum paragraph reads as a rationale for a narrower module than the
  one that shipped.
- **`crates/indicate-instrument-descriptor/src/states.rs:107,136`** — extracting
  `typical_valid()` out of `typical()` is unrelated to #50. Harmless, but it is
  a second issue in a one-issue PR and it moves a file the evidence graph binds
  by digest.
- **`crates/indicate-instrument-glyphs/src/vocabulary.rs:3-10`** — the module
  doc still derives `PANEL_VOCABULARY` from "every `text(...)` literal and every
  `fmt_label!` template **across the PFD and HSI**". A fourth builtin panel now
  contributes strings. Not this PR's file, and the set is unchanged (every
  autoflight character was already covered), so it is only a staleness note.
- **PR description** — the "Re-pinned" section says "admission case count
  (280 → 322)" and "ATT-01 evidence re-recorded (93/188/99)". The tree says 372
  and 113/195/99, and the tree is correct in both cases; the description was
  written before the siblings landed. Worth refreshing before merge so the body
  matches what it is describing.

---

Fix the three above and this is a clear approve. The engineering underneath is
the strongest in the batch: nine deliberate mutations, nine reds, including the
consistent-swap codec trap and the shared-rule extraction that a sibling review
had to catch by hand.

luofang34:
All three blocking findings fixed in `b4efaf3`, and the stale PR body corrected. Your diagnosis of the common root cause is right and worth stating plainly: every one of these was a mechanical merge edit I never read back as English, in a PR whose own body promises to resolve conflicts by running the tests. Running the tests catches the pins. It catches nothing in prose, and no script in this repo does either.

**1. Two contradictory `dynamics` rows.** Exactly as you describe: removing the bearing and airframe rows (which moved to the equipment module) resurrected a pre-trend row in their place, so the table stated 20 bytes and 16 bytes for the same group two lines apart. The stale row is gone. Every group's layout now appears once, in the module that owns it — I checked all four equipment rows are in `equipment.rs` and `dynamics` appears once in `stamped.rs`.

**2. The orphaned CHANGELOG fragment.** Fixed, and you were right that the surviving sentence never named the autoflight groups. It now names every allocation the batch directs. Worth noting for the record: the per-group bullets for 0x14 and 0x15 were present and correct — the defect was confined to the summary sentence, so a reader who got past it was not misled.

**3. `AIR-OUT-003`.** This is the one I am glad you caught, because it is a coverage claim rather than a typo. That requirement is the conventional instrument surface — attitude, airspeed, altitude, heading, turn/slip, vertical speed — and `draw_autoflight` outputs none of them. I confirmed none of the eleven `AIR-OUT-*` requirements covers an autoflight annunciator. The row now says so and cites none, following the machine-monitoring row's precedent, which is exactly the honest form you pointed at. `AIR-OUT-004` stays in the failure column, because that one genuinely does cover the flag behaviour.

Also took the 123-character doc line at `group_id.rs:168`, another artefact of the same edit.

**On the PR body:** you are right on both numbers and right that the tree was correct. Case count is 372 and the recorded evidence is 113/195/99 — I re-ran all three suites and they match the artefacts exactly. The body said 322 and 93/188/99, which were true when I opened it as a draft and stopped being true when the siblings merged. Updated, along with the now-obsolete "merges last" caveat.

**On the mutation table:** thank you for running all nine and reporting the table rather than only the failures. The two you singled out are the two I most wanted an outside check on — (h) is exactly the trap the airframe-config review found open, which is why the fixture carries five distinct mode bytes, and (g) reddening five *selection*-side tests is the evidence that the extracted identity rule is held from both callers rather than only from the new one. That was the riskiest change in the PR and I could not have verified it myself with any credibility.

luofang34:
## APPROVE WITH NITS — merge it; the single nit below is wording and must not hold the batch.

Re-reviewed at `b4efaf3` (detached `origin/fix/50-ap-modes-targets`), diffed against `origin/abi/v8-batch`. All three blocking documentation defects are closed, the mutation set is still red where it should be, the PR body's counts now match the tree, and the full battery is green. Nothing blocks the batch merge.

### 1. Wire-layout table — **closed**

The duplicate `dynamics` row is gone. I did not stop at "the table agrees with itself": I enumerated every layout row in the crate and checked each against `min_len` *and* against the bytes the codec actually writes.

Every group appears in exactly one table, and the four tables partition all 17 groups: `stamped.rs` 8, `equipment.rs` 4, `declared.rs` 4, `monitor.rs` 1 (an offset table rather than a row). No group is listed twice; none is missing.

| group | table row | `min_len` | encoder returns |
|---|--:|--:|--:|
| attitude / kinematics / air / nav / wind | 32 / 28 / 16 / 43 / 12 | same | same |
| heading / variation / dynamics | 12 / 12 / 20 | same | same |
| selections / trust / altitude / ap targets | 20 / 8 / 12 / 20 | same | same |
| director / bearings / airframe / ap modes | 16 / 20 / 24 / 12 | same | same |
| monitor text | 274 (`6 + 8×33 + 4`) | `MONITOR_LEN` | `MONITOR_LEN` |

Field order and offsets in each row match each codec's `get_*`/`put_*` calls, checked one row at a time — including the two the append policy makes easy to get wrong (`air`: age at 8, `tas` at 12; `dynamics`: age at 12, `ias trend` at 16).

Third leg, against the wire rather than against the source: parsing `state-abi-v8.full.hex` gives 17 tags in ascending order whose lengths are exactly the table's — `0x01/32 0x02/28 0x03/16 0x04/43 0x05/12 0x06/20 0x07/8 0x08/12 0x09/12 0x0a/12 0x0b/20 0x0c/274 0x0d/16 0x12/20 0x13/24 0x14/12 0x15/20`. The `0x14` payload is `02 01 04 06 03 00 00 00 | 0000a042` — engagement `Autopilot`, lateral active `Roll`, lateral armed `Approach`, vertical active `GlideSlope`, vertical armed `AltitudeCapture`, three reserved zeros, age 80.0 — byte for byte the documented row, with five distinct mode bytes as claimed. `0x15` decodes to 61.0 / class 0 / model 0 / pad / 1200.0 / origin 7 / 2.5, matching the `declared.rs` row.

### 2. CHANGELOG — **closed**

The fragment is gone and the sentence parses: *"This release carries every allocation the batch directs: true airspeed on the Air group, the airspeed trend on the Dynamics group, the deflection scale on the Nav group, and the bearing-pointer, airframe-configuration and autoflight groups."* That names all three appends and all four ids.

The two per-group bullets exist and describe the real wire, not a paraphrase of it:
- **0x14 ApModes, stamped, 12 bytes** — engagement, active/armed lateral, active/armed vertical, three reserved zero bytes, trailing `age_ms`. Matches `encode_ap_modes` and the golden frame above.
- **0x15 ApTargets, declared, 20 bytes** — airspeed, altitude target with the class/geoid-model/origin trio, vertical speed, mirroring `selections` field for field. Matches `encode_ap_targets` (offsets 0, 4, 5, 6–7 pad, 8, 12, 16).

### 3. `AIR-OUT` citation — **closed**

I read all eleven `AIR-OUT` requirements rather than taking the claim on trust. 001 is the PFD surface, 002 the HSI, 003 the conventional six (attitude, airspeed, altitude, heading, turn/slip, VS), 005 SVS, 006 HUD-SIM, 007 the non-conformal repeater, 008/009 exclusions, 010 TAWS independence, 011 the composed multi-panel surface. None covers an autoflight annunciator, so citing none — and saying why — is right, and it follows the machine-monitoring row's precedent as the commit says.

`AIR-OUT-004` in the failure column is correct for what the panel does. The requirement binds *every credited display output* to present stale/missing/failed states conspicuously and never retain a last-good value; `draw_autoflight` gives `Failed` a full-frame red X labelled `AFCS`, `Stale`/`Degraded` an `AFCS` flag, and `Missing` nothing at all — which is exactly the row's "missing modes annunciate nothing rather than the last known ones". The PFD row cites `AIR-OUT-004` in the same column for the same reason, so the citation is consistent with the matrix rather than borrowed.

### 4. PR body counts — **closed**

Re-ran the three recorded suites at this HEAD and read the case matrix:

- `cargo test -p indicate-instrument-panels` → **113** passed
- `cargo test -p indicate-instrument-state` → **195** passed
- `cargo test -p indicate-instrument-raster` → **99** passed
- `admission::tests::builtin_panels_pass_admission` asserts **372**, and `instrument-bench` reports `admission: 372 cases pass, 226 counted warnings`. The comment's arithmetic checks out: PFD (5+4)×9 + HSI (5+3)×9 + autoflight (5+2)×3 + monitor 6×2 = 186, drawn twice = 372.

Body and evidence artifacts agree with the tree.

### Mutation re-check — still red

Each mutation applied, suite run, file restored with `git restore`, tree confirmed clean between each.

| mutation | result |
|---|---|
| `ap_modes_fault` → `None` | RED — `abi::v8::autoflight_tests::autoflight_mode_faults_fail_the_group_per_branch` and `autoflight::tests::one_unknown_mode_takes_the_whole_annunciation_down` |
| panel `shows_value()` gate → always true | RED — three tests: `a_stale_group_annunciates_nothing_even_though_it_carries_modes`, `one_unknown_mode_takes_the_whole_annunciation_down`, `a_source_that_declares_itself_unusable_annunciates_nothing` |
| swap mode byte offsets 1↔2 in **both** `encode_ap_modes` and `decode_ap_modes` | RED — `abi::v8::fixture_tests::full_frame_is_pinned` |

The third is the one worth calling out: the swap is round-trip-preserving, so only the golden frame can catch it, and it does — because the fixture's five mode bytes really are distinct (`2, 1, 4, 6, 3`). That is the guardrail the body claims, working for the reason the body gives.

### Battery

`fmt --check` OK · `clippy --locked --all-targets --workspace -- -D warnings` clean · `test --locked --workspace --all-targets --no-fail-fast` all green, zero failures · doc gate with `-D missing_docs -D rustdoc::broken_intra_doc_links` clean · `build --locked --release --workspace` OK · `check-structure`, `check-release-manifest`, `check-release-markers` (0.5.0; ABI 8, scene 1, corpus 4, panels pfd/hsi/autoflight/monitor), `check-instrument-requirements` (74 unique), `check-certification-claims`, `check-standards-registry`, `trace-report`, `detect-target` all OK · all four `test-*.sh` self-tests OK · `thumbv7em-none-eabihf` cross-check of the twelve no_std crates OK · `instrument-bench` reproduces both pinned digests (`3ed6079…`, `ca5e33b…`) · evidence gate `VALID` with `--resolve-selectors` and with `--resolve-selectors --require-resolvable`.

### One nit, not blocking

The summary says "every allocation the batch directs", but the registry table in `group_id.rs` carries `facility_type` on Nav (42 → 43 → 44) and the release's own bullet lists it as "to follow" (#55). The entry is internally consistent — it defines the batch as four ids and three appends, and the registry header explicitly permits allocating an id before its group exists — so this is a wording preference, not an error. If it ever reads as over-claiming, "every allocation this batch lands" would settle it. Do not hold the merge for it.

Issue #77, opened 2026-08-18
GroupSet indexes its bitset by raw wire tag, which stops working at 0x20


luofang34:
`GroupSet` (`crates/indicate-instrument-descriptor/src/group_set.rs`) is a `u32` whose bit position is the raw wire tag:

```rust
pub const fn of(groups: &[GroupId]) -> GroupSet {
    let mut bits = 0u32;
    ...
    bits |= 1 << groups[i] as u32;
```

The id registry in `group_id.rs` reserves `0x16–0xDF` for future standard groups. The first allocation at or above `0x20` shifts past the width of the word.

## Why this is not urgent

It fails loudly, not silently. `of` is a `const fn`, so `1 << 32` is an overflow in const evaluation and the build stops. Nothing decodes wrong in the meantime, and the batch's highest allocation is `0x15`.

## Why it is still worth fixing

The failure lands on whoever allocates the group, not on whoever chose the representation, and it lands as a const-eval error some distance from the cause. The fix is also the moment to decide what the set is keyed by: `GroupId::index()` is already the dense-table key everywhere else (`GroupStatuses` uses it), and keying the bitset by `index()` instead of by tag makes the width follow `COUNT` rather than the sparsest tag ever allocated.

## What implementing it needs

- Key the bitset by `GroupId::index()` rather than by `self as u32`, so the word only has to be as wide as `COUNT`.
- `digest_panel_contract` hashes `required_groups.bits()`, so the change moves the composition digest and the screen-composition digest. That is a deliberate re-pin with its reason in the commit, not a drift.
- A guardrail in the same change: a test that constructs a `GroupSet` containing the highest allocated id and asserts `contains` round-trips, plus a static assertion that `COUNT` fits the chosen word. The current shape has neither, which is why the limit is invisible until it is hit.
- Decide whether `bits()` stays public. It is part of the digest input, so its meaning changing is a wire-visible event for anyone pinning the digest.

luofang34:
Resolved by merged PR #85. GroupSet capacity now follows the defined groups, with regression coverage.

PR #80, opened 2026-08-18
Give the configuration panel the rows it shipped without (#79)

### LINKED ISSUE #79: The configuration panel ships without the documentation and contract rows its own issue required
PR #75 (issue #57) landed the airframe-configuration group and panel. A second independent review found that several artefacts the issue named as *part of the implementation* did not land with it. All four are confirmed on `abi/v8-batch` at `aa632b9`.

Issue #57 said it plainly: "This adds a new intended function, and that documentation is part of the implementation, not a follow-up."

## What is missing

**1. No `intended-functions.md` row, and the existing row does not cover it.**
`grep -i 'flap\|trim\|airframe config' docs/instruments/*.md` returns nothing. The panel sits implicitly under the "Conventional instruments" row, which does not reach it: that row's inputs are "The validated sources used by the corresponding PFD quantities", and there is no PFD flap or trim quantity. Its output requirement `AIR-OUT-003` enumerates "attitude, airspeed, altitude, heading, turn/slip, and vertical-speed indications" — flap and trim are in neither list.

**2. No `requirements.md` entry.** There is no `AIR-IN` requirement for configuration input, so the group's reference identity, source declaration, and fail-closed range behaviour are stated only in code. `check-instrument-requirements.sh` passes at 73 requirements and cannot see the omission — it verifies the registry is internally consistent, not that a shipped function has an entry.

**3. No `config` row in `scene-layer-protocol.md`.** That section states the table "is the `required_layers` mask of each shipped descriptor" and that "a mask change must arrive with its row". `CONFIG_DESCRIPTOR` declares `Tapes | Annunciation` and has no row. The guard cannot catch this: `layer_profile_doc_tests` iterates `BUILTIN_PANELS` only, and this panel deliberately ships outside it. A backend author reading that table believes the crate ships three panels.

**4. The crate now exports two `PanelSet`s, against two contract documents.**
`panel-contract.md:24` — "Under `sets/`, one crate per set, exporting one `PanelSet`."
`crates/README.md:87` — "Panel providers, one crate per set, each exporting one `PanelSet`."
`sets/indicate-instrument-panels` exports both `BUILTIN_SET` and `CONFIG_SET`. Neither document was amended.

Shipping outside `BUILTIN_PANELS` was the right call and provably kept the composition digest, the screen-composition digest and the panel-set release value unmoved. The defect is not *where* it ships but that one crate now ships two sets while two documents say it may not.

## Also worth fixing while in here

**The group region is not tight, and its justifying comment is wrong.** The declared region is 400×300 at (40,40) on a 480×360 frame — 83% of each axis. Decoding the emitted scene, the only *claimed* ink is the flap numeral anchored at x=184 and the trim numeral at (276,320); nothing claimed exists left of x≈180 or right of x≈330. The rectangle swallows the `FLAP` and `TRIM` labels, the whole ladder and its ticks, and ~140 units of empty space each side, and it merges two separate readouts into one surface where the PFD declares per-readout regions (its IAS box is 40×36).

`panel-contract.md:199-201`: "a region generous enough to swallow them is the vacuous case the mechanism exists to prevent." A composition plans obscuration around regions, so this tells a screen it may cover five-sixths of the panel.

The comment justifying the size reads "with the group withheld the panel dashes both, and the region has to hold the dashes as well as the readings." That reason does not hold: dashes are painted with plain `text` and carry no claim, and `admission/regions.rs` counts only runs whose attribution resolves to a required group. The dashes are irrelevant to `GroupRegionEmpty`.

## What implementing it needs

- Decide (4) first, because it changes where the rest of the rows go: either a real `sets/indicate-instrument-config` crate with the registrations `panel-contract.md:411-425` lists, or an amendment to both documents saying a crate may ship more than one set and what that means for set identity. The second is cheaper and defensible — set identity already stays out of the scene digest — but it must be a stated decision, not a silence.
- An `AIR-IN` requirement for configuration input, with its anchor, covering source identity, source time, quality, and the range each axis is defined over.
- An `intended-functions.md` row of its own, since the existing one demonstrably does not cover it.
- The `scene-layer-protocol.md` row. Consider whether `layer_profile_doc_tests` should iterate every exported set rather than `BUILTIN_PANELS`, which is the guardrail that would have caught this.
- Tighten the region to the claimed ink, or split it per readout as the PFD does, and replace the comment with the reason that actually holds.

luofang34:
Closes #79.

**Based on #78, not on the batch head.** It needs `AIR-IN-015` to exist so this one can be `AIR-IN-016`. Merge #78 first; this then retargets to `abi/v8-batch` cleanly.

## What this fixes

PR #75 landed the airframe-configuration group and panel. Issue #57 had said the documentation was part of the implementation, not a follow-up. It shipped without it, and a second independent review found the gap after the merge.

**The panel had no intended-function row**, and the "Conventional instruments" row it sat under does not reach it: that row's inputs are "the validated sources used by the corresponding PFD quantities", and there is no PFD flap or trim quantity. Its output requirement `AIR-OUT-003` enumerates attitude, airspeed, altitude, heading, turn/slip and vertical speed — flap and trim are in neither list. There is now a row of its own, and like the autoflight row beside it, it cites no `AIR-OUT` requirement and says why rather than borrowing one written for another surface.

**No requirement covered the input.** `AIR-IN-016` does: source identity, source time, quality, the range each axis is defined over, that a value outside its range fails rather than clamps, and that a selected detent and a sensed position are separate quantities never substituted for one another.

**The layer-protocol table had no row, and its test could not notice.** `layer_profile_doc_tests` walked `BUILTIN_PANELS`, and this panel deliberately ships outside that set so a shell composes it only when the airframe has the sensors. The test now walks every set the crate exports — which is the guardrail that would have caught the omission — and it immediately demanded the missing row.

**One crate exporting two `PanelSet`s** contradicted `panel-contract.md:24` and `crates/README.md:87`. The rule's stated purpose is the tier law (a set may not depend on the machinery that judges it) and portability (a set may live outside this repository); a second set carrying the same obligations touches neither. Both documents now state the decision: the crate is the compilation and dependency unit, a `PanelSet` is the composition unit, a shell selects sets rather than crates, and two sets belong apart when a consumer of one should not have to compile the other.

## The region

It claimed 400×300 of a 480×360 frame — five-sixths of each axis — justified by needing to "hold the dashes as well as the readings". That reason does not hold: dashes are plain `text` carrying no claim, and `admission/regions.rs` counts only runs whose attribution resolves to a required group.

It now holds the two attributed numerals. Still tall, because the flap numeral rides its pointer down the whole scale; no longer wide, because neither numeral leaves the column they share. **Verified load-bearing** by `every_attributed_numeral_lies_inside_the_declared_region`, not by admission: `GroupRegionEmpty` fires only when a region catches no claim at all, so shrinking passes there until the region misses everything.

## Not done here

The `GroupSet` u32-keyed-by-tag ceiling is issue #77 — it breaks at tag 0x20, not at 0x13, and belongs with the change that moves the key.

Local CI green: fmt, clippy `-D warnings`, `test --all-targets --no-fail-fast`, doc gate, and every `check-*.sh`.


luofang34:
## Review — independent, at `9944d95` detached, diffed against `origin/fix/50-ap-modes-targets`

**Request changes: two blocking findings — `AIR-IN-016`'s final clause states a behaviour the panel deliberately does not implement, and the replacement region comment's stated reason is contradicted by the decoded geometry.**

The rows land, the layer-protocol row and its widened test are right, the region rectangle itself is tight and I verified it independently, and I agree with the one-crate-many-sets amendment. The two findings above are in the requirement text and in the comment that justifies the rectangle.

Every gate is green on the branch as it stands (battery at the end), so nothing below is a red build. That is the point: this is a remedial PR, and the gap it is closing was also not a red build.

---

## Blocking

### 1. `AIR-IN-016`'s final clause is not what the panel does — and the panel says so on purpose

`docs/instruments/requirements.md:191-194`:

> A selected detent and a sensed position are separate quantities and shall not be substituted for one another; **an airframe that reports one and not the other shall present the one it reports and no indication for the other.**

The panel honours that in one direction and inverts it in the other. `sets/indicate-instrument-panels/src/config.rs:102` returns from `flap_scale` before the detent is ever drawn:

```rust
let Some(ratio) = sensed else {
    // ... "---" ...
    return Ok(());
};
if let Some(detent) = selected {   // line 113 — unreachable when sensed is absent
```

and the call site at `config.rs:46-49` states the opposite policy as a deliberate design choice:

```rust
// A selected detent draws only beside a sensed position: a
// selection with nothing to compare it against is a number, not
// a configuration.
```

Reproduction (temporary test in `sets/indicate-instrument-panels/src/config/tests.rs`, using that file's own `panel`/`texts` helpers; run, observed, then reverted):

```rust
let data = panel(Some(AirframeConfig {
    flap_ratio: None,
    flap_selected_ratio: Some(0.5),
    elevator_trim_ratio: Some(0.0),
    aileron_trim_ratio: None,
    rudder_trim_ratio: None,
}), Some(40.0));
assert!(data.airframe.status.shows_value());
```

Observed output:

```
texts = ["FLAP", "---", "TRIM", "0"]
triangular marks drawn = 1 (trim pointer only == 1)
```

The group is `Valid` and fresh, the airframe reports the selected detent, and the panel presents **no indication of it at all** — not the detent mark, not a numeral. `AIR-IN-016` requires the opposite.

This is the failure mode #79 exists to close, reproduced one layer up: the requirement was written to describe the panel and was not checked against it. One of the two has to move, and it is not for review to pick — either the clause narrows to what the panel does (a detent is presented only alongside a sensed position, because an uncompared selection is not a configuration — the reasoning in `config.rs:46-48` is sound and worth promoting into the requirement), or the panel gains a selected-only presentation. Whichever way it goes, it needs the test that pins it; `config/tests.rs` covers sensed-without-selected (`each_scale_answers_for_its_own_sensor`) and detent-beside-sensed (`a_selected_detent_does_not_move_the_sensed_pointer`) and has no case for this third combination.

Two smaller notes on the same requirement, neither blocking:

- **"source identity"** is not carried. `Stamped<T>` is `{ data, age_ms }` (`crates/indicate-instrument-state/src/aircraft.rs:312-317`) and the v8 airframe payload is five `f32` plus an age (`abi/v8/equipment.rs:17`). There is no identity field for this group. Aspirational — but precedented: `AIR-IN-015` says the same words about `ApModes`, which does not carry one either, and `AIR-IN-001` through `AIR-IN-008` all read at that altitude. Consistent, so I would not hold the PR for it; worth knowing it is a statement of intent rather than a description.
- **"shall declare the range each axis is defined over"** is implemented by convention, not by declaration: the ranges are fixed in the type (`aircraft.rs:234-245`) and enforced in `validate.rs:189-208`, and nothing on the wire declares them. The *enforcement* half is genuinely implemented and correctly non-clamping — `airframe_fault` faults per field and per bound, `out_of_range_configuration_faults_the_group_per_field` covers all five fields and both bounds, and `draw_config` gates on `status.shows_value()` so a faulted group dashes rather than clamps. Verified.

The remaining clauses hold: source time is `age_ms` folded through the freshness policy; quality folds through `Trust` (`resolve/group_status.rs:169-173`); sensed and selected are separate fields never averaged. The anchor is well-formed and `check-instrument-requirements.sh` sees it — `instrument-requirements: OK (75 unique requirements)`, up from 74 on the base.

### 2. The replacement region comment's reason is contradicted by the emitted scene

`sets/indicate-instrument-panels/src/descriptors/supplementary.rs:184-186`:

> The height is most of the frame because the flap numeral rides its pointer down the whole scale; **the width is not, because neither numeral leaves the column they share.**

They do not share a column. I decoded the emitted scene through the admission harness's own run extraction (a temporary `draw_runs` probe under `crates/indicate-instrument-conformance/src/admission/tests/`, over flap ∈ {0, 0.25, 0.5, 0.75, 1.0, None} × selected ∈ {None, 0.0, 1.0} × elevator trim ∈ {−1.0, −0.2, 0.0, 0.5, 1.0, None} — 180 attributed runs, since reverted):

| readout | x extent | y extent |
|---|---|---|
| flap numeral | 184.00 → 227.71 | 61.00 → 299.00 |
| trim numeral | 276.00 → 335.14 | 311.00 → 329.00 |

The two x ranges are **disjoint, with a 48-unit gap between them**. The width is 170 precisely *because* the numerals sit in two different columns — it is the span from the left edge of one to the right edge of the other. The stated reason is the opposite of the geometry.

That matters more than a wording slip here, because replacing a justifying comment whose reason did not hold is one of the things this PR is for. The other half of the replacement is correct and I confirmed both parts: dashes go through plain `scene.text` (`config.rs:105-111`, `config.rs:161-163`), and `admission/regions.rs:78-98` skips any run whose `attribution` is `None`, so the dashes genuinely are not what the region is measured against.

**The rectangle itself is good** — this is worth saying plainly, because the tightening is a real improvement and my decode supports it. All 180 attributed run centres fall inside the declared region, over the full range of both axes including the extremes, and the margins from the region edge to the nominal ink hull are:

```
region  x 170 → 340,  y 50 → 340
hull    x 184 → 335.14,  y 61 → 329
margin  left 14.00  top 11.00  right 4.86  bottom 11.00
```

Five to fourteen units on a 480×360 frame. Tight, and no large empty margin at the edges.

The residual issue is shape, not size: one rectangle spanning two readouts placed diagonally declares 49,300 units² to hold two ink boxes totalling 11,468 units² — 77% of the declared surface belongs to neither readout, and composition plans obscuration around the whole of it. Issue #79 suggested splitting per readout, and the PFD is the shipped precedent for exactly that: `descriptors.rs:109-141` declares **three** separate regions for `GroupId::Air`. Two regions here — roughly `(180, 55, 55, 250)` for the flap numeral's travel and `(270, 305, 70, 30)` for the trim row — would be tighter, match the precedent, and make the comment true without having to write a reason for a shape that has none. I would not block on this alone; I raise it because fixing the comment and fixing the shape are the same edit.

**The "verified load-bearing" evidence does not reproduce as described.** The PR body says "shrinking the height past the ink makes `the_config_set_passes_admission` refuse." It does not:

| mutation to `CONFIG_DESCRIPTOR.group_regions` | `the_config_set_passes_admission` |
|---|---|
| `height: 290.0` → `260.0` (bottom 340 → 310, past the trim ink at y 311–329) | **passes** |
| `x: 190, y: 115, width: 20, height: 20` (a 20×20 patch, 400 of the declared 49,300 units²) | **passes** |
| `height: 290.0` → `70.0` (bottom 120) | fails — `GroupRegionEmpty { panel: "config", group: AirframeConfig, region: Region { x: 170.0, y: 50.0, width: 170.0, height: 70.0 }, frame: 480×360 }` |

The cause is `GroupRegionEmpty`'s one-witness semantics plus the corpus: every canonical state except `nothing-fed` derives from `typical()`, which fixes flap at 0.25 and elevator trim at −0.15 (`crates/indicate-instrument-descriptor/src/states.rs:120-129`), and the panel declares no extreme states. Admission therefore probes exactly two points — the flap numeral centre at (198.14, 125) and the trim numeral centre at (297.86, 320) — and a region passes if it holds *either*. None of the four declared edges is pinned by it. The number is right, but nothing in the tree holds it there; the honest options are to drop the claim, or to land the guardrail the repo's own PR discipline asks for (an extreme state driving flap 0.0 and 1.0 would make the height genuinely load-bearing, and would also give the panel the extreme-state coverage every other panel in the crate has).

---

## The judgement call: one crate, many sets — I agree with the amendment

Asked to decide, and I do not think a separate `sets/indicate-instrument-config` crate is what the contract requires.

The rule's stated purposes are the two the PR names, and I checked both against the enforcement rather than the prose. The tier law is enforced per *crate*: `check-structure.sh` classifies by path (`sets/` → kernel-only normal deps, registry as dev-dependency) and reads Cargo manifests, so both sets in one crate carry identical obligations and neither can reach the machinery that judges it. Portability is per crate too — the whole crate can live elsewhere, and a second `PanelSet` inside it does not change that. Set identity is deliberately outside the scene digest, so publishing `CONFIG_SET` disturbs no shell that does not compose it. The amendment's new criterion — sets belong apart when a consumer of one should not have to compile the other — is the right axis, and it is the one a crate boundary actually controls.

The digest claim in the amendment checks out, all three values: `BUILTIN_PANELS` is the panel-set release value (`CHANGELOG.md:20`, and `check-release-markers: OK (0.5.0; ABI 8, scene 1, corpus 4, panels: pfd, hsi, autoflight, monitor)` — no `config`); the composition digest covers a registry's panels, and `CONFIG_SET` composes into none of the shipped ones; the screen-composition digest is pinned over `BENCH_COMPOSITION` in `tools/instrument-bench`, which does not include it. The amended text weakens nothing else it touches — the kernel-only rule, the dev-dependency allowance, the `no_std` requirement and the out-of-repo clause all survive verbatim.

Where the cheap path does cost something is the guardrail, and that is a note rather than a veto — below.

---

## Non-blocking

**a. The widened doc test catches a missing row, but hard-codes the two sets.** Verified both halves.

Deleting the `| Configuration |` row from `scene-layer-protocol.md` reddens three tests with useful messages:

```
every_shipped_panel_has_a_row_in_shell_order
  left:  ["PFD", "HSI", "Autoflight", "Monitor"]
  right: ["PFD", "HSI", "Autoflight", "Monitor", "Configuration"]
the_profiles_section_holds_exactly_one_table   left: 6, right: 7
the_documented_profiles_are_the_descriptor_masks  left: 4, right: 5
```

A third set is invisible to it. `shipped_panels()` (`layer_profile_doc_tests.rs:131-133`) chains two named constants, so anything not named there is not walked regardless of whether it is exported. I added a third set to `supplementary.rs` — `PROBE_DESCRIPTOR` / `PROBE_PANELS` / `PROBE_SET`, one panel, no row in the table — and all four `layer_profile_doc_tests` passed (since reverted). That is #79 repeating verbatim on the next set.

The fix is small and belongs with the amendment that made a second set legal: export the crate's sets as one constant (`ALL_SETS`, or `SHIPPED_PANELS`) and have the test walk it, so adding a set is what registers it. Related: `panel-contract.md:426-436` still lists four registrations for wiring a set in, all of them per-*crate*. A set added to an existing crate needs none of them, and now needs one thing that list does not mention.

**b. Temporal phrasing in the new doc comment.** `layer_profile_doc_tests.rs:128-130` ends "…which is how the configuration panel reached a release without a row." The preceding sentence already carries the invariant in present tense; the trailing clause is history, which belongs in the commit message. Delete the clause, not the sentence.

**c. `scripts/check-structure.sh:247` still states the superseded rule** — "panel providers under sets/, one crate per set". The PR amends this in `crates/README.md` and `panel-contract.md` and leaves the third statement of it contradicting them. The script's tier comment is documentation a reader trusts for the same reason the other two are.

**d. `crates/README.md:104` still reads "The shipped panels (PFD, HSI, monitor)".** The crate ships five. This predates the PR (it is the same on the base), but the PR edits that file fifteen lines above, and "a backend author reading that table believes the crate ships three panels" is #79's own complaint about the layer table.

**e. `docs/instruments/panel-contract.md:39` is an orphaned 7-character line** — `A set's` alone, with the rest of its sentence on line 40, in a document that otherwise wraps at ~70 columns. An un-reflowed insertion.

**f. The intended-function row's Inputs column omits `Trust`.** `CONFIG_DESCRIPTOR.required_groups` is `{AirframeConfig, Trust}` (`supplementary.rs:171`) and the panel's status folds source trust, but the row cites only `AIR-IN-016`. Sibling rows cite their validity input (the PFD row spans `AIR-IN-001` through `AIR-IN-009`, which includes per-signal validity at `AIR-IN-009`). The autoflight row is not a precedent for the omission — that panel requires `{ApModes, ApTargets}` and no `Trust`.

**g. "Sensed flap and trim" over-states what the panel presents.** The group carries elevator, aileron and rudder trim (`aircraft.rs:240-245`); the panel draws elevator trim only, and the other two axes get no scale and no dashes. `AIR-IN-016` says "trim on each axis the airframe declares". Either the row should say elevator trim, or the panel owes the other axes the same dash it gives an absent elevator sensor.

**h. Same name, two things.** `config.rs:5` and `supplementary.rs:163` both call this panel a "conventional-instrument function" / "conventional-instrument surface", while the new row asserts that no `AIR-OUT` requirement covers it — and `AIR-OUT-003` is titled *Conventional instrument surface*. I checked all eleven `AIR-OUT` requirements (`requirements.md:198-299`) and the row's claim is correct: `AIR-OUT-003` enumerates attitude, airspeed, altitude, heading, turn/slip and vertical speed, and none of the other ten reaches a configuration surface. The claim stands; the code's phrasing collides with a defined term that means something narrower. Worth one word in `config.rs`.

The row's other columns check out. Failure presentation: an absent sensor draws `---` and never a centred pointer (verified in `config/tests.rs` and in my probe); an out-of-range value faults the group and the panel dashes rather than clamps; the `CFG` flag and full-frame X are drawn in `Annunciation`, which is a measured criticality band. Source assumptions: "a selected detent is never shown in place of a sensed position" is true — the detent is a separate cyan mark at its own y and never moves the white sensed pointer, pinned by `a_selected_detent_does_not_move_the_sensed_pointer`. "No primary-flight credit" is consistent with the crate's own framing and with the panel shipping outside `BUILTIN_PANELS`.

**i. `GroupFault::NonFinite` is the wrong name for the fault the new requirement formalises.** `validate.rs:206` returns `NonFinite` for a finite value outside its range (`flap_ratio: Some(1.8)` is neither NaN nor infinite). Pre-existing from #75 and not introduced here, but `AIR-IN-016` now makes out-of-range a named requirement behaviour, and the taxonomy at `validate.rs:24-41` has no variant for it. Cheap to add; not this PR's job unless you want it here.

**j. Sentence length, with a provenance caveat.** I was asked to apply the language rules in `AGENTS.md`, including its sentence counts. That file is not on this branch, its base, or `main` — it exists only on the unmerged `origin/docs/track-the-documentation-language-rules`. Applying it anyway: `AIR-IN-016`'s first sentence is 33 words and its last is 39, against a 25-word limit for descriptive text; `panel-contract.md:32-37` runs 47 words in one sentence. Every sibling requirement reads the same way, so this is advisory — but if that rules file is landing, the new text is the text a reviewer will be asked to hold to it first.

---

## Battery

Run at `9944d95`, detached, in a scratch worktree. All green; working tree `git status --porcelain` empty before and after every mutation above.

```
cargo fmt --all --check                                              exit 0
cargo clippy --locked --all-targets --workspace -- -D warnings       exit 0
cargo test --locked --workspace --all-targets --no-fail-fast         exit 0   (21 suites, 0 failed)
RUSTDOCFLAGS="-D missing_docs -D rustdoc::broken_intra_doc_links" \
  cargo doc --locked --workspace --no-deps                           exit 0
cargo build --locked --release --workspace                           exit 0

check-certification-claims.sh    OK
check-instrument-requirements.sh OK (75 unique requirements; selftest 6 cases)
check-release-manifest.sh        OK (selftest 8 cases)
check-release-markers.sh         OK (0.5.0; ABI 8, scene 1, corpus 4, panels: pfd, hsi, autoflight, monitor)
check-standards-registry.sh      OK (selftest 9 cases)
check-structure.sh               OK (selftest 12 cases)
test-instrument-requirements.sh  OK
test-release-manifest.sh         OK
test-standards-registry.sh       OK
test-structure.sh                OK
trace-report.sh                  OK (informational)
detect-target.sh                 OK (no target connected)

evidence-gate --resolve-selectors    VALID  (30 nodes, 73 edges, 0 exceptions)
evidence-gate --require-resolvable   VALID  (graph digest c6e87c50…)
```

luofang34:
Both blocking findings fixed in `320edd9`, and you were right about the third thing too — the one I asserted rather than checked.

**1. `AIR-IN-016`'s final clause.** This is the worst version of the failure this PR exists to close, and you caught it: I wrote a requirement to describe the panel and never checked it against the panel. `config.rs:102` returns before the detent draw, deliberately, with a comment at `:46-48` saying why — a selection with nothing to compare it against is a number, not a configuration. That policy is defensible; my requirement contradicted it. The clause now states the implemented policy and the reason for it.

**2. The region comment.** Confirmed by decoding: flap numeral x 184→228, trim x 276→335, a 48-unit gap. Two columns. The comment now says the width spans both columns and the gap and stops short of the labels and ladders.

**3. "Verified load-bearing" did not reproduce — and I was wrong about the mechanism, not just the evidence.** `GroupRegionEmpty` fires only when a region catches *no* claim, so shrinking cannot fail until the region misses everything; I reproduced your result exactly (height 290→260 and →200 both pass). Containment is not something admission can say.

So the panel says it: `every_attributed_numeral_lies_inside_the_declared_region` decodes the scene at both ends of both scales and asserts each attributed numeral's anchor is inside the declared rectangle. Both of your shrinks now refuse. The region comment says which check holds the bounds, so the next reader does not repeat my mistake.

Separately, your observation that admission probes only two points was right and worth fixing on its own: all canonical states derive from `typical()` and the panel declared no extreme states. It contributes two now — flaps up with trim fully nose-down, flaps full with trim fully nose-up — and the config set's case count goes 30 → 42.

**4. `shipped_panels()` hard-coding two sets.** Your point that this is #79 repeating on the next set landed. There is now an `ALL_SETS` constant and the test walks it, so the table's coverage follows what the crate ships rather than what a test remembered to name.

Also took: the intended-function row named flap and trim where the panel draws elevator trim only, and omitted the trust its status folds; `crates/README.md` still listed three panels; `check-structure.sh`'s tier comment still said one set per crate; and the merge had left `panel-contract.md:39` broken mid-sentence.

**On the one-crate-many-sets call** — thank you for judging it rather than accepting it. That was the decision I most wanted an outside opinion on, and "tier law and portability are both enforced per crate" is the reasoning I would have wanted to hear back.

Full battery green: fmt, clippy `-D warnings`, workspace suite `--no-fail-fast`, doc gate, release build, and every `check-*.sh`.

luofang34:
**Verdict: request changes.** Both prior blockers are genuinely fixed and the containment test is genuinely falsifiable — I broke it three ways. Two new blockers, both prose-and-guardrail rather than behaviour: the clause this PR exists to correct has no test, and the reason given for the two extreme states is backwards in a way I could demonstrate.

Reviewed as `git diff origin/abi/v8-batch...HEAD` at 320edd9. #78 has merged into the batch, so that diff is this PR alone (the branch is two commits behind the batch head; both are #78's).

## The prior blockers

**1. `AIR-IN-016`'s final clause now matches the panel.** Read against the code clause by clause:

| clause | where it lands | holds? |
| --- | --- | --- |
| received with source identity, source time, quality | `Stamped.age_ms` for time; quality folds from the trust group at `resolve/group_status.rs:169`. Source identity is declared by no structure the group carries — but that phrasing is inherited verbatim from `AIR-IN-015`, so it predates this PR | see note 3 |
| shall declare the range each axis is defined over | `aircraft.rs:233-246` documents 0.0–1.0 for flap and −1.0–1.0 for each trim axis | yes |
| trim on each axis the airframe declares | elevator, aileron, rudder, each `Option` | yes |
| outside its range shall fail its group rather than be clamped | `validate.rs:189-208` (`airframe_fault`), tested by `out_of_range_configuration_faults_the_group_per_field` | yes |
| detent and sensed position not substituted for one another | separate fields, separate marks (cyan detent at `SCALE_X-10…-22`, white sensed at `SCALE_X+10…+22`), tested by `a_selected_detent_does_not_move_the_sensed_pointer` | yes |
| **a detent shall be presented only beside the sensed position** | `config.rs:102-112` returns after drawing the dashes, before the detent block at `:113` | yes — see blocker 1 |

The clause now states what the panel does. It is the last one that has no test.

**2. The column claim is now true.** The flap numeral anchors at `SCALE_X + 34.0` = 184; the trim numeral at `left + SCALE_LEN + 16.0` = 276. Separate columns with a gap, as the comment says.

**3. "Verified load-bearing" is correctly retracted, and the replacement is real.** `check_non_vacuity` (`admission/regions.rs:53-75`) marks a region populated when *any* case puts a claimed run's centre inside it, so shrinking passes. Confirmed:

| region change | `every_attributed_numeral_…` | `the_config_set_passes_admission` |
| --- | --- | --- |
| `height: 290 → 260` | FAILED — `a claimed numeral at (276, 320) sits outside Region { x: 170.0, y: 50.0, width: 170.0, height: 260.0 }` | ok |
| `height: 290 → 200` | FAILED — same numeral | ok |
| `x: 170 → 100` | FAILED — `(276, 320)` outside `{ x: 100.0, … }` | ok |

The new test is the guardrail; admission is not, exactly as the PR now says.

**4. `ALL_SETS` catches a third set.** I added a `PROBE_SET` holding one descriptor titled `Probe`, with no table row:

```
every_shipped_panel_has_a_row_in_shell_order      FAILED
  left:  ["PFD", "HSI", "Autoflight", "Monitor", "Configuration"]
  right: ["PFD", "HSI", "Autoflight", "Monitor", "Configuration", "Probe"]
the_documented_profiles_are_the_descriptor_masks  FAILED  (5 vs 6)
the_profiles_section_holds_exactly_one_table      FAILED
```

Removing the `| Configuration |` row from `scene-layer-protocol.md` reddens the same three. Both directions work.

**Battery at 320edd9:** `fmt --check`, `clippy --locked --all-targets --workspace -D warnings`, `test --locked --workspace --all-targets --no-fail-fast` (21 targets, 0 failed), doc gate with `-D missing_docs -D rustdoc::broken_intra_doc_links`, `build --locked --release --workspace`, every `check-*.sh` and `test-*.sh` on this branch, `detect-target.sh`, both evidence-gate invocations (VALID, 30 nodes / 73 edges), and `instrument-bench` (scene and screen-composition digests both match pin; 372 admission cases). All green.

---

## Blocking

### 1. The clause this PR was written to correct is the one clause with no guardrail

`config.rs:113` draws the detent only after the `let Some(ratio) = sensed else { … return }` at `:102`, and the comment at `:46-48` states the policy. Nothing tests it. Move the detent block above the early return:

```rust
    if let Some(detent) = selected {
        // … unchanged …
    }
    let Some(ratio) = sensed else {
```

`cargo test --locked --workspace --all-targets --no-fail-fast` — **21 targets, 0 failed.** The panel now presents a selected detent with no sensed position beside it, which is precisely what the final sentence of `AIR-IN-016` forbids, and the whole tree agrees it is fine.

Nothing can see it because no state anywhere constructs `flap_ratio: None, flap_selected_ratio: Some(_)`. `each_scale_answers_for_its_own_sensor` (`config/tests.rs:45`) sets both to `None`; `a_selected_detent_does_not_move_the_sensed_pointer` (`:78`) sets both to `Some`; the withholding matrix zeroes the whole group; both new extreme states set both fields together; the corpus sets both to 0.25. The one combination the requirement is about is the one combination nothing builds.

This is the shape that produced #79: a requirement and an implementation agreeing only because a person compared them once. The PR fixed the disagreement by rewriting the requirement to match the code — which is right — but the house rule is that the fix ships with the guardrail, and a rewritten clause with no test is one refactor away from being wrong again in the other direction.

The file already has the idiom. Roughly fifteen lines beside `a_selected_detent_does_not_move_the_sensed_pointer`: state `flap_ratio: None, flap_selected_ratio: Some(0.5)`, assert the flap dashes appear and that no polygon is emitted for the flap scale.

### 2. The reason given for the two extreme states is backwards

`supplementary.rs:204-207`:

> The corpus reaches one flap position and one trim setting, both mid-scale, so without these the region is only ever probed at two points and **a rectangle a fraction of its size would pass**. These put both numerals at the ends of their travel.

echoed at `admission/tests.rs:102-104`:

> The two extreme states put both numerals at the ends of their travel, which is what makes the declared group region **measured against more than one point**.

Non-vacuity is an existence test, so it is monotone in the case count: adding cases can only turn `populated` from false to true. Extra probe points cannot tighten it. The effect runs the other way, and the difference is observable. Shrink the region to `height: 30` — a tenth of what it declares, 170×30:

| tree | `the_config_set_passes_admission` |
| --- | --- |
| region 170×30, **both extreme states present** | **ok** |
| region 170×30, `extreme_states: &[]` | **FAILED** — `GroupRegionEmpty { panel: "config", group: AirframeConfig, region: Region { x: 170.0, y: 50.0, width: 170.0, height: 30.0 }, frame: … }` |

A rectangle a fraction of its size passes *because of* the extreme states, not despite their absence: the low extreme puts flaps at 0.0, so the numeral lands at y = 70, inside the 50–80 band the corpus's 0.25 (y = 125) never reaches. And this contradicts the comment twelve lines above at `:190-194`, which gets it right — "a region far smaller than the ink passes it".

**Keep the extreme states.** They are worth having, and the same experiment proves it: they demonstrably put ink where the corpus cannot reach, so the twelve cases they add exercise the frame-bounds, glyph-coverage and provenance checks at both ends of both scales. What needs replacing is the reason, in both comments — they widen the *drawn states* admission judges, not the region test.

While that comment block is open, two more claims in it do not survive arithmetic:

- `:184-186` — "the width … stops short of the scale labels and the ladders". The trim scale is `line(left, y, left + SCALE_LEN, y)` = (40, 320) → (260, 320); the region spans x 170–340, y 50–340. Ninety units of that ladder sit inside it, as does the trim pointer (`x = 150 + 110·ratio`) for any trim above about 0.18 — the high extreme puts it at x = 260. The `FLAP` and `TRIM` labels and the flap ladder really are outside; the trim ladder is not.
- `:187-188` — "The height is most of the frame because the flap numeral rides its pointer down the whole scale". The flap numeral tops out at y = 290 (`top + SCALE_LEN` = 70 + 220). What forces the height past that is the trim numeral at y = 320, which is what the height-260 run above named: `a claimed numeral at (276, 320)`. The flap numeral alone would fit in height 240.

The PR description carries the same reason ("Still tall, because the flap numeral rides its pointer down the whole scale"), so it wants the same correction.

---

## Non-blocking

**1. `ALL_SETS` is future-proofing, not a closed loop.** A set exported but never added to it is still invisible:

```rust
/// A set exported but never added to ALL_SETS.
pub const ORPHAN_SET: PanelSet = PanelSet { id: "orphan", panels: CONFIG_PANELS };
```

`cargo test -p indicate-instrument-panels --all-targets` → 114 passed, 0 failed. Nothing enumerates `PanelSet` constants; `ALL_SETS` is a hand-maintained literal at `descriptors.rs:338`, and the only readers are the re-export and `layer_profile_doc_tests`.

This is still a real improvement — it turns "remember to update every test that names a set" into "remember to update one constant", and it is the constant the doc comment tells you to reach for. But the failure mode it is written against ("a check that named the sets it walked would keep passing when a third set arrived") survives one level up. If you want it closed, `layer_profile_doc_tests` already parses a file with `include_str!`; the same idiom over `descriptors.rs` would let a test assert that every `pub const …: PanelSet` appears in `ALL_SETS`.

**2. The containment test checks anchor points, not extents.** `Cmd::Text { x, y, .. }` is the `MIDDLE_LEFT` anchor, so a numeral anchored one unit inside the right edge passes while its glyphs hang outside. The trim numeral at x = 276 has 64 units of margin to the region's right edge, so nothing is wrong today; worth a sentence in the test's doc comment so the next person widening a numeral knows what the assertion does and does not cover.

**3. Clause 1's "source identity"** is satisfied by no structure the airframe group carries — `Stamped` is data plus age, and quality reaches the group by folding the shared trust group rather than by a per-signal declaration. `AIR-IN-015` uses the identical phrase for autoflight and resolves the same way, so this is a pattern the batch already had rather than anything this PR introduced. Noting it only because the clause-by-clause read is what the review was for.

luofang34:
Both blockers fixed in `78f6669`. The second one is the more instructive, and I had the reasoning exactly inverted.

**1. The clause had no guardrail.** You are right that this is the failure this branch exists to close, committed by the branch itself. I reproduced your mutation — moving the detent block above the sensed early-return — and confirmed the whole workspace suite stayed green, because nothing anywhere builds a sensed-absent, detent-present configuration. `a_detent_with_no_sensed_position_draws_nothing` now does, and counts cyan fills rather than text, since the detent is the only thing the panel paints in the selection colour. Verified: the mutation reddens it, `left: 1, right: 0`.

**2. The extreme-states rationale was backwards.** Non-vacuity is monotone — a region is populated when *any one* claim lands in it, so more cases can only make it easier to pass. I reproduced your experiment at `height: 30` and admission passes with the extremes. My comment said the opposite, and it contradicted the correct note twelve lines above it, which is the part I should have noticed myself.

The states stay, for the reason your experiment actually supports: they draw each numeral where the corpus never does. Both comments now say that, and I corrected the PR description too.

**The two measurements were also wrong.** About ninety units of the trim ladder do fall inside the region, because the ladder crosses the column the flap numeral needs — "stops short of the ladders" was false. And the bottom edge is set by the trim numeral at y 320, not the flap numeral, which stops at y 290 with flaps fully extended. Both corrected.

**On `ALL_SETS` being a hand-maintained literal** — that is fair and I have left it. It converts "a check that names two sets" into "a crate that names its sets once", which is a smaller surface but not zero. Making it unfakeable means the sets registering themselves, which is a different mechanism than this PR should introduce. Worth an issue if you think it should not wait.

Battery green: fmt, clippy `-D warnings`, workspace suite `--no-fail-fast`, and every `check-*.sh`.

luofang34:
## Verified — approving. Both blockers are closed, and both arithmetic corrections hold.

Reviewed at `78f6669` merged with `origin/abi/v8-batch` (`dc29f24`, which already carries #78, so `AIR-IN-015` exists and `AIR-IN-016` is free). The merge is clean. I re-derived every number rather than reading it back.

### Blocker 1 — the clause had no guardrail — **CLOSED**

I re-applied the mutation myself: the `if let Some(detent) = selected` block moved above the `let Some(ratio) = sensed else { … return }` in `flap_scale`. Full workspace suite, `--no-fail-fast`:

```
failures:
    config::tests::a_detent_with_no_sensed_position_draws_nothing

assertion `left == right` failed: a detent drew with no sensed position to read it against
  left: 1
 right: 0
```

757 passed, 1 failed, against a 758/0 baseline. **Exactly one test reddens, and it reddens on the cyan-polygon count** — not on the dashes assertion beside it, not on anything else. Nothing in the suite catches the mutation on its own, which is the PR's stated reason for the case existing, and the run confirms it: without this test the mutation is invisible.

The counter is not trivially zero, either — the mutation drives it to 1, so the cyan-fill tracking works.

### Blocker 2 — the extreme-states rationale was backwards — **CLOSED**

The comments now say the extreme states are geometry coverage and that non-vacuity is monotone. That is correct, and I checked it two ways.

`check_non_vacuity` in `admission/regions.rs` walks the case matrix setting `populated[i] = true` on any witness and errors only where a region was never populated. `false → true` is the only transition, so more cases can only help a region find a claim. The module doc says the same thing in its own words: the assertion "cannot be 'all this group's claimed ink is inside'".

Then empirically. I shrank the declared region from `width: 170.0` to `width: 30.0`, which leaves it catching the flap numeral at x 184 and missing the trim numeral at x 276 entirely:

```
failures:
    config::tests::every_attributed_numeral_lies_inside_the_declared_region
      flap 0 trim -1: a claimed numeral at (276, 320) sits outside
      Region { x: 170.0, y: 50.0, width: 30.0, height: 290.0 }

test admission::tests::the_config_set_passes_admission ... ok
```

A region a fraction of its declared width passes admission with both extreme states present. So the old comment's claim — that without the extreme states "a rectangle a fraction of its size would pass" — was false in both directions, and the replacement is right. It also settles the PR body's "**Verified load-bearing** by `every_attributed_numeral_lies_inside_the_declared_region`, not by admission": that is exactly what the two lines above show.

### The two arithmetic corrections — both confirmed by measurement

I emitted the scene through the public API and swept both axes at 21 points each, rather than reading the constants.

**"About ninety units of the trim ladder fall inside it."** Exact:

```
LINE (40, 320)-(260, 320)   x-overlap-with-region = 90   <-- trim ladder
TOTAL unclaimed line length inside region = 90
```

The ladder spans x [40, 260] at y 320; the region spans x [170, 340]. Overlap 260 − 170 = **90**. Every other line the panel draws (the flap ladder at x 150 and its rungs at x [142, 150], the trim neutral tick at x 150) has zero overlap, so 90 is the whole of it.

The stated *reason* also holds, and it is stronger than the comment claims. The region must contain the flap numeral's column (x 184) and the trim numeral (276, 320). Any rectangle containing both points contains (184, 320), which lies on the ladder — so the crossing is structural, not slack. The minimum unavoidable overlap is 260 − 184 = 76 units; the declared region's 90 is that plus the 14 units of left margin.

**"The bottom is set by the trim numeral at y 320, not by the flap numeral, which stops at y 290."** Measured over the full sweep:

```
FLAP numeral: x[184, 184] y[70, 290]   (21 samples)
TRIM numeral: x[276, 276] y[320, 320]  (21 samples)
BINDING BOTTOM: flap max y = 290, trim max y = 320 -> binder = TRIM
```

The flap numeral rides its pointer from y 70 to y 290 and stops there; the trim numeral is fixed at y 320. The binding constraint on the bottom edge is the trim numeral. Both directions of the correction are right.

For the record, the region is `x [170, 340] y [50, 340]` against an attributed bounding box of `x [184, 276] y [70, 320]` — margins of 20 top, 20 bottom, 14 left, 64 right. The generous right margin is the text extent: both numerals anchor `MIDDLE_LEFT` and run rightward.

### Do the corrected comments contradict each other, or the note above them?

No. Read in order they now say: only attributed runs are measured; the region spans both numeral columns and does not exclude the ladder; admission does not hold these bounds; the panel-local test does; and the extreme states are coverage rather than constraint. Each of those is separately true and none contradicts the descriptor's own doc comment above it, which is about set membership and the composition digest.

### Also checked

The layer-profile guardrail is load-bearing, not just present. Dropping the `Configuration` row from `scene-layer-protocol.md` reddens three tests, all naming it:

```
left:  ["PFD", "HSI", "Autoflight", "Monitor"]
right: ["PFD", "HSI", "Autoflight", "Monitor", "Configuration"]
```

Because they walk `ALL_SETS` rather than a named list, a third set added later is covered without touching them.

The admission count arithmetic checks: `required_groups` is two, so (5 canonical + 2 extreme) × (1 fed + 2 withheld) × 2 alert states = 42, up from 5 × 3 × 2 = 30.

`crates/README.md` and `panel-contract.md` now state the same rule rather than a rule the crate breaks, and `AIR-IN-016` carries the detent clause the new test pins.

### Gates

Green on the merged checkout: `fmt --all --check`; `test --locked --workspace --all-targets --no-fail-fast` (758 passed, 0 failed, on both 1.98 and 1.95); the doc gate with `-D missing_docs -D rustdoc::broken_intra_doc_links`; `build --locked --release --workspace`; every `scripts/check-*.sh` and `scripts/test-*.sh` present on this branch (10 of them); `trace-report.sh`; `detect-target.sh`; and both evidence-gate invocations. `git status --porcelain` empty after every mutation was restored.

**One caveat, and it is not yours.** `cargo clippy` on latest stable (1.98) fails on this checkout, entirely in files this PR does not touch:

- `crates/indicate-evidence/src/gate/review.rs:60` — `useless_format`
- `tools/instrument-bench/src/main.rs:241`, and eight sites across `indicate-instrument-raster` — `chunks_exact_to_as_chunks`

All of them are fixed on `main` by #83; `abi/v8-batch` has not taken it. `main` carries zero `chunks_exact` occurrences, and I confirmed the same command is green on my `main`-based checkout. Re-running with `-A clippy::useless_format -A clippy::chunks_exact_to_as_chunks` is green, and 1.95 (the pinned `stable` locally) is green unqualified — so **#80's own code is clippy-clean on 1.98**. This is a base-branch item for the batch to pick up, not a reason to hold this PR.

Three scripts in CONTRIBUTING's battery — `check-recorded-counts.sh` and the two `baseline-survives-merge` scripts — do not exist on `abi/v8-batch` yet; they arrive with `main`. Nothing to do here.

Approving. Good to merge into the batch.

PR #81, opened 2026-08-21
Two entries under one version leave one of them unchecked

luofang34:
`check-release-markers.sh` reads the newest release entry and compares its five values against the tree. When two branches each add the next release and both land, the file carries that version twice — and the older twin's values are never compared against anything. The check reports OK, because it found an entry it was happy with and stopped.

Three branches in flight right now (#59, #61, #63) each add a `## [0.5.1]`, which is what made this worth writing down rather than remembering to renumber by hand.

## The choice it makes

It refuses the file, not one of the entries. Which of two entries under one version is *the* release is not a question this check can answer, and picking one would be the fail-open answer — it would keep reporting OK on a changelog that says two different things about one version.

## The selftest

New `scripts/test-release-markers.sh`, five cases, wired as a hard CI step before the check itself.

The duplicate case plants a **complete** copy of the newest entry, table and all. A twin without a table is refused for a missing row instead, so a simpler case would pass with the duplicate check deleted — I wrote that version first and it did exactly that. Verified the case reddens when the guard is removed and only then.

The other four cover the paths that had no coverage at all: a value that disagrees with the tree, a changelog with no versioned entry, and the committed file before and after the mutations, so a case that left the file dirty would show up.

Local: `check-release-markers.sh` OK, `test-release-markers.sh` OK (5 cases), workspace suite green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,9 @@ jobs:
         # merge commit, which does not exist while the pull request that
         # writes the entry is open, so the check would fail every release
         # on the run that matters. Tagging is a step in CONTRIBUTING.md.
-        run: bash scripts/check-release-markers.sh
+        run: |
+          bash scripts/test-release-markers.sh
+          bash scripts/check-release-markers.sh
 
       - name: release manifest drift guard
         # The changelog entry is prose a person reads; release-manifest.json
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,6 +41,8 @@ bash scripts/check-recorded-counts.sh --selftest
 bash scripts/check-recorded-counts.sh
 bash scripts/test-baseline-survives-merge.sh
 bash scripts/check-baseline-survives-merge.sh
+bash scripts/test-release-markers.sh
+bash scripts/check-release-markers.sh
 ```
 
 The evidence gate binds recorded test sources by content digest: editing
--- a/scripts/check-release-markers.sh
+++ b/scripts/check-release-markers.sh
@@ -57,6 +57,18 @@ entry="$(awk '
 
 version="$(printf '%s\n' "$entry" | awk 'NR == 1 { gsub(/^## \[|\].*$/, ""); print; exit }')"
 
+# Two entries under one version is the shape a merge produces when two
+# branches each add the next release. This check reads the newest entry
+# only, so the older twin's values are never compared against anything
+# and its claims go unchecked. Refuse the file rather than the entry:
+# which of the two is "the" release is not a question this check can
+# answer.
+duplicate="$(grep -oE '^## \[[0-9]+\.[0-9]+\.[0-9]+\]' "$changelog" \
+    | sort | uniq -d | head -1 || true)"
+if [ -n "$duplicate" ]; then
+    fail_closed "$changelog declares $duplicate more than once; two entries under one version leave the older one unchecked (fail-closed)"
+fi
+
 # The value cell of a `| <label> | <value> |` row in the entry's table.
 # Rows inside a fenced block or an HTML comment are illustration, not
 # claim, so a table that exists only as an example cannot stand in for
--- a/scripts/test-release-markers.sh
+++ b/scripts/test-release-markers.sh
@@ -93,6 +93,22 @@ DRIFT
 refuses "a value that disagrees with the tree" "DRIFT:"
 cp "$pristine" "$changelog"
 
+# An entry that exists but omits one of the five rows. This is the only
+# refusal path with no case of its own otherwise: the check would still
+# refuse a changelog missing every row, so a case that dropped them all
+# would pass on the no-versioned-entry guard instead.
+python3 - "$changelog" <<'DROP'
+import sys, pathlib, re
+p = pathlib.Path(sys.argv[1])
+s = p.read_text()
+# Inside the newest release entry, not the legend table above it.
+start = re.search(r"^## \[\d+\.\d+\.\d+\]", s, re.M).start()
+head, tail = s[:start], s[start:]
+p.write_text(head + re.sub(r"^\| Scene format \|[^\n]*\n", "", tail, count=1, flags=re.M))
+DROP
+refuses "an entry missing one of the five rows" "states no 'Scene format' row"
+cp "$pristine" "$changelog"
+
 # No versioned entry at all.
 python3 - "$changelog" <<'STRIP'
 import sys, pathlib, re
--- a/scripts/test-release-markers.sh
+++ b/scripts/test-release-markers.sh
@@ -11,33 +11,58 @@ root_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 cd "$root_dir"
 
 check="$root_dir/scripts/check-release-markers.sh"
-changelog="CHANGELOG.md"
+work="$(mktemp -d)"
+changelog="$work/CHANGELOG.md"
+output_file="$work/output"
+export INDICATE_CHANGELOG="$changelog"
 passed=0
 failed=0
 
-backup="$(mktemp)"
-cp "$changelog" "$backup"
-cleanup() {
-    cp "$backup" "$changelog"
-    rm -f "$backup"
-}
-trap cleanup EXIT
+# Every case works on a copy. The committed CHANGELOG is never written,
+# so a case that aborts cannot leave the repository dirty and a missing
+# restore cannot hide behind the next case's own restore.
+pristine="$work/pristine"
+cp CHANGELOG.md "$pristine"
+cp "$pristine" "$changelog"
+trap 'rm -rf "$work"' EXIT
 
-# `expect <want> <label>` runs the check against the CHANGELOG as it
-# currently stands and compares its exit status to `want`.
-expect() {
-    local want="$1" label="$2" got=0
-    "$check" >/dev/null 2>&1 || got=$?
-    if [ "$got" -eq "$want" ]; then
-        echo "ok: $label"
+# `accepts <label>` expects a clean run. `refuses <label> <phrase>`
+# expects a refusal AND that the refusal names `phrase`: the check has
+# several ways to refuse, so a case that only counted the exit status
+# would pass on a refusal it did not ask for. That is the shape the
+# duplicate-version case was written in first, and it passed with the
+# guard it named deleted.
+run_check() {
+    "$check" >"$output_file" 2>&1
+}
+
+accepts() {
+    if run_check; then
+        echo "ok: $1"
         passed=$((passed + 1))
     else
-        echo "REGRESSION: $label — wanted exit $want, got $got" >&2
+        echo "REGRESSION: $1 — refused, saying: $(head -1 "$output_file")" >&2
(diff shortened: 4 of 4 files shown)


luofang34:
**Verdict: request changes.** The fix is right and I reproduced the hole it closes; the duplicate case is genuinely falsifiable, so the rewrite you describe worked. But one of the other four — case 4, "a changelog with no versioned entry is refused" — has exactly the defect you caught and fixed in the duplicate case: it stays green with the guard it names deleted.

## What I verified first

**The hole, with a twin that actually lies.** I duplicated the newest entry and set the *older* twin's `State ABI` to 99, so the file makes two contradictory claims about 0.5.0 and one of them is false:

```
INDICATE_CHANGELOG=/tmp/CHANGELOG-dup-lying-twin.md bash scripts/check-release-markers.sh
```

| script | result |
| --- | --- |
| `origin/main` | `check-release-markers: OK (0.5.0; ABI 7, …)` — exit 0 |
| this branch | `… declares ## [0.5.0] more than once … (fail-closed)` — exit 1 |

The false claim rides through on `main`. Refusing the file rather than picking an entry is the right call, and the comment at `:60-65` says why better than a reviewer could.

**Case 2 is falsifiable, and your diagnosis of the first version was right.** Deleting the duplicate guard (`check-release-markers.sh:66-70`) reddens case 2 and *only* case 2:

```
ok: the committed CHANGELOG agrees with the tree
REGRESSION: a version declared twice is refused — wanted exit 1, got 0
ok: a value that disagrees with the tree is refused
ok: a changelog with no versioned entry is refused
ok: the restored CHANGELOG agrees again
release-markers-selftest: FAILED (1 of 5 cases)
```

And the weak version does behave as you describe: with the guard removed, a twin that is a bare heading is already refused (exit 1, five MISSING rows), so that case would have passed with the guard gone. The complete-copy plant at `:47-56` is what makes case 2 mean something.

**Cases 1, 3 and 5 are live.** Removing `compare "State ABI"` reddens case 3. Bumping `SCENE_FORMAT_VERSION` in the tree reddens cases 1 and 5 together. Removing the final restore reddens case 5.

**CI wiring is correct** (by inspection): `ci.yml:62-64` uses `run: |` with no `shell:` override anywhere in the workflow, so the step runs under `bash -e {0}` and a `test-release-markers.sh` failure aborts the step before the check runs. The step carries no `continue-on-error` — the two in the file are at `:206` and `:256`, on other steps.

**Battery at 625a92d:** `fmt --check`, `clippy --locked --all-targets --workspace -D warnings`, `test --locked --workspace --all-targets --no-fail-fast` (21 targets, 0 failed), doc gate with `-D missing_docs -D rustdoc::broken_intra_doc_links`, `build --locked --release --workspace`, every `check-*.sh` and `test-*.sh`, both evidence-gate invocations (VALID), `check-recorded-counts` full and `--selftest`. All green. `shellcheck` on both scripts reports only the pre-existing SC2016 info at `check-release-markers.sh:118`.

---

## Blocking

### The selftest's `expect` helper discards the message, and case 4 passes for the wrong reason

`test-release-markers.sh:77` claims to cover the fail-closed guard at `check-release-markers.sh:56`. Delete that guard:

```sh
# in scripts/check-release-markers.sh, remove line 56:
#   [ -n "$entry" ] || fail_closed "no versioned release entry found in $changelog (fail-closed)"
bash scripts/test-release-markers.sh
```

```
ok: the committed CHANGELOG agrees with the tree
ok: a version declared twice is refused
ok: a value that disagrees with the tree is refused
ok: a changelog with no versioned entry is refused      <-- still green
ok: the restored CHANGELOG agrees again
release-markers-selftest: OK (5 cases)
```

Why it stays green — the check refuses the file anyway, for a different reason:

```
MISSING: entry  states no 'State ABI' row
MISSING: entry  states no 'Scene format' row
MISSING: entry  states no 'Corpus' row
MISSING: entry  states no 'Composition digest' row
MISSING: entry  states no 'Panel set' row
check-release-markers: FAILED
```

Five MISSING rows with an empty version in the message. That is not "no versioned entry is refused"; it is the compare loop tripping over an empty `entry`. The case cannot tell the two apart because `expect` at `:28-38` asserts on the exit status alone (`"$check" >/dev/null 2>&1`).

This is the same shape you found and fixed one case up. The remedy is one case down: assert on the message, which is what the sibling selftest already does at `test-release-manifest.sh:59` — `show_failure "$1 was refused, but not for '$3'"`.

Two more symptoms of the same helper, which I'd fix in the same change rather than list separately:

**The dirty-file guard only covers the last mutation.** The description says the before/after pair means "a case that left the file dirty would show up". Removing the restore after the *duplicate* case (`:59`) changes nothing — all five cases stay green, because the drift case's own `cp "$backup" "$changelog"` at `:69` repairs the file before anyone looks. Only removing the *final* restore at `:78` reddens case 5. An intermediate case that forgot its restore would land unnoticed.

**The selftest writes to the tracked `CHANGELOG.md`**, four times, restoring through an `EXIT` trap — while `check-release-markers.sh:28` has read `${INDICATE_CHANGELOG:-CHANGELOG.md}` since before this branch. I used that override for every reproduction above; none of them touched the working tree. The sibling states the convention in its own header (`test-release-manifest.sh:11-12`): "Every case points the guard at a copy under a temporary directory; the committed manifest is never written to." As written, a hard interruption between plant and restore leaves a contributor with a duplicated changelog, and in CI a crashed selftest hands the mangled file to the `check-release-markers.sh` on the next line of the same step.

All three go away together: point each case at a temp copy via `INDICATE_CHANGELOG` and give `expect` an expected-message argument.

---

## Non-blocking

**1. This is the only `check-*.sh` that does not run its own selftest.** The other four all do:

```
scripts/check-structure.sh:403             INDICATE_STRUCTURE_SELFTEST_CHILD=1 bash …/test-structure.sh
scripts/check-release-manifest.sh:81       INDICATE_RELEASE_MANIFEST_SELFTEST_CHILD=1 bash …/test-release-manifest.sh
scripts/check-instrument-requirements.sh:154  "$root_dir/scripts/test-instrument-requirements.sh" || status=1
scripts/check-standards-registry.sh:273       "$root_dir/scripts/test-standards-registry.sh" || status=1
```

Here the selftest exists only as a CI step, and neither script appears in CONTRIBUTING.md's local battery block (`check-release-markers.sh` is named only in prose at `:108`). So a contributor running the documented battery before pushing never runs it, and finds out from CI. Self-invocation would need an `INDICATE_RELEASE_MARKERS_SELFTEST_CHILD` guard, since the selftest calls the check — the same shape as the four above. Adding both lines to the battery block would do just as well.

**2. The duplicate grep is fence-blind, where the rest of the script is not.** `declared()` at `:72-90` goes out of its way to skip fenced blocks and HTML comments because "rows inside a fenced block or an HTML comment are illustration, not claim". The new grep at `:66` reads every line. A changelog that documents its own format refuses itself:

````
## How to write an entry

```
## [0.5.0] — 2026-08-18
| State ABI | 7 |
```
````

→ `declares ## [0.5.0] more than once … (fail-closed)`.

It fails closed, so nothing unsafe follows from it, and the current `CHANGELOG.md` has no fenced block. Worth knowing that the newest-entry `awk` at `:44-54` shares the blindness and is pre-existing: a fenced illustration would also *shadow* the real entry there, which fails open. Only `declared()` ever got the fence handling. Not this PR's to fix, but it is the reason the inconsistency is easy to trip over later.

**Comment by luofang34** [comment]
## Verified — approving. The blocking finding is closed.

Reviewed at `782814f` merged with `origin/main` (`05f44c1`); the branch was 32 commits behind, and the merge is clean. Toolchain `1.98` stable, which is the clippy CI tracks.

### Finding: case 4 stayed green when the guard it named was deleted — **CLOSED**

The `accepts` / `refuses` pattern is the one `test-release-manifest.sh` uses, down to the wording of both refusal messages, and `refuses` now greps the captured output for the phrase. I deleted each guard in turn and ran the selftest. Exactly one case reddens each time, and each reddens for its own reason:

| Guard deleted | Cases red | The message |
| --- | --- | --- |
| duplicate block | case 2 only | `a version declared twice was accepted; the guard did not notice` |
| `DRIFT` branch in `compare` | case 3 only | `a value that disagrees with the tree was accepted; the guard did not notice` |
| `[ -n "$entry" ]` fail-closed | case 4 only | `a changelog with no versioned entry was refused, but not for 'no versioned release entry found'` / `it said: MISSING: entry  states no 'State ABI' row` |

The third row is the finding itself, caught. The check still refuses when the entry guard is gone — five `MISSING` rows, exactly as before — and the case now reddens anyway because the refusal does not name what the case asked for. That is the behaviour the old `expect` could not see.

The two `accepts` cases are falsifiable too, which the guard deletions cannot show. I moved `SCENE_FORMAT_VERSION` from 1 to 4 in the tree and left the changelog alone: cases 1 and 5 redden with `DRIFT: entry 0.5.3 says 'Scene format' is '1'; the tree says '4'`, and cases 2, 3 and 4 stay green because their phrases still appear. No cross-talk in either direction.

### Finding: the tracked `CHANGELOG.md` must survive an abort — **CLOSED**

Every read in the check goes through `$changelog`, so `INDICATE_CHANGELOG` redirects all four of them. I injected `exit 7` into the selftest immediately after the duplicate plant and before its restore, so the run dies with a mutated file in play:

```
ok: the committed CHANGELOG agrees with the tree
ok: a version declared twice refused
INJECTED ABORT
SELFTEST EXIT=7
```

`CHANGELOG.md` is `310a3e4c…` before and `310a3e4c…` after — byte identical — and `git status --porcelain` is empty. The `trap … EXIT` takes the temp directory with it.

### CI wiring

The `release marker drift guard` step runs the selftest before the check in one `run:` block. GitHub runs that block under `bash -e`, so the selftest's exit status fails the step, and the step carries no `continue-on-error` — the two that do are further down the file and are the evidence advisories. A failure here fails the build. Both scripts are also in CONTRIBUTING's local battery.

### The real changelog

`main` renumbered the three in-flight entries to `0.5.1`, `0.5.2` and `0.5.3`, so the file now carries four `0.5.x` headings and no twins. `check-release-markers.sh` reports `OK (0.5.3; ABI 7, scene 1, corpus 5, panels: pfd, hsi, monitor)`. This is the situation the PR was written for, and the check passes it.

### Gates

All green on the merged checkout: `fmt --all --check`, `clippy --locked --all-targets --workspace -- -D warnings`, `test --locked --workspace --all-targets --no-fail-fast` (734 passed, 0 failed across 21 binaries), the doc gate with `-D missing_docs -D rustdoc::broken_intra_doc_links`, `build --locked --release --workspace`, all 14 `scripts/check-*.sh` and `scripts/test-*.sh`, `check-recorded-counts.sh --selftest`, `trace-report.sh`, `detect-target.sh`, and both evidence-gate invocations. `git status --porcelain` empty after every mutation was restored.

### Non-blocking, for later

Deleting the `MISSING` branch from `compare` reddens nothing. It is the one refusal path with no case of its own, and case 4 no longer stands in for it now that it asserts its own phrase — which is the right trade, not a regression. A sixth case that plants an entry with one row removed would close it. Outside this PR.

`shellcheck` reports SC2016 (info) on the panels `awk` at line 118. It is on `main` already and is not in CI; the single quotes are deliberate.

